PR #47, opened 2022-12-21
Cnmp 375 actions build containers

szamfir:


stevegsa (approved):
Looks great!  One question~ is there a reason we changed the conf and xml file permissions to executable?

szamfir:
> Looks great! One question~ is there a reason we changed the conf and xml file permissions to executable?

As far as I know we did not explicitly change them to executable. We changed the file ownership to nifi. If they are executable that may be due me being on windows.

PR #71, opened 2023-01-18
Cnpt 548 patient profile UI

hclarkEnq:
- Added patient profile page
- Added top banner for the patient profile
- Created reusable static table component


mpeels:
Can we go ahead and hook up the link from the patient search results to the patient profile page? Even if the profile page isn't complete, we will at least be able to navigate to it after performing a search.

hclarkEnq:
> Can we go ahead and hook up the link from the patient search results to the patient profile page? Even if the profile page isn't complete, we will at least be able to navigate to it after performing a search.

Yes it is doing it right not from the basic patient search results after clicking on the patient name

mpeels (approved):

PR #73, opened 2023-01-19
CNPT-661 finalize demographic api

Emelias-Enquizit:


patient-search/api/src/main/java/gov/cdc/nbs/controller/PatientController.java
```diff
@@ -60,5 +61,11 @@ public Optional<Person> findPatientById(@Argument Long id) {
     public Person createPatient(@Argument PatientInput patient) {
         return patientService.createPatient(patient);
     }
+    
+	@MutationMapping()
+	@PreAuthorize(FIND_PATIENT)
+	public PatientUpdateResponse updatePatient(@Argument Long id, @Argument PatientInput patient) {
+		return patientService.sendUpdatePatientEvent(id, patient);
+	}
```
mpeels, line 69: I believe we need to check for `FIND_PATIENT` as well as `EDIT_PATIENT` permissions here. And upon looking over these permissions again, we might need to add `VIEW_PATIENT` as well.

image

Emelias-Enquizit: @mpeels  sure I will updated I placed at FIND_PATIENT since technically is the updated is being made on the consumer side. Also the view above you  seeing it using your account or msa?

mpeels: the above list is accessed by logging into NBS as `msa` then navigating to -> System Management -> Security Management -> Manage User Permission Sets -> Then click any permission set to view rules available

Emelias-Enquizit: ok Thank you.

mpeels: On where to check permissions, my thoughts were that the patient-search API would check permissions and then post the request to the queue as it has the current user credentials in hand. Then the listener would know that any request it receives has been validated.

I can be persuaded to handle it differently if you think another approach is better.

Emelias-Enquizit: @mpeels no that makes sense I'll will update the permissions on the controller function to Include  FIND_PATIENT + EDIT_PATIENT + VIEW_PATIENT so it is already validated on request.

mpeels (approved):

PR #74, opened 2023-01-19
CNPT-723 CNPT-641 Replace ElasticSearch with OpenSearch

stevegsa:
Leave the elasticsearch/Dockerfile and create a new opensearch/Dockerfile so we can compare the differences and switch between them if issues arise.  Right now the only difference is that opensearch has it's own security plugin and does not use xpack.

Up for feedback on whether we should rename all references to `elasticsearch` in the containers and documentation to `opensearch`.  As it's currently implemented here it retains the `elasticsearch` container name.

Here we validate opensearch is running and working with nifi:
Screenshot 2023-01-19 at 12 59 40 AM

Screenshot 2023-01-19 at 12 38 04 PM

TODO:
spring-data-elasticsearch 4.3.x and up do version checks on elasticsearch and won't work with elasticsearch so we need to remove the annotations that are from 4.3.x for example `org.springframework.data.elasticsearch.annotations.ValueConverter`
(small amount of rewrite)
or switch to opensearch libraries (big rewrite)

### DIFF
--- a/.github/workflows/dockerfile-to-ecr.yaml
+++ b/.github/workflows/dockerfile-to-ecr.yaml
@@ -65,7 +65,7 @@ jobs:
         IMAGE_NAME: elasticsearch
         ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
         ECR_REPOSITORY: ${{ secrets.ECR_REPO_NAME }}
-        IMAGE_TAG: "${{ env.github_repo_name }}-elasticsearch-${{ env.github_branch }}-${{ env.github_sha_short }}"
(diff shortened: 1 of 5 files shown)


.github/workflows/dockerfile-to-ecr.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
         IMAGE_TAG: "${{ env.github_repo_name }}-elasticsearch-${{ env.github_branch }}-${{ env.github_sha_short }}"
```
szamfir, line 68: image tag should reflect name change to opensearch.

szamfir (commented):
No major issues stand out, with only one small change. Please hold off on merging to main as this may create issues. Will bring it up at standup tomorrow 1/20.

stevegsa:
closing this pr for now as we are sticking with elasticsearch in the short term

PR #75, opened 2023-01-19
CNPT 727 proxy rewrite urls

mpeels:
Update the urls for the redirect routes to better accommodate the Kubernetes  deployment. Fix the Navbar links


.github/workflows/dockerfile-to-ecr.yaml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - CNPT-727-proxy-rewrite-urls
```
szamfir, line 6: Please delete this line from the push to main

mpeels: good catch. Removed

szamfir (approved):
LGTM, Thanks!

PR #78, opened 2023-01-23
-> Patient profile UI

hclarkEnq:
- added all tabs
- Added tables for each of the tabs
- Back to top function


patient-search/ui/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,91 @@
...
+    tableBody,
+    isPagination = false,
+    totalResults = 20,
+    currentPage = 1,
+    handleNext,
+    buttons,
+    tableSubHeader
+}: TableContentProps) => {
+    return (
+        <div className="">
```
GangulaMounika, line 28: Class name is missing here. can we have a placeholder.

patient-search/ui/src/components/Table/Table.tsx
```diff
@@ -0,0 +1,91 @@
+import { Button, Icon, Table, Pagination } from '@trussworks/react-uswds';
+import React from 'react';
+
+type TableContentProps = {
+    tableHeader?: string;
+    tableSubHeader?: React.ReactNode | React.ReactNode[] | string;
+    tableHead: { name: string; sortable: boolean }[];
+    tableBody: any[];
```
GangulaMounika, line 8: Can we have type for tableBody here. looks like one of the property is title

patient-search/ui/src/pages/patientProfile/Demographics.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+export const Demographics = () => {
+    const [tableBody, setTableBody] = useState<any>([]);
+    const [nameTableBody, setNameTableBody] = useState<any>([]);
+    const [currentPage, setCurrentPage] = useState<number>(1);
+
+    useEffect(() => {
+        const tempArr = [];
+        for (let i = 0; i < 3; i++) {
+            tempArr.push({
```
GangulaMounika, line 13: This hard coded data is going to be removed anyways. 
But in future may be we can think about having a fixtrure with sample.json while the api is not ready. We can use the fixture sample. And we can leave the fixture for local development purposes.

GangulaMounika (commented):
For styling, again not for this PR but for future PRs, do we continue with the "inline styles" using the style attribute? Or do we go with a .scss file for each component. I think the later is better what say?

GangulaMounika (commented):
Since these are new files, can we start adding unit test cases? If not able to scope in this jira ticket, lets create another one and allocate some points to get this done.

patient-search/ui/src/pages/advancedSearch/components/InvestigationResults.tsx
```diff
@@ -108,6 +108,22 @@ export const InvestigationResults = ({
                         style={{ wordBreak: 'break-word', cursor: 'pointer' }}>
                         {name}
                     </p>
+                    <p
+                        onClick={async () => {
```
GangulaMounika, line 112: Better have this function defined with a specific name above and reference here.

patient-search/ui/src/components/Table/Table.tsx
hclarkEnq: Yes good idea

patient-search/ui/src/pages/patientProfile/Demographics.tsx
hclarkEnq: Yes also agree with this style of implementation for usage of dummy data. will integrate onwards

patient-search/ui/src/components/Table/Table.tsx
hclarkEnq: > 

Correct we can, will implement

hclarkEnq:
> For styling, again not for this PR but for future PRs, do we continue with the "inline styles" using the style attribute? Or do we go with a .scss file for each component. I think the later is better what say?

I agree I should be consistent with the way of using styles and should always be an external file only used for demo purposes probably but will clean up next

patient-search/ui/src/pages/patientProfile/style.scss
```diff
@@ -1,9 +1,33 @@
 .delete-btn {
```
GangulaMounika, line 1: I think important! will become difficult to handle later down the line. You might be able to provide a stricter selector other than just .delete-btn as a work around. Fine for this PR though.

GangulaMounika (approved):

patient-search/ui/package.json
```diff
@@ -16,6 +16,7 @@
         "cypress-cucumber-preprocessor": "^4.3.1",
         "graphql": "^16.6.0",
         "lodash.debounce": "^4.0.8",
+        "moment": "^2.29.4",
```
mpeels, line 19: From the [momentjs docs](https://momentjs.com/docs/#/-project-status/)


> We recognize that many existing projects may continue to use Moment, but we would like to discourage Moment from being used in new projects going forward

mpeels:
closing as this was merged as part of https://github.com/CDCgov/NEDSS-Modernization/pull/113

PR #79, opened 2023-01-24
Feature/cnpt 667 UI unit test cases introduction

GangulaMounika:
Added Unit test cases for around 20 files and rest just added as placeholders


mpeels (approved):
Test results:

image

PR #99, opened 2023-02-07
Change project structure. Create database-entites library

mpeels:
Creates a `database-entities` library to hold `@Entity` and `@Document` models as these will need to be shared between the API and the Kafka Listener projects.

Restructures the repo into a more standard monorepo format.

cdc-sandbox
apps    
  -- modernization-api
  -- modernization-ui 
  -- patient-listener
libs    
-- database-entities


szamfir (commented):
A couple of comments on .github/workflows/dockerfile-to-ecr.yaml.

1. If we are renaming patient-search to nbs-modernization does this imply that all future features/services will fall into this one container? What might the naming look like?
2. Line 109 would need to be modified to reflect the name change pending reflection on the first point. This may be a minor point as build file will likely be separated out for each container built.

szamfir:
A question I have on this change has to do with future containers related to new services. Having the container image named 

> > A couple of comments on .github/workflows/dockerfile-to-ecr.yaml.
> > 
> > 1. If we are renaming patient-search to nbs-modernization does this imply that all future features/services will fall into this one container? What might the naming look like?
> > 2. Line 109 would need to be modified to reflect the name change pending reflection on the first point. This may be a minor point as build file will likely be separated out for each container built.
> 
>     1. I think I prefer removing the `patient-search` or `nbs-modernization` folder altogether. That way at the root level we have `cdc-sandbox`, `api`, `ui`, `common`, and `patient-listener`. In the future, any other listeners or services we need would just be a new folder at the root level. thoughts?
> 
>     2. I may have to rely on your expertise here in making sure the image build works correctly.

1. Considering 2 services, say patient-search and syphilis, how would this look in the situation being described? I.e. would we have something like patient-search-api and patient-search-ui or something different? Talks of repo structure (mono vs multi) are still somewhat ongoing so this may be best discussed further.
2. The name I was referencing is just an image tag. It will not break anything to change it, we will just have to look for the new name when deploying into K8s cluster.

.github/workflows/sonar.yaml
```diff
@@ -5,21 +5,21 @@ name: "Sonar Scanner"
 on:
   push:
     paths:
-      - 'patient-search/**'
-      - '.github/workflows/codeql.yaml'
+      - "apps/modernization-api/**"
```
szamfir, line 8: This could either be rewritten as 
1. "apps/**", which would run the build process whenever a change within the apps directory is made
2. OR "apps/modernization-api/**" and "app/modernization-ui/**", to run the build process whenever something within these either of these two folders, but NOT patient-listener were made

.github/workflows/dockerfile-to-ecr.yaml
```diff
@@ -101,15 +101,15 @@ jobs:
...
+      id: modernization-api
       env:        
         ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
         ECR_REPOSITORY: ${{ secrets.ECR_REPO_NAME }}
-        IMAGE_TAG: "${{ env.github_repo_name }}-patient-search-${{ env.github_branch }}-${{ env.github_sha_short }}"
+        IMAGE_TAG: "${{ env.github_repo_name }}-modernization-api-${{ env.github_branch }}-${{ env.github_sha_short }}"
       run: |
         # Build a docker container and push it to ECR 
-        docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG ./patient-search
+        docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG ./apps/modernization-api
```
szamfir, line 112: Are you now splitting up the api and ui into 2 separate containers? That is what I am getting from the 2 dockerfiles. If that is the case we will need to account for that in these build files. This may be a good time to split these up (probably an A&I task). Will require additional work for deployment to k8s as well.

szamfir (changes requested):
One small fix in dockerfile-to-ecr.yaml. Otherwise looking good!

szamfir (approved):

adamloup-enquizit (approved):
👍

stevegsa (approved):

PR #101, opened 2023-02-09
updates Maven 3.8.6 to 3.8.7

adamloup-enquizit:
There was a patch release of maven `3.8` increasing the  version from `3.8.6` to `3.8.7`.  The binaries for `3.8.6` were no longer available causing the building of NBC Classic via `cdc-sandbox/build.sh` to fail.  Build arguments were introduced to reduce having to keep the version numbers the same through the script.


szamfir (approved):
Looks good thanks for taking this on. Hoping to simplify the legacy deploy in the near future.

PR #102, opened 2023-02-13
Modernization-api externalized configuration

adamloup-enquizit:
Changed the `modernization-api` container to pull the `elasticsearch`, 'nbs-mssql`, and 'wildfly` configuration from Spring Config sources (i.e environment variables, java system properties).  

Reduces the build context of the `modernization-api` by excluding the `cdc-sandbox` directory.  This can be removed in the future if the gradle project is moved into a sibling folder of `cdc-sandbox`.

Reduces the size of the resulting `modernization-api` image by only including the files required to run the application.  This can further reduced if needed with a more precise base image for the application container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+cdc-sandbox
+node_modules
+build
+log
--- a/.github/workflows/dockerfile-to-ecr.yaml
+++ b/.github/workflows/dockerfile-to-ecr.yaml
@@ -109,7 +109,7 @@ jobs:
         IMAGE_TAG: "${{ env.github_repo_name }}-modernization-api-${{ env.github_branch }}-${{ env.github_sha_short }}"
       run: |
         # Build a docker container and push it to ECR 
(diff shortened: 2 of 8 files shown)


mpeels (approved):

szamfir (approved):
It looks good to me.
For deployment to a cluster, environment variables for the container are supplied directly helm in the form (--nbs.elasticsearch.url=<some-value>) at container start-up. If we should be using expect environment variables instead, we'll need to parameterize a few more variables and modify existing ones like the urls.

adamloup-enquizit:
> It looks good to me.
> For deployment to a cluster, environment variables for the container are supplied directly helm in the form (--nbs.elasticsearch.url=) at container start-up. If we should be using expect environment variables instead, we'll need to parameterize a few more variables and modify existing ones like the urls.

The existing method of setting the urls variables will work.  I got into the habit of building configuration items like urls from other variable due to only parts of the url changing between environments.  Host and Port will change way more often for a JDBC connection string.

PR #111, opened 2023-02-17
CNPT-948 CNPT-946 CNFT1-187 fix phone,email search,add soundex

stevegsa:
Make the default operator `and` instead of `or` to require all the `-` and `@` separated values.  ie searching for 123-456-7890 could return phones with just 123 before (and appeared to be returning all phones when in actuality it was just phones in that area code).  This also allows searching by say area code 123 if that is the query string.

Added soundex search API capability.  Decided new analyzer wasn't needed in elasticsearch.  Legacy already has a populated soundex field.  Pass that through and key off of it in the api.  In the api require that the fname/lname searches satisfy at least one of a traditional or soundex match (logical OR ie `should`)

### DIFF
--- a/apps/modernization-api/build.gradle
+++ b/apps/modernization-api/build.gradle
@@ -82,6 +82,8 @@ dependencies {
 	// itext pdf generation
 	implementation "com.itextpdf:itextpdf:5.5.13.3"
 
+	implementation group: 'commons-codec', name: 'commons-codec', version: '1.15'
+
 	// test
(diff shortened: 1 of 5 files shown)


mpeels (changes requested):
Getting a couple test failures when running:
```
./gradlew :modernization-api:test -Dcucumber.filter.tags="@patient_data_search" --tests "RunCucumberTest"
```
image

mpeels (approved):

PR #117, opened 2023-02-21
Feature/cnft1 550 kafka consumer delete

adasika-cdc:


apps/modernization-api/src/main/java/gov/cdc/nbs/controller/PatientController.java
```diff
@@ -68,8 +68,8 @@ public Optional<Person> findPatientById(@Argument Long id) {
 
     @MutationMapping()
     @PreAuthorize(DELETE_PATIENT)
```
mpeels, line 70: It looks like legacy requires also requires the `View Patient` permission when doing deletes

Image from System Admin -> manage permissions screen
image

PR #119, opened 2023-02-22
CNFT1 629/create patient

adamloup-enquizit:
Adds creation of a patient as a result of a valid user requesting the creation of a patient.


apps/patient-listener/src/main/java/gov/cdc/nbs/patientlistener/service/PatientCreator.java
```diff
@@ -0,0 +1,125 @@
...
+                request.deceased(),
+                request.deceasedTime(),
+                request.maritalStatus(),
+                request.ethnicity(),
+                request.createdBy(),
+                request.createdAt()
+        );
+    }
+
+    private PatientCommand.AddName as(final PatientCreateRequest request, final PatientCreateRequest.Name name) {
```
mpeels, line 65: can we rename the `as` methods to `asName`, `asPostalAddresses`, etc to match the `PatientCreateRequestResolver` naming

adamloup-enquizit: Sure can

mpeels (approved):

PR #120, opened 2023-02-22
Patient search/cnft1 155 state names

jlee-enquizit:


apps/modernization-ui/src/pages/advancedSearch/components/patientSearch/AddressForm.tsx
```diff
@@ -2,74 +2,85 @@ import { Grid } from '@trussworks/react-uswds';
...
+                                name="state"
+                                render={({ field: { onChange, value } }) => (
+                                    <SelectInput
+                                        defaultValue={value}
+                                        onChange={onChange}
+                                        htmlFor={'state'}
+                                        label="State"
+                                        options={searchCriteria.states.map((state) => {
+                                            return {
+                                                value: state.codeDescTxt!,
```
mpeels, line 57: Based on what I'm seeing in the search API, you will need to use the value of `state.id`. However, this will display in the search as this: 
image

Not sure if this is a problem or if we need to change how those `Chip`s are displayed.

hclarkEnq (approved):

PR #127, opened 2023-03-01
Update gql generator to support 'extend type'

mpeels:
1. Change GraphQL code generator from `gqlg` to the migrated `gql-generator` npm package. This newer version supports the `extend type` syntax.
2. Move the Query declarations from the `queries.graphqls` file into separate files

Before running the new generator, you must run `npm i` to replace the old `gqlg` command


adamloup-enquizit:
When running `npm run generate` the `BasicErrorControllerService` seems to revert to it's previous state with the method signatures changing from `errorX` to `errorHtmlX` and the return types from `Record` to `ModelAndView`.

mpeels:
> When running `npm run generate` the `BasicErrorControllerService` seems to revert to it's previous state with the method signatures changing from `errorX` to `errorHtmlX` and the return types from `Record` to `ModelAndView`.

This seems to be an issue with swagger and how the `/v2/api-docs` are generated. Restarting the modernization-api seems to have a 50/50 chance of generating them one way or the other. Heres a partial diff from the different api-docs json.

image


One option is to just remove the generated BasicErrorControllerService, as it is not being used.

adamloup-enquizit (approved):
👍 <sup>👍<sup>👍</sup></sup>

PR #138, opened 2023-03-03
CNFT1-599 Morbidity Reports Pull Data For Profile View API

Emelias-Enquizit:


libs/database-entities/src/main/java/gov/cdc/nbs/repository/PersonRepository.java
```diff
@@ -7,14 +7,18 @@
...
-        @Query("SELECT coalesce(max(p.id), 0) FROM Person p")
-        Long getMaxId();
+	@Query("SELECT coalesce(max(p.id), 0) FROM Person p")
+	Long getMaxId();
 
-        Page<Person> findByIdIn(List<Long> ids, Pageable pageable);
+	Page<Person> findByIdIn(List<Long> ids, Pageable pageable);
+
+	@Query(value="SELECT p.person_uid from Person p where p.person_parent_uid=:personParentUid", nativeQuery=true)
+	List<Object[]> getPersonIdsByPersonParentId(@Param("personParentUid") Long personParentUid);
```
mpeels, line 23: Since `Person.person_uid` is a Long, you can return `List<Long>` here instead of `List<Object[]>`

apps/modernization-api/src/main/java/gov/cdc/nbs/service/EventService.java
```diff
@@ -793,6 +826,21 @@ public Page<Investigation> findOpenInvestigationsForPatient(Long patientId, Grap
...
+		List<Long> actIds = new ArrayList<Long>();
+		List<Object[]> results = personReposity.getPersonIdsByPersonParentId(patientId);
+		for (Object[] obj : results) {
+			subjectEntityIds.add((BigInteger) obj[0]);
+		}
+		List<Object[]> actIdResults = participationRepository
+				.findIdActUidByIdTypeCdAndIdSubjectEntityUidIn(Constants.REPORT_TYPE, subjectEntityIds);
+		for (Object[] objAct : actIdResults) {
+			actIds.add(((BigInteger) objAct[0]).longValue());
+		}
```
mpeels, line 841: Once the repo is updated to return a Long, this can be simplified to not have to handle the Object[]'s

libs/database-entities/src/main/java/gov/cdc/nbs/repository/PersonRepository.java
Emelias-Enquizit: Hi @mpeels  it is  returning  a list of Long it is just that the nativeQuery will return it as a list of Object[] array with each object array containing a row for results of the specific fields specified.  I initially had it as a List<Long> which threw some errors.

Emelias-Enquizit: @mpeels  updated. Thanks.

apps/modernization-api/src/main/java/gov/cdc/nbs/service/EventService.java
Emelias-Enquizit: @mpeels  updated. Thanks.

mpeels (approved):

PR #139, opened 2023-03-03
Cnft1 127 event search add unassigned field to case status and change "status" fields to multi-select

adasika-cdc:


adasika-cdc:
Screenshot 2023-03-03 at 12 23 09 AM

mpeels:
Is there a follow-up story to update the API to handle the changed format? 

If not then the [EventService](https://github.com/CDCgov/NEDSS-Modernization/blob/8e3ba09b91d794349450b9de439d2793f1368578/apps/modernization-api/src/main/java/gov/cdc/nbs/service/EventService.java#L566-L631) logic will have to be updated, as well as the [InvestigationFIlter](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/modernization-ui/src/generated/models/InvestigationFilter.ts) object to reflect the removed `includeUnassigned` boolean for CaseStatusus, NotificationStatus, and ProcessingStatus. These probably wont need to be separate objects anymore since they will would only have a single field.

adasika-cdc:
Hi Michael, 
I am not sure if there is a follow up to update the APIs here ~ my guess is there is

PR #145, opened 2023-03-06
On redirect set RecordStatus

mpeels:
Fix redirect missing RecordStatus value. Add redirect test to ensure search is successful

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/PatientFilter.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/PatientFilter.java
@@ -1,14 +1,14 @@
 package gov.cdc.nbs.graphql.filter;
 
 import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
-
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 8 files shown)


apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/PatientFilter.java
```diff
@@ -21,6 +21,11 @@
 @EqualsAndHashCode
 @JsonInclude(Include.NON_NULL)
 public class PatientFilter {
+    public PatientFilter(RecordStatus... recordStatus) {
```
adamloup-enquizit, line 24: This can still allow zero `recordStatus`.  it's tricky to enforce this with jackson.  Maybe something like
```java
public PatientFilter(RecordStatus required, RecordStatus...rest) {
...
}

public PatientFilter() {
  this(RecordStatus.ACTIVE);
}
```

This would introduce a default of `ACTIVE` though.

mpeels: I think a default of `ACTIVE` is fine. Updated to your recommendation

adamloup-enquizit (approved):
👍<sup>👍 <sup><sup>👍 </sup></sup>

PR #147, opened 2023-03-06
CNFT1-624 Finalize Demographic Details

Emelias-Enquizit:


apps/modernization-api/src/main/java/gov/cdc/nbs/config/KafkaConfig.java
```diff
@@ -37,38 +44,95 @@ public class KafkaConfig {
...
-    public NewTopic createPatientSearchTopic() {
-        if (!kafkaEnabled) {
-            return null;
-        }
-        return TopicBuilder.name(patientSearchTopic).partitions(topicPartitionCount).replicas(topicReplicationFactor)
-                .compact().build();
-    }
+	@Bean
+	public NewTopic createPatientSearchTopic() {
+		if (!kafkaEnabled) {
```
adamloup-enquizit, line 49: The  `@ConditionalOnProperty`  annotation could be used at the class level to enable/disable the entire configuration.  It would be something like `@ConditionalOnProperty(prefix = "kafka", name = "enabled", havingValue = "true")`.  It would include the Kafka Templates also so if you go that route you might want to separate the creation of the topics from the templates as two Configurations.

Emelias-Enquizit: Hi @adamloup-enquizit  for now  I would prefer to keep pulling kafkaEnabled flag from the application.yaml setting and  after these changes merge I can look into using the @condittionalOnProperty flag separating the creation of the topics  from the templates  in to a separate PR would those updates tested separately.

apps/modernization-api/src/main/java/gov/cdc/nbs/service/util/Constants.java
```diff
@@ -8,6 +8,13 @@ private Constants() {
 
     //Application ID
     public static final String APP_ID = "PATIENT-SEARCH";
+    
+    //Patient Update Types
+    public static final String UPDATE_GENERAL_INFO ="GeneralInformation";
+    
+    public static final String UPDATE_SEX_BIRTH="Sex&Birth";
+    
+    public static final String UPDATE_MORTALITY="Mortality";
```
mpeels, line 17: Instead of duplicating these constants here and in the patient-listener, we could move them into the `event-schema` library since they pertain to the kafka messaging.

apps/modernization-api/src/main/java/gov/cdc/nbs/controller/PatientController.java
```diff
@@ -81,8 +81,20 @@ public PatientCreateResponse createPatient(@Argument PatientInput patient) {
...
+    @PreAuthorize(FIND_AND_EDIT_AND_VIEW)
+    public PatientUpdateResponse updatePatientSexBirth(@Argument Long id, @Argument PatientInput patient) {
+        return patientService.updatePatientSexBirth(id, patient);
+    }
+    
+    @MutationMapping()
+    @PreAuthorize(FIND_AND_EDIT_AND_VIEW)
+    public PatientUpdateResponse updateMortality(@Argument Long id, @Argument PatientInput patient) {
+        return patientService.updateMortality(id, patient);
     }
```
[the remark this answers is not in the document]
Emelias-Enquizit: @mpeels  updated

apps/modernization-api/src/main/java/gov/cdc/nbs/service/util/Constants.java
Emelias-Enquizit: @mpeels updated

mpeels (approved):

PR #148, opened 2023-03-06
CNFT1 156/search pages skipping

adamloup-enquizit:
There were at least two issues found that could cause an error in the `findPatientsByFilter` Query resulting in search pages being "skipped" when clicked.

- Corrects how the locator id was being set for Entity Locators Participations
- Adds a Suffix converter for Elasticseach results that will disregard unknown suffix values instead of throwing an error.

The invalid Suffix values were most likely introduce by anonymization of the data.


mpeels (approved):

stevegsa (approved):
Excellent job!

PR #150, opened 2023-03-07
Cnft1 127 event search add unassigned field to case status

adasika-cdc:


apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/InvestigationFilter.java
```diff
@@ -32,23 +32,20 @@ public class InvestigationFilter {
...
         private List<NotificationStatus> statusList;
     }
 
 
     @Getter
     @Setter
     public static class ProcessingStatuses {
-        private boolean includeUnassigned;
         private List<ProcessingStatus> statusList;
     }
```
mpeels, line 50: Instead of having an Object to contain the Status list, we can move these Lists into the base InvestigationFilter

apps/modernization-ui/src/generated/schema.graphqls
```diff
@@ -220,6 +223,7 @@ enum InvestigationEventDateType {
 }
 
 enum InvestigationStatus {
+    UNASSIGNED
```
mpeels, line 226: I don't believe `UNASSIGNED` applies to Investigation status. I think they can only be `OPEN` or `CLOSED`

mpeels (approved):

PR #154, opened 2023-03-09
Fix broken test

mpeels:
Add a null / empty list check for EntityIds. Was causing a test failure when no EntityId was present on the Person object.

After fix:
`modernization-api:test results: SUCCESS (241 tests, 241 successes, 0 failures, 0 skipped)`


stevegsa (approved):
thanks!

PR #160, opened 2023-03-16
CNFT1-673/client date handling

adamloup-enquizit:
Searching from the home screen of NBS Classic with a Date of Birth value was causing the UI to render a blank page due
to an unexpected value for the Date of Birth field. This happened due to the Object -> JSON (Jackson) mapping performed by the encryption service. By default, Jackson serializing Instant to a number. The `DatePickerInput` on the client is expecting a string formatted as `mm/dd/yyyy`

Updating the encryption service to serialize Instant to the expected formatted string allowed the page to render
however, there were new issues.

- The value of the Date of Birth was not visible in the `DatePicker`
- The DOB chip contains the value as a full ISO 8601 value with date and time
- The Date of Birth value would change due to the original value from NBS being set to the UTC timezone.

To Resolve this the type of the `dateOfBirth` field on the `PersonFilter` was changed from `Instant` to `LocalDate`. In
order to accomplish this a new GraphQL Scalar needed to be introduced to model this value. The existing `Date` scalar
was renamed to `DateTime` and a new `Date` scalar was added.

Documentation was added to explain the Date/Time formats coming from `ui` -> `api` and `api` -> data-stores (MSSQL,
Elasticsearch).  

Testing:
1. Open the NBS home page
2. In simple search enter a date of birth
3. Click "Search"

Outstanding issue:

- The value of the Date of Birth is not visible in the `DatePicker` when routed from NBS Classic.  This also occurs during a browser refresh of the `advanced-search` page.


apps/modernization-api/src/main/java/gov/cdc/nbs/time/graphql/DateScalarConfig.java
```diff
@@ -0,0 +1,23 @@
...
+@Configuration
+public class DateScalarConfig {
+
+  @Bean
+  public RuntimeWiringConfigurer dateScalarConfigurer() {
+
+    return wiringBuilder -> wiringBuilder.scalar(
+        GraphQLScalarType.newScalar()
+            .name("Date")
+            .description("Java Instant as scalar.")
```
mpeels, line 17: Isn't it `Java LocalDate as scalar`?

adamloup-enquizit: you exposed my copy/paste

mpeels (approved):

PR #163, opened 2023-03-21
Cnft1 649 event search  clear all button takes to patient search

GangulaMounika:


adamloup-enquizit (changes requested):
There are changes to `DatePickerInput` that are causing merge conflicts.

PR #168, opened 2023-03-23
migrates Investigation to 'gov.cdc.nbs.investigation' package

mpeels:
Moves the Investigation logic to the "package by feature" layout.

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/InvestigationFilter.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/InvestigationFilter.java
@@ -4,6 +4,7 @@ import java.time.Instant;
 import java.util.List;
 
 import gov.cdc.nbs.message.enums.PregnancyStatus;
+import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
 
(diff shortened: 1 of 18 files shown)


apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/InvestigationFilter.java
adamloup-enquizit, line 1: Should this class also exist within the `investigation` package?

apps/modernization-api/src/test/java/gov/cdc/nbs/investigation/InvestigationResolverTest.java
```diff
@@ -0,0 +1,105 @@
...
+    void should_limit_max_page_size() {
+        Long patientId = 123L;
+        var requestedPageSize = maxPageSize + 1;
+        QueryException ex = null;
+        try {
+            resolver.findOpenInvestigationsForPatient(patientId, new GraphQLPage(requestedPageSize));
+        } catch (QueryException e) {
+            ex = e;
+        }
+        assertNotNull(ex);
```
adamloup-enquizit, line 87: Exceptions can be tested with `assertThatThrownBy` from assertj or the existence of the exception should fail the test.

```java
assertThatThrownBy(() -> resolver.findOpenInvestigationsForPatient(patientId, new GraphQLPage(requestedPageSize)))
.isInstanceOf(QueryException.class);
```
Other assertions can be chained if you wanting to test the message also.  I'd also argue that this is a test of a feature of the `GraphQLPage` object :upside_down_face:

apps/modernization-api/src/main/java/gov/cdc/nbs/graphql/filter/InvestigationFilter.java
mpeels: yep

adamloup-enquizit (approved):
🕵️

PR #170, opened 2023-03-30
Allows sandbox proxy to use local ui and/or backend servers

adamloup-enquizit:
Changes the sandbox reverse proxy to resolve the host name of the `modernization-api` and `modernization-ui` from environment variables.  This would allow routing from sandboxed reverse proxy to a locally running `modernization-api` and `modernization-ui` instance.

For example, if the `modernization-api` was started locally as follows;

```shell
./gradlew :modernization-api:bootRun --args='--server.port=9080'
```

then starting the reverse proxy by running this command from the `cdc-sandbox` folder

```shell
MODERNIZATION_API_SERVER=host.docker.internal MODERNIZATION_API_PORT=9080 docker compose up -d reverse-proxy
```

---

Details on the configuration (and how to reset to defaults) have been added to the [README](https://github.com/CDCgov/NEDSS-Modernization/tree/configurable_traefik#configuring-the-reverse-proxy-to-use-local-modernization-api).

### DIFF
--- a/cdc-sandbox/docker-compose.yml
+++ b/cdc-sandbox/docker-compose.yml
@@ -21,15 +21,15 @@ services:
       - FORCE_RESTORE
     networks:
       - nbs
-  wildfly: 
+  wildfly:
     build: ./
     container_name: wildfly
     depends_on:
       - nbs-mssql
-    ports: 
+    ports:
       - "9990:9990"
       - "7001:7001"
-      - "8787:8787" 
+      - "8787:8787"
     networks:
       - nbs
   reverse-proxy:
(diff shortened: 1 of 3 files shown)


cdc-sandbox/docker-compose.yml
```diff
@@ -41,15 +41,20 @@ services:
...
     networks:
       - nbs
+    environment:
+      - MODERNIZATION_UI_SERVER=${MODERNIZATION_UI_SERVER:-modernization-ui}
+      - MODERNIZATION_UI_PORT=${MODERNIZATION_UI_PORT:-80}
+      - MODERNIZATION_API_SERVER=${MODERNIZATION_API_SERVER:-modernization-api}
+      - MODERNIZATION_API_PORT=${MODERNIZATION_API_PORT:-8080}
     # volumes:
-    #   - ./proxy/:/etc/traefik/    
+    #   - ./proxy/:/etc/traefik/
```
JimBarrows, line 50: Is there a reason we're leaving commented out code in the file?  if so, can you document it in the file?  This is the same for all of the commented out code.

adamloup-enquizit: It was pre-existing but that doesn't mean I can remove it now

JimBarrows (changes requested):
If we leave commented out code in, then we need to document why it's commented out, and when to uncomment it.

adamloup-enquizit:
> If we leave commented out code in, then we need to document why it's commented out, and when to uncomment it.

Removed that commented out line.  I am not sure what the reason for commenting was as it was done for [PR 47](https://github.com/CDCgov/NEDSS-Modernization/pull/47)

JimBarrows (approved):

PR #172, opened 2023-03-31
Microservice for CRUD -questionbank

adasika-cdc:


apps/questionbank/gradlew
mpeels, line 1: gradle is already present in the project root directory, so we don't need it here as well

mpeels:
These changes will be bundled into https://github.com/CDCgov/NEDSS-Modernization/pull/179

PR #173, opened 2023-03-31
Patient profile integrations and UI elements for all tabs 

hclarkEnq:
This includes sorting
integrations
fixes in data entry and patient search


apps/modernization-ui/src/pages/addPatient/components/LeftBar/LeftBar.tsx
```diff
@@ -22,14 +22,14 @@ export const LeftBar = ({ activeTab }: any) => {
                     New patient
                 </h6>
             </div>
-            <div className="border-base-light grid-row flex-no-wrap border-top padding-y-1 margin-x-3 cursor-pointer width-full">
+            {/* <div className="border-base-light grid-row flex-no-wrap border-top padding-y-1 margin-x-3 cursor-pointer width-full">
```
JimBarrows, line 25: Is this commented out code left in for a reason?  If so can you add that reason to the comment.
The same for all the commented out code.

apps/modernization-ui/src/pages/patientProfile/Demographics.tsx
```diff
@@ -497,14 +574,46 @@ export const Demographics = ({ patientProfileData, handleFormSubmission, ethnici
 
             <div className="margin-top-6 margin-bottom-2 flex-row common-card">
                 <TableComponent
+                    handleAction={(type, data) => {
```
JimBarrows, line 577: This looks like common code, and functionality for a table component.  Can this be reused as a component?  Even if it is just for this page?

apps/modernization-ui/src/pages/patientProfile/components/DemographicDetails/DetailsIdentificationModa.tsx
```diff
@@ -0,0 +1,49 @@
+import { Grid } from '@trussworks/react-uswds';
```
JimBarrows, line 1: The file name is mispelled.

JimBarrows (changes requested):
Mostly questions, and a few minor nits :)

apps/modernization-ui/src/pages/patientProfile/Demographics.tsx
hclarkEnq: We are already getting this prop from the table component as we are using different pop ups for each table so we have to pass the modal ref, and the model ref should not be re usable for each modal that were creating

apps/modernization-ui/src/pages/patientProfile/Demographics.tsx
```diff
@@ -469,14 +515,45 @@ export const Demographics = ({ patientProfileData, handleFormSubmission, ethnici
...
+                            setRaceModal(false);
+                            setAddressModal(true);
+                            detailsAddressModalRef.current?.toggleModal();
+                        }
+                    }}
                     isPagination={true}
                     buttons={
                         <div className="grid-row">
-                            <ModalToggleButton modalRef={addAddressModalRef} opener className="display-inline-flex">
+                            <Button
```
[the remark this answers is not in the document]
hclarkEnq: Yes the button can be used as a component thats reusable but at the moment Im just using buttons from the USWDS I do agree with utilizing our own custom component for buttons but will have to implement everywhere where the button is being used and that can be done in a new ticket probably in the upcoming sprint, would be a good idea.

JimBarrows-Enquizit (approved):

PR #174, opened 2023-03-31
Cnft1 715 update multi select dropdown component

jlee-enquizit:
This updates the multi-select dropdown with the `react-select` component.
This enables:

- Clear all fields
- No default selection
- better UI

### DIFF
--- a/apps/modernization-ui/package.json
+++ b/apps/modernization-ui/package.json
@@ -26,6 +26,7 @@
         "react-refresh": "^0.14.0",
         "react-router-dom": "^6.3.0",
         "react-scripts": "5.0.1",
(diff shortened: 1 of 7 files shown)


adamloup-enquizit (commented):
This might not be covered by the ticket but the options of the `MultiSelectControl` are being cut off when outside of the scroll area.

image


Occurs in FIREFOX, Chrome 🆗  
Also, on Event Search with Event Type: Investigation when clicking on Jurisdiction the `Search` and `Clear all` buttons are moving upwards.

Default state:
image

After clicking Jurisdiction
image

adamloup-enquizit (approved):

PR #176, opened 2023-04-03
Migrate labreport to package

mpeels:
Moves the Lab Report logic to a 'labreport' package. Copies file layout of treatment package with a "Resolver" and "Finder" class.
Separates the query builder method into its own class to better facilitate testing


adamloup-enquizit (approved):
👍

adamloup-enquizit (approved):

PR #188, opened 2023-04-11
Maven version 3.8.7 not found

mpeels:
When building fresh, the following error occurs when building the Wildfly container:

image

Change to using the maven amazoncorretto 8 docker image resolves this issue


adamloup-enquizit (approved):
🐜 🏗️

PR #189, opened 2023-04-11
Elasticsearch memory default

mpeels:
When setting up the local environment, elasticsearch is using 6+ Gb of memory with no data. This change will set the default limit to 2Gb.

@szamfir, will this change affect the deployed environments?


szamfir:
LGTM

PR #194, opened 2023-04-17
Cnft1 726 backend task patient profile display shortened id 

adamloup-enquizit:
Implements the features of [CNFT10726](https://cdc-nbs.atlassian.net/browse/CNFT1-726) by adding the patient short identifier to the Patient Profile and Patient search results.  

An initial seed value is required to compute the Short identifier.  A configuration property, `nbs.identifier.person.initial`, has been added.  The default value is `10000000`.  The default is set in code so that a log message could be displayed to warn that the default has been specified.

A Patient Profile can also be queried using the short identifier.  The `patient` identifier can also be used to query for a Patient Profile, if both values are present the `patient` identifier takes precedence.  

```graphql

query profile($short: Int!) {
  findPatientProfile(shortId: $short) {
    id
    local
    shortId
  }
}

```


apps/modernization-api/src/test/java/gov/cdc/nbs/patient/PatientMother.java
```diff
@@ -11,37 +12,43 @@
...
+        this.patients = patients;
+        this.cleaner = cleaner;
+    }
+
+    void reset() {
+        this.cleaner.clean(settings.starting());
+        this.patients.reset();
+    }
+
+    public Person patient() {
```
mpeels, line 43: maybe this could be named `createPatient` or something, to better illustrate that it is being persisted.

apps/modernization-api/src/test/resources/features/patient/PatientProfile.feature
```diff
@@ -0,0 +1,18 @@
...
+    When a profile is requested by patient identifier
+    Then the profile is found
+
+  Scenario: I can retrieve a profile by short identifier
+    Given I have the authorities: "FIND-PATIENT" for the jurisdiction: "ALL" and program area: "STD"
+    When a profile is requested by short identifier
+    Then the profile is found
+
+  Scenario: I cannot retrieve a profile without proper authorities
+    Then the profile is not found
```
mpeels, line 18: still needs the `When a profile is request by patient identifier` right? Could add a second to do `When a profile is requested by short identifier` as well.

PR #201, opened 2023-04-19
create test db

mpeels:
1. Creates a `cdc-sandbox/test-db` folder with restore files for a stripped down NBS database for use in testing. The test-db is added to the cdc-sandbox docker-compose for use in running tests locally.
1. Fixes a couple of tests that were failing with the new db due to missing expected data.
1. Adds the `Build-modernization-test-db.yml` workflow to build and push the test-db image to ECR.

The next step will be to verify the image is pushed to ECR and is available, then update the tests to pull this image in a test container.

### DIFF
--- a/apps/modernization-api/src/test/resources/application-test.yml
+++ b/apps/modernization-api/src/test/resources/application-test.yml
@@ -32,7 +32,7 @@ spring:
         hbm2ddl:
           jdbc_metadata_extraction_strategy: individually
   datasource:
-    url: jdbc:sqlserver://${nbs.datasource.server}:1433;database=nbs_odse;encrypt=true;trustServerCertificate=true;
+    url: jdbc:sqlserver://${nbs.datasource.server}:1434;database=nbs_odse;encrypt=true;trustServerCertificate=true;
     username: sa
     password: fake.fake.fake.1234
   mvc:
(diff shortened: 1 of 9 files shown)


cdc-sandbox/docker-compose.yml
adamloup-enquizit, line 1: Should the test database be part of the sandbox compose file?  it should only ever be used for testing

apps/modernization-api/src/test/resources/application-test.yml
```diff
@@ -32,7 +32,7 @@ spring:
         hbm2ddl:
           jdbc_metadata_extraction_strategy: individually
   datasource:
-    url: jdbc:sqlserver://${nbs.datasource.server}:1433;database=nbs_odse;encrypt=true;trustServerCertificate=true;
+    url: jdbc:sqlserver://${nbs.datasource.server}:1434;database=nbs_odse;encrypt=true;trustServerCertificate=true;
```
adamloup-enquizit, line 35: `nbs.datasource.port` property maybe?  That was we can override it locally if we want to run tests against the local development database.  We may have to specify this value like we do with the elasticsearch test-container

adamloup-enquizit (commented):
Just a few questions about configuration and local testing.

adamloup-enquizit (approved):

PR #202, opened 2023-04-20
Remove cucumber test filter

mpeels:
Removes the Cucumber test filter so they are executed during sonar scan.


szamfir (commented):
Will let Adam handle approval. Otherwise it looks good to me, still the known issue of pulling and standing up the db from container repository.

stevegsa (dismissed):

adamloup-enquizit (approved):

adamloup-enquizit (approved):

PR #205, opened 2023-04-20
CNFT1-771 Text meet color contrast

jlee-enquizit:
Bumped the contrast on "No Data" fields in `patient-profile`--up to USWDS class .text-grey-50.

We could include a function to calculate text color based on contrast-ratio on the fly a la: 
https://www.w3.org/TR/AERT/#color-contrast

Would be up for discussing that.


stevegsa (approved):

mpeels:
Merged as part of https://github.com/CDCgov/NEDSS-Modernization/pull/228

PR #206, opened 2023-04-21
CNFT1-802 Support Ssn search without dashes

stevegsa:
for ssn (which isn't currently being used and is a single valued field attached to the person table)
1. Update NIFI indexing to strip non-digit characters off ssn.  
2. Update API to strip non-digit characters when searching by ssn.  
3. Add a partial data search by ssn feature test to prove it's not tokenized.

for identification (which is used for ssn and stored in rootExtensionText in the entity_id table alongs with other forms of id):
1. Update NIFI indexing to strip all non-alphanum characters off of identification
2. Update API to strip non-alphanum characters when searching by identification.  
3. Add a partial data search by identification feature test to prove it's not tokenized.
NOTE: there are over a dozen forms of identification.  stripping non-alphanum should be correct.  the front end and backend updater will prevent any alphas from getting into digit oriented ids so the strip doesn't have to do a type check


apps/modernization-api/src/main/java/gov/cdc/nbs/patient/PatientService.java
```diff
@@ -173,7 +173,13 @@ public Page<Person> findPatientsByFilter(PatientFilter filter, GraphQLPage page)
         }
 
         if (filter.getSsn() != null && !filter.getSsn().isEmpty()) {
-            builder.must(QueryBuilders.matchQuery(ElasticsearchPerson.SSN_FIELD, filter.getSsn()));
+            String allDigitSsn = filter.getPhoneNumber().replaceAll("\\D", "");
```
mpeels, line 176: `filter.getPhoneNumber` might not be correct here

mpeels (dismissed):

mpeels (approved):

PR #208, opened 2023-04-21
Use local test db image

mpeels:
Reduce elasticsearch test container memory usage. Use local test db until ECR config is resolved.

Pre-req to running tests is to build test-db image.


adamloup-enquizit (approved):
🐳

PR #211, opened 2023-04-22
CNFT2 161 beginning of questionnaire API

mpeels:
This is the very first iteration of a questionnaire API. The main reason for this PR is to get the initial configurations of the Question Bank microservice working. 

1. The cucumber test is working and generating a report
2. The h2 database is working as expected with Flyway initializing the tables on startup.
3. The GraphQL endpoint is functioning

Things TODO:
1. Authentication
2. Refine data models
3. Determine an approach for handling multiple GraphQL APIs, potentially leverage [Apollo Federation](https://www.apollographql.com/docs/federation/)


gradle/libs.versions.toml
```diff
@@ -39,20 +39,35 @@ cucumber-java = { module = "io.cucumber:cucumber-java", version.ref = "cucumber"
...
+junit-platform = { module = "org.junit.platform:junit-platform-suite" }
 
 [bundles]
 security = ["java-jwt", "spring-security"]
 jackson = ["jackson-core", "jackson-datatype-jsr310"]
 jpa = ["springBoot", "springBoot-jpa"]
 elasticsearch = ["spring-elasticsearch"]
 kafka = ["spring-kafka"]
-testing = ["assertj-core", "mockito-inline"]
+testing = ["assertj-core", "mockito-inline", "junit", "junit-platform"]
```
adamloup-enquizit, line 51: The `junit-platform` dependency was added to `modernization-api` because Cucumber requires the JUnit Platform.  The libs projects use the Jupiter Platform which is provided through the `jvm-test-suite` plugin.

We might need to add a separate bundle for `junit` and `junit-platform` or maybe add those two to the existing `cucumber` bundle.

apps/question-bank/build.gradle
```diff
@@ -2,13 +2,30 @@ import org.springframework.boot.gradle.plugin.SpringBootPlugin
...
-
+    id 'jacoco'
     alias libs.plugins.springBoot
     alias libs.plugins.springDependencies
 }
 
 group = 'gov.cdc'
 version = '0.0.1-SNAPSHOT'
+// Displays cucumber test results
+apply from: file('../../scripts/gradle/cucumber-logging.gradle')
```
adamloup-enquizit, line 13: We should be able to apply this to all sub-projects from the root.  It might end up displaying the test results for each module.

apps/question-bank/src/main/resources/graphql/questionnaire.graphqls
```diff
@@ -0,0 +1,88 @@
+type Query {
+  questionnaire(input: QuestionnaireContext): Questionnaire
+}
+
+input QuestionnaireContext {
```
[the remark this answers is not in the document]
mpeels: depends who you ask

adamloup-enquizit (dismissed):

adamloup-enquizit (approved):

PR #213, opened 2023-04-25
CNFT1-752 Space in two word states

jlee-enquizit:
Thought I fixed this already but here it is again.

There are still a few components that use the 'stateList' JSON from constants/states that should probably be refactored out--a couple of them such as Mortality.tsx aren't even referenced anywhere in the app.

Perhaps a separate ticket to remove this altogether should be made.


apps/modernization-ui/src/pages/advancedSearch/components/patientSearch/AddressForm.tsx
```diff
@@ -2,7 +2,7 @@ import { Grid } from '@trussworks/react-uswds';
 import { Input } from '../../../../components/FormInputs/Input';
 import { SelectInput } from '../../../../components/FormInputs/SelectInput';
 import { Controller } from 'react-hook-form';
-import { stateList } from '../../../../constant/states';
```
mpeels, line 5: Might as well just go ahead and delete the stateList. That way we are sure it wont pop up again

jlee-enquizit: @mpeels There's a couple of dead/unlinked pages that reference the stateList--I was hesitant to just delete them, preventing from removing it altogether:
- Mortality.tsx
- SexBirth.tsx
- AddIdentificationModal
- AddressModal

hclarkEnq (approved):

apps/modernization-ui/src/pages/advancedSearch/components/patientSearch/AddressForm.tsx
jlee-enquizit: Done

PR #214, opened 2023-04-28
CNFT1-809 changes state description from full name to two character name

adamloup-enquizit:
Implements features of [CNFT1-809](https://cdc-nbs.atlassian.net/browse/CNFT1-809) by changing the state description returned by the Patient Profile API's from the full state name to the two character state name.

The source of the state name is changed however the data structure has not changed.  There is no changed required in the `modernization-ui` for users of the API.

[CNFT1-809]: https://cdc-nbs.atlassian.net/browse/CNFT1-809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


stevegsa (approved):
Awesome!  Thanks for doing this!

PR #215, opened 2023-05-01
CNFT1-787 Patient search by relevance (closest match)

stevegsa:
**Why**: Currently we can search patients that have multiple names, their legal name, and names that are similar sounding (soundex) names with both first and last name.  The legacy system sorts by name.  However, when a patient has multiple names we must select a name to sort by and the modernized system uses legal name.  However if a match is found on a secondary or similar sounding name the sorting is counterintuitive and if you search for `Doe` and `John Doe` has another name `John America`.  Sorting will put `John America` first if that is his legal name.  Thus, we must provide a more search friendly result namely relevance search (like Google) which uses scoring where individual fields are boosted in terms of relevance and effect overall score.
**How**: Boosting can be accomplished in the index which results in faster searches but we need to reindex everytime we change boosting factors or in the query which is slower but doesn't need the reindex.  We will add boosting in the search query on the fly.  The boosting will be as follows.

Note~  This is our first cut at relevance scoring.  All fields must be assessed in terms of relevance in the future because boosting these must be in the context of every other field but for now let's ensure these are in order.

1. Legal Name: Boost 2.0
2. Additional Name (in addition to a legal name): Boost 1.0
3. Soundex Name: Boost 0.5 (aka last)

For testing always include 3 similarly named patients in the randomly generated ones from personmother: a soundex name, a legal name, a secondary name.

**Still working on some failed specs***


mpeels:
When running tests I receive a failure for the "ends with a space" tests:
image
image

stevegsa:
> When running tests I receive a failure for the "ends with a space" tests: image image
yes. i noted the specs were failing in the description.  hardcoding those 3 testable patients into the generated random patients proved to be cumbersome so I decoupled it totally from the original feature test.

PR #216, opened 2023-05-01
Cnft1 786 hide export and print buttons patient search UI

GangulaMounika:
CNFT1-786: hide export and print buttons patient search UI,
CNFT1-791: Inactive Sort Button for One or Zero rows Should be Disabled Programmatically and Visibly
CNFT1-793: Patient Search- Display Logic- Closest Match- UI- Sort By Option


apps/modernization-ui/src/generated/schema.graphqls
```diff
@@ -1480,6 +1480,7 @@ type OrganizationParticipation {
 enum SortField {
   lastNm
   birthTime
+  relevance
```
adamloup-enquizit, line 1483: This file is generated based on the schema files defined in `modernization-api`.  If you add `relevance` [here](https://github.com/CDCgov/NEDSS-Modernization/blob/951b70200cb5890f1338f1e5bf27e87c79083f1f/apps/modernization-api/src/main/resources/graphql/shared.graphqls#L47) it will populate the generated `schema.graphqls` file.

The graphql schema and client files are generated by using the `npm run generate` command in the `modernization-ui` while the `modernization-apu` is running.

adamloup-enquizit (changes requested):

adamloup-enquizit (approved):
👍

PR #218, opened 2023-05-02
CNFT1 805 event date bug

mpeels:
[CNFT1-805](https://cdc-nbs.atlassian.net/browse/CNFT1-805)
1. Fix date formats for `Event Date` searches
2. Fix clear all not clearing all inputs
3. Fix some `chip` close issues where the Event Id searches weren't clearing the dependent fields

"When we get time" it would be nice to further split the logic for the Investigation / Lab report searches so there are fewer methods that have to handle both event types.

[CNFT1-805]: https://cdc-nbs.atlassian.net/browse/CNFT1-805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-ui/src/generated/models/EventId0.ts
```diff
@@ -0,0 +1,22 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+export type EventId0 = {
```
adamloup-enquizit, line 5: Is this an Investigation Event Id?

mpeels: yea, renamed one to `InvestigationEventId`, and the other to `LabReportEventId` for clarity

hclarkEnq (approved):

PR #223, opened 2023-05-05
Update trussworks

mpeels:
1. Update Trussworks lib to `4.2.1`
2. Fix NavBar button click adding white background
Before:
image
After:
image
3. Fix `Add New` dropdown 
Before:
image
After:
image
4. Fix `Sort` dropdown
Before:
image
After:
image


adamloup-enquizit (approved):
semi unrelated:  should we be using `rem` over `px` for margins and padding?

mpeels:
> semi unrelated: should we be using `rem` over `px` for margins and padding?

yea, we should.

PR #225, opened 2023-05-08
Cnft1 852 patient profile investigation navigation

adamloup-enquizit:
Implements features of [CNFT1-852](https://cdc-nbs.atlassian.net/browse/CNFT1-852) by putting in place a foundation for allowing navigation to Classic NBS from Modernized NBS and back again.  Redirected requests from the Modernized Patient Profile to Classic NBS will contain a `Return-Patient` cookie that contains the identifier of the Patient for the current Patient Profile.  The value of the cookie will be used to resolve the Patient Identifier when a `Return to File: Events` or `Return to File: Summary` link is clicked.

*Requires rebuild of the Traefik reverse-proxy container*  
*Routing will only work in a local development environment*

- Encapsulates Authentication and Cookie handling for Redirects to Classic NBS
- Adds endpoint `/nbs/api/profile/{patient}/investigation` to prepare the classic session for adding an Investigation and then returns the Classic url in the `Location` header of the response.
- Adds endpoint `/nbs/api/profile/{patient}/investigation/{investigation}` to prepare the classic session for viewing an Investigation and then returns the Classic url in the `Location` header of the response.  
- Add endpoint `/nbs/api/profile/{patient}/investigation/{investigation}/compare/{other}` to prepare the classic session for comparing two Investigations then returns the Classic url in the `Location` header of the response.
- Allows the `/nbs/redirect/patientProfile` endpoint to read either the `MPRUid` or the `uid` query parameter as the identifier of a Patient.
- Adds `ClassicLink` and `ClassicButton` components to encapsulate the interaction withe a redirector endpoint.

[CNFT1-852]: https://cdc-nbs.atlassian.net/browse/CNFT1-852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/java/gov/cdc/nbs/controller/RedirectController.java
```diff
@@ -19,92 +18,72 @@
...
+    )
+    public void preparePatientDetails(HttpServletRequest request) {
+        String url = wildFlyUrl + "/nbs/HomePage.do?method=patientSearchSubmit";
+        // copy cookie header that contains the JSESSIONID from the original request
+        HttpHeaders headers = new HttpHeaders();
+        headers.add("cookie", request.getHeader("cookie"));
+        // send an empty search request to wildfly
+        var emptySearchRequest = RequestEntity.get(url).headers(headers).accept(MediaType.ALL).build();
+        restTemplate.exchange(emptySearchRequest, String.class);
+    }
```
mpeels, line 87: Is this still being used anywhere? Since we are building out the `Add Patient` and `Patient Details` pages, we might be able to remove this.

adamloup-enquizit: There are a few places in the client that are using this endpoint still.  It will be phased out after [CNFT1-853](https://cdc-nbs.atlassian.net/browse/CNFT1-853) and [CNFT1-854](https://cdc-nbs.atlassian.net/browse/CNFT1-854).  There will be PR's for these stories soon, the latter will remove `/preparePatientDetails`

mpeels (approved):

PR #226, opened 2023-05-09
CNFT1-683 CNFT1-823 Administrative and Patient Name APIs

stevegsa:
**Why**: Administrative API (update since it's a field on person) and Patient Name API (add/update/delete) since it's a 1:many on person.


apps/patient-listener/src/main/java/gov/cdc/nbs/patientlistener/request/update/PatientUpdateRequestHandler.java
```diff
@@ -104,6 +108,89 @@ public void handlePatientSexAndBirthUpdate(final UpdateSexAndBirthData data) {
...
+    var person = findPerson(patientId, requestId);
+    patientUpdater.update(person, new DeleteNameData(patientId, personNameSeq, requestId, userId, null));
+    updateElasticsearchPatient(person);
+    statusProducer.successful(
+        requestId,
+        "Successfully deleted patient name",
+        userId
+    );
+  }
+
```
mpeels, line 193: I don't believe there is any test coverage for these new methods.

stevegsa: ah i thought coverage calcs were wrong! good catch...

mpeels (approved):

PR #231, opened 2023-05-11
storybook update, card check for data, spelling mistake

hclarkEnq:


mpeels (dismissed):

apps/modernization-ui/src/components/Table/Table.tsx
```diff
@@ -191,7 +191,7 @@ export const TableComponent = ({
                               <tr key={row}>
                                   {item.tableDetails.map((detail: TableDetail, column: number) =>
                                       detail.title ? (
-                                          detail.title === 'Not available yet' ? (
```
adamloup-enquizit, line 194: This was displaying the `Not available yet` fields in a certain way without evaluating anything else on the detail.  If we do not have to support displaying `Not available yet` any more this check can be removed.  `No data` should be displayed by this component automatically if the title is not present.

apps/modernization-ui/src/pages/patient/profile/contact/PatientNamedByContactTable.spec.tsx
```diff
@@ -49,14 +49,14 @@ describe('when the patient has not been named by a contact', () => {
         }
     };
 
-    it('should display Not Available', async () => {
+    it('should display No data', async () => {
```
adamloup-enquizit, line 52: ❤️

adamloup-enquizit (approved):
📖 📚

PR #232, opened 2023-05-11
CNFT2-407 Implement JWT authentication for question bank

Emelias-Enquizit:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/config/WebSecurityConfig.java
```diff
@@ -1,26 +1,76 @@
...
+	    private final ObjectMapper mapper;	
+	    @Value("${spring.graphql.path: graphql}")	
+	    private String graphQLEndpoint;
 
     @Bean
     public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
         return http.authorizeRequests()
-                .anyRequest()
-                .permitAll()
+        		 .antMatchers("/graphql", "/encryption/*")
```
mpeels, line 40: can remove the `"/encryption/*"` matcher. The encryption endpoints exist only on the modernization-api and are used to encrypting query parameters.

apps/question-bank/src/main/resources/application.yml
```diff
@@ -1,8 +1,13 @@
 nbs:
   datasource:
     server: localhost
-
-
+  security:
+    tokenSecret: qJ5+oDurrJnxfaAz6h4Eg1saP7NSeHuQQ/Yq4vh+mbi4B4QYThc/M+33VtPJ5O2A
+    tokenIssuer: gov.cdc.nbs
+    tokenExpirationMillis: 1800000 # 30 minutes
+    parameterSecret: Yq4vh+mbi4B4QYThc/M+33VtPJ5O2Adc
```
mpeels, line 8: i think `parameterSecret` is only used for the encryption endpoint, prob don't need it.

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/config/WebSecurityConfig.java
Emelias-Enquizit: will updated

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/controller/UserController.java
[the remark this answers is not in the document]
Emelias-Enquizit: @mpeels  Oh okay, this was added so that question-bank could be run separately on port 8080 and handle login requests individually. With this file I am able to start question-bank solo and hand login requests but sure it can be removed.

mpeels (approved):

PR #234, opened 2023-05-11
CNFT1-857/ Patient Profile Documents Navigation

adamloup-enquizit:
Implements features of [CNFT1-857](https://cdc-nbs.atlassian.net/browse/CNFT1-857) by adding an endpoint to support viewing a Document from the Modernized Patient Profile.  Documents can only be viewed from the Patient Profile.  Returning from the Classic NBS Document uses the redirection that is already in place for `ContextAction=ReturnToFileEvents`

*Routing will only work in a local development environment*

- Adds the endpont GET /nbs/api/profile/{patient}/document/{identifier} which prepares the Classic NBS Session to view a Document and then returns a redirect response with the url the client will navigate to  view a Document in Classic NBS. The identifier of the patient is added as the value of the Returning-Patient cookie.

[CNFT1-857]: https://cdc-nbs.atlassian.net/browse/CNFT1-857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/pages/patient/profile/document/PatientProfileDocuments.spec.tsx
+++ b/apps/modernization-ui/src/pages/patient/profile/document/PatientProfileDocuments.spec.tsx
@@ -3,13 +3,12 @@ import { render } from '@testing-library/react';
 import { FindDocumentsForPatientDocument } from 'generated/graphql/schema';
 
 import { MockedProvider } from '@apollo/client/testing';
-import { PageProvider } from 'page';
 
 describe('when rendered', () => {
     it('should display sentence cased document headers', async () => {
         const { container } = render(
             <MockedProvider addTypename={false}>
-                <PatientProfileDocuments nbsBase={'base'} pageSize={1}></PatientProfileDocuments>
+                <PatientProfileDocuments pageSize={1}></PatientProfileDocuments>
             </MockedProvider>
         );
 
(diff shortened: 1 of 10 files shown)


apps/modernization-ui/src/pages/patient/profile/document/PatientProfileDocuments.spec.tsx
```diff
@@ -114,10 +113,6 @@ describe('when at least one document is available for a patient', () => {
         const dateCreated = await findByText(/10\/07\/2021/);
 
         expect(dateCreated).toHaveTextContent('10:01 AM');
```
mpeels, line 115: Thought this might fail on non EST systems. Changed my timezone and it still passes.

mpeels: ahhh [global-setup.test.js](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/modernization-ui/global-setup.test.js)

mpeels:
Seems to be some mismatch in the logic for displaying the documents in Classic™ and the modernized system.

Classic™ View:
image

Modernized view:
image

Results of clicking a doc link:
image

mpeels:
Found the issue -- the local dev database is missing the `effective_time` column for the `NBS_document` table. All looks good after updating the db

```sql
ALTER TABLE nbs_odse..nbs_document
    ADD effective_time DATETIME NULL;
```

mpeels (approved):

PR #241, opened 2023-05-16
CNFT2 211 question bank base

mpeels:
## Overview
Creates the base objects that will be used in the new page builder. Sets up the initial database entities. 
There are a few key things that we want to accomplish. 
1. Allow grouping of questions. These groups should be re-usable in multiple pages
2. Remove reporting information from question table (will be implemented later as a collaboration with the Data Access team)
3. Remove messaging information from question table (will be implemented later as a collaboration with the EpiSync team)

## Notes
1. The rules table has been created by no columns added. 
2. Not all required fields are created. Will need to add additional columns as we build out features and refine requirements
3. A questionnaire is a list of either individual display elements or groups of display elements.
4. Questionnaires can be associated with multiple conditions.

image


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/questionnaire/QuestionnaireMapper.java
```diff
@@ -0,0 +1,123 @@
...
+     * Converts a {@link gov.cdc.nbs.questionbank.entities.DisplayElementEntity DisplayElementEntity} to its
+     * corresponding {@link gov.cdc.nbs.questionbank.questionnaire.model.Questionnaire.Element Questionnaire.Element}.
+     * 
+     * @param element
+     * @return
+     */
+    Questionnaire.DisplayElement toDisplayElement(DisplayElementEntity element) {
+        if (element instanceof TextEntity t) {
+            return new Questionnaire.Text(t.getId(), t.getText());
+        } else if (element instanceof TextQuestionEntity tq) {
```
Emelias-Enquizit, line 65: could potentially be converted to switch case

Emelias-Enquizit:
The attached image to the pr: https://user-images.githubusercontent.com/109251240/238670733-41d12906-64c8-4542-85e0-e05eecbbfc80.png could  be useful as a confluence document ( if does not exist already) , documenting the changes to the question bank model as it develops.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/questionnaire/QuestionnaireMapper.java
mpeels: Tried that with a couple approaches:
image
This works and we could possibly make the "type" field public on the different entities as to not use "text", but its a bit longer than the if statement and involves doing the class casting separately.

Ideally we would use the new pattern matching for switch statements, but it is not released as of java 17

https://www.atlantbh.com/pattern-matching-for-instanceof-and-switch-in-java/


If I overlooked a better approach let me know

Emelias-Enquizit: No that is fine if it will make it unnecessarily longer and more complex your current solution is the better option. And  yeah that new pattern matching for switch statements feature currently in preview looks cool and will definitely be helpful.

Emelias-Enquizit (approved):

PR #246, opened 2023-05-19
Move Auth entites to authentication lib

mpeels:
Moves authentication related entities into `authentication` library.


adamloup-enquizit (approved):
🔐

PR #248, opened 2023-05-22
CNFT1-846 Patient Search- By Short ID

stevegsa:
**How** Convert Short id to regular patient id (prefix + offsetId + suffix) then use that in the exact matcher.

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/patient/PatientService.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/patient/PatientService.java
@@ -76,6 +76,12 @@ public class PatientService {
 
     @Value("${nbs.max-page-size: 50}")
     private Integer maxPageSize;
+    @Value("${nbs.uid.suffix: GA01}")
+    private String patientIdSuffix;
+    @Value("${nbs.uid.seed: 10000000}")
+    private long patientIdSeed;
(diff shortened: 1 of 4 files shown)


apps/modernization-api/src/main/java/gov/cdc/nbs/patient/PatientService.java
```diff
@@ -141,7 +149,24 @@ public Page<Person> findPatientsByFilter(PatientFilter filter, GraphQLPage page)
...
-            builder.must(QueryBuilders.matchQuery(ElasticsearchPerson.LOCAL_ID, filter.getId()));
+            if (Character.isDigit(filter.getId().charAt(0))) {
+                try {
+                    long shortId = Long.parseLong(filter.getId());
+                    PatientIdentifierSettings settings = new PatientIdentifierSettings(
+                        patientIdPrefix,
+                        patientIdSeed,
+                        patientIdSuffix
+                    );
+                    PatientLocalIdentifierResolver resolver = new PatientLocalIdentifierResolver(settings);
```
adamloup-enquizit, line 160: Should be able to inject the `PatientLocalIdentifierResolver` directly instead of having to create it each time.

adamloup-enquizit (commented):
Nice!  This works great, I had one note about the `PatientLocalIdentifierResolver` in PatientService.

PR #257, opened 2023-05-26
CNFT2-468 Api Delete Question

Emelias-Enquizit:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/QuestionRepository.java
```diff
@@ -0,0 +1,24 @@
...
+import gov.cdc.nbs.questionbank.entities.DisplayElementEntity;
+import gov.cdc.nbs.questionbank.entities.VersionId;
+
+public interface QuestionRepository extends JpaRepository<DisplayElementEntity, VersionId> {
+
+	@Transactional
+	@Modifying
+	@Query("UPDATE DisplayElementEntity SET active=:questionActive where id=:questionId")
+	public int deleteQuestion(@Param("questionId") UUID questionId,
+			@Param("questionActive") boolean questionActive);
```
mpeels, line 22: since the method is named `deleteQuestion`, you could skip passing in the `questionActive` param and always just set it to `false`

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/DeleteQuestionService.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.RequiredArgsConstructor;
+
+@Service
+@RequiredArgsConstructor
+public class DeleteQuestionService {
+
+	@Value("${kafkadef.topics.questionbank.deletestatus}")
+	private String questionDeleteStatusTopic;
+
+	private final KafkaTemplate<String, DeleteQuestionRequest> kafkaQuestionDeleteTemplate;
```
mpeels, line 25: could rename the `DeleteQuestionRequest` to something like `QuestionDeletedEvent` since its no longer a "request" thats looking for something to process it. Its now just a notification that an event took place

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/QuestionRepository.java
Emelias-Enquizit: sure will update.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/DeleteQuestionService.java
Emelias-Enquizit: sure will update.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/QuestionRepository.java
Emelias-Enquizit: updated.

mpeels (approved):

PR #260, opened 2023-05-30
Unzip restore

mpeels:
The "new" restore file contains a subfolder, `restore` and contains `.mdf` and `.ldf` files.  Added hopefully clearer text for the restore instructions as well as a sample command to unzip directly into the proper directory.


lramakr (approved):
LGTM

PR #268, opened 2023-06-02
CNFT1-1025/ Patient Delete

adamloup-enquizit:
Implements features of [CNFT1-1025](https://cdc-nbs.atlassian.net/browse/CNFT1-1025) by ensuring that a patient is deleted for the patient delete response is returned.  The publishing of Patient Events will be handled by another PR.

The `deletePatient` mutation was change to return the result of deletion as a `PatientDeleteFailed` or a `PatientDeleteSuccessful` response.

```graphql
mutation deletePatient($patient: ID!){
    deletePatient(patient: $patient){
      __typename
        ... on PatientDeleteFailed {
          patient
          reason
        }
        ... on PatientDeleteSuccessful {
          patient
        }
        
    }
}
```

A successful patient delete will return a `PatientDeleteSuccessful` response.  The patient data is not deleted from the database, the record status is changed to `LOG_DEL`.

```json
{
  "data": {
    "deletePatient": {
      "__typename": "PatientDeleteSuccessful",
      "patient": 10055289
    }
  }
}
```

A Patient can only be deleted if the patient has no revisions (revisions are created when a patient is associated with events).  When attempting to delete a patient that has revisions the response will be a `PatientDeleteFailed`

```json
{
  "data": {
    "deletePatient": {
      "__typename": "PatientDeleteFailed",
      "patient": 10054292,
      "reason": "Cannot delete patient with Active Revisions"
    }
  }
}
```

[CNFT1-1025]: https://cdc-nbs.atlassian.net/browse/CNFT1-1025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/java/gov/cdc/nbs/patient/delete/PatientDeleteController.java
```diff
@@ -1,23 +1,56 @@
...
+    PatientDeleteResult delete(@Argument long patient) {
+
+        NbsUserDetails user = SecurityUtil.getUserDetails();
+
+        RequestContext context = new RequestContext(user.getId(), Instant.now(this.clock));
+
+        try {
+
+            deleter.delete(context, patient);
+            this.indexer.index(patient);
```
stevegsa, line 44: is this deleting the document in elastic or reindexing it?

stevegsa (approved):
Looks good!

apps/modernization-api/src/test/resources/features/patient/PatientDelete.feature
```diff
@@ -1,22 +1,77 @@
...
+  Background:
+    Given I have a patient
+    And I am logged into NBS
+
+  Scenario: I can delete a patient that is not associated with events
+    Given I have the authorities: "DELETE-PATIENT,VIEW-PATIENT" for the jurisdiction: "ALL" and program area: "STD"
+    When I delete the patient
+    Then the patient is deleted
+
+  Scenario Outline: I delete a patient without the proper permissions
```
mpeels, line 13: Think this should just be `Scenario` instead of `Scenario Outline`

apps/modernization-api/src/main/java/gov/cdc/nbs/patient/delete/PatientDeleteController.java
adamloup-enquizit: It's re-indexing it, the status changes to `LOG_DEL` when a patient is deleted.  There is a use case where certain users can search over deleted patients

stevegsa: Cool

apps/modernization-api/src/test/resources/features/patient/PatientDelete.feature
adamloup-enquizit: Either should work.  Intellij shows an error if you have a Scenario with examples but it still runs.  The test shows up in the report if set as a `Scenario` or a `Scenario Outline`

PR #270, opened 2023-06-05
CNFT1-1035 Incrementally index newly added patients in Nifi

stevegsa:
**Why**: Newly added patients were not being indexed in Nifi
**How**: When Nifi was pulling records from the database if there was only 1 (greater than the latest person_parent_uid) nifi would create a flowfile containing a Hash and not an Array of 1 hash.  This was breaking the scripts (2) that were expecting an Array.
ie:
```
expected:
    [{"person_parent_uid": 10000001},{"person_parent_uid": 10003004},...
received:
    {"person_parent_uid": 10054282}
```

Improvements made:
- Fixed both scripts that were expecting a hash
- Added stacktrace information to log line numbers where scripts blowup
- Throttled up scheduling from 60 seconds to 1 second to check for new records.  This can be done because we have an index on person_parent_uid (see attached query plan) otherwise it would have to do a db full scan (potentially on 10s of millions of records)
Screenshot 2023-06-05 at 12 33 22 AM
- removed indexer for generic tables (not currently being used).  For some reason even though we save it as "off" it comes on in production and adversely effects indexing performance.  We have a copy of the templates if we decide to use it.

To verify:
- clear state on first processor for patient ingestion
- verify bulk indexing
- add new patient
- verify incremental indexing

Future improvements:
- This only handles incremental adds.  Updating a patient is more complicated to do performance wise because last modified is not indexed and will do a full scan of potentially 10s of millions of records with each check.
- Lab Report and Event currently check last modified but are not throttled to 1 sec.  They will need to be revisited also.


adamloup-enquizit (approved):
- Added a new patient with NiFi turned off.
- Turned on NiFi and ensured that the steps were updated
- Verified that the Patient added while NiFi was off was now searchable
- Added a new patient
- Verified that the patient was now searchable

PR #274, opened 2023-06-06
CNFT1-1043-The-field-at-path-findAllAssigningAuthorities-content-0-code_short_desc_txt-was-declared-as-a-non-null-type-but-the-code-involved-in-retrieving-data-has-wrongly-returned-a-null-value

JimBarrows-Enquizit:


stevegsa (approved):
Looks good!

PR #280, opened 2023-06-09
Id generator lib

mpeels:
Moves the Id generation logic into a `id-generator` lib.


libs/id-generator/src/main/java/gov/cdc/nbs/id/IdGeneratorService.java
```diff
@@ -60,6 +57,10 @@ public static class GeneratedId {
         private Long id;
         private String prefix;
         private String suffix;
+
+        public String toLocalId() {
```
adamloup-enquizit, line 61: genus

adamloup-enquizit (approved):
🆔 🔐

PR #283, opened 2023-06-12
CNFT1-1024-Complete-update-function-by-writing-to-the-database

JimBarrows-Enquizit:


apps/modernization-api/src/main/java/gov/cdc/nbs/patient/PatientService.java
```diff
@@ -80,31 +73,43 @@ public class PatientService {
...
+    private EntityLocatorParticipationRepository entityLocatorParticipationRepository;
+
+
 
     private <T> BlazeJPAQuery<T> applySort(BlazeJPAQuery<T> query, Sort sort) {
         var person = QPerson.person;
         if (sort == null) {
             // if no sort provided, default sort by lastNm then Id
             return query.orderBy(person.lastNm.asc().nullsLast())
-                .orderBy(person.id.desc().nullsLast());
```
stevegsa, line 95: I love standardizing on one code format but it seems we all have a separate one.  I think @mpeels did some work on this so we'd pull the same params.  It would make it easier to diff.

apps/modernization-api/src/main/java/gov/cdc/nbs/patient/PatientService.java
```diff
@@ -412,36 +417,162 @@ private void addRecordStatusQuery(Collection<RecordStatus> recordStatus, BoolQue
...
+            person.update(addName);
+            return personRepository.save(person);
+        });
         return sendPatientEvent(event);
     }
 
     public PatientEventResponse updateAdministrative(AdministrativeInput input) {
         var user = SecurityUtil.getUserDetails();
         var event = AdministrativeInput.toRequest(user.getId(), getRequestId(), input);
+        personRepository.findById(input.getPatientId()).map(person -> {
```
stevegsa, line 510: sonarcloud tagging these maps

adamloup-enquizit: `map` isn't terminal so they aren't doing anything, will have to use `ifPresent`

PR #284, opened 2023-06-12
Cnft1 1010 update ethnicity card hide spanish origin

GangulaMounika:
CNFT1-1011- Correct Spelling on Patient Search- "Superceded -> Superseded,
CNFT1-1010- Update Ethnicity Card- Hide "Spanish Origin" behind selection of "Hispanic or Latino",
CNFT1-980- UI Updates- Patient Profile- Display both Type/Use for Phone and Address Fields,
Integration of View Patient after saving New Patient

### DIFF
--- a/apps/modernization-ui/src/components/DemographicsForm/Ethnicity.tsx
+++ b/apps/modernization-ui/src/components/DemographicsForm/Ethnicity.tsx
@@ -4,6 +4,7 @@ import { Button, ButtonGroup, Grid } from '@trussworks/react-uswds';
 import { SelectInput } from '../FormInputs/SelectInput';
 import { formatInterfaceString } from '../../utils/util';
 import { SearchCriteriaContext } from '../../providers/SearchCriteriaContext';
+import { BaseSyntheticEvent, useState } from 'react';
 
 export const EthnicityForm = ({ setEthnicityForm }: any) => {
     const methods = useForm();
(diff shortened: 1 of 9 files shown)


apps/modernization-ui/src/components/DemographicsForm/Ethnicity.tsx
```diff
@@ -13,6 +14,13 @@ export const EthnicityForm = ({ setEthnicityForm }: any) => {
         setEthnicityForm();
     };
 
+    const [selectedEthinicity, setSelectedEthinicity] = useState<'2135-2' | '2186-5' | 'UNK' | 'NOT-TO-ANS' | ''>('');
```
mpeels, line 17: I'm not sure we can set the type like this as different NBS installations could have different / additional ethnicity options.

apps/modernization-ui/src/generated/graphql/schema.ts
```diff
@@ -2214,7 +2214,7 @@ export enum RecordStatus {
   Active = 'ACTIVE',
   Inactive = 'INACTIVE',
   LogDel = 'LOG_DEL',
-  Superceded = 'SUPERCEDED'
+  Superseded = 'SUPERSEDED'
 }
```
mpeels, line 2218: Anything in the `/generated` folder is auto generated based on the Java API declaration. The `SUPERCEDED` value corresponds to what is stored in the database and shouldn't be changed.

mpeels (changes requested):
The underlying enum will need to stay the same and just display the new spelling in the UI.

apps/modernization-ui/src/components/DemographicsForm/Ethnicity.tsx
adamloup-enquizit: Agreed, I think this is good for now because this can ultimately be handled by the backend returning a property indicating that detailed ethnicities exist.

mpeels:
Duplicate of https://github.com/CDCgov/NEDSS-Modernization/pull/286

PR #293, opened 2023-06-14
Removes the push trigger for sonar

adamloup-enquizit:
Removes the push trigger from the Sonar workflow.  Frequent pushes (good practice!) with many developers is causing the workflow actions to queue for long periods of time.


szamfir:
Instead of removing all pushes we can just limit the interaction to the main branch. By doing something like this:

on:
  push:
    paths:
    branches:
      - main

szamfir (approved):
LGTM thank you!

PR #295, opened 2023-06-14
Fix test and sonar bug

mpeels:
1. Fix the patient search "ends with space" test that was sporadically failing.
1. Fix some things sonar was complaining about 
a. Return of `.map` not being used
b. Lack of tests in test file
c. Lambda in test having multiple invocations that could throw an exception


adamloup-enquizit (approved):
👍

PR #296, opened 2023-06-14
Fix/remove request

adamloup-enquizit:
Prevents the Patient Profile mutations from posting the PatientRequest kafka event and updates the tests.


stevegsa (dismissed):
LGTM

PR #303, opened 2023-06-15
Cnft1 351 update mortality integration

hclarkEnq:


adamloup-enquizit (approved):
Seems to function fine.  Only issues I am seeing are from the API

PR #304, opened 2023-06-16
CNFT1-1077 Update Nifi Indexing w/Update of Patient Demographic fields

stevegsa:
**Why**: We need to update elasticsearch not just for new patient indexing but also for any changes associated with a patient/person (ie changes to the associated tables).  

**How**: The strategy for keeping elasticsearch up to date is a two fold approach whereby kafka will be notified of changes and reindex documents on the fly but also nifi will have a recurring query that will check for any changes in the db since the last change (this is kept in nifi state) as last_chg_time and changes to any records after the max of that will incrementally be reindexed.  This allows for the condition where messages could potentially be lost with kafka (ie we do not need to guarantee messages) but also allows elasticsearch search to work with the legacy system with no changes to the legacy system.  So a change made in legacy is automatically reflected in elasticsearch.  Finally changes directly to the db are also reflected in elasticsearch.

The query needs to be fairly lightweight since it will be recurring.  last_chg_time is not indexed in the db so we must do a full scan of the db tables to check for last_chg_times being changed.  This can be very taxing so some optimizations were made in the query to minimize the load.

The essence of the query to check for changes is this:

```
select person_parent_uid, max(last_chg_time) last_chg_time from (
  select person_parent_uid,last_chg_time from person where person_uid=person_parent_uid
  union
  select person_uid person_parent_uid,last_chg_time from person_name where last_chg_time is not null and person_uid in (select person_parent_uid from person)
  union
  select person_uid person_parent_uid,last_chg_time from person_race where last_chg_time is not null and person_uid in (select person_parent_uid from person)
  union
  select entity_uid person_parent_uid,last_chg_time from entity_id where last_chg_time is not null and entity_uid in (select person_parent_uid from person)
  union
  select entity_uid person_parent_uid,last_chg_time from Entity_locator_participation where last_chg_time is not null and entity_uid in (select person_parent_uid from person)
  union
  select elp.entity_uid person_parent_uid,tl.last_chg_time  from Entity_locator_participation elp join tele_locator tl on tl.tele_locator_uid=elp.locator_uid where tl.last_chg_time is not null and elp.entity_uid in (select person_parent_uid from person)
  union
  select elp.entity_uid person_parent_uid,pl.last_chg_time  from Entity_locator_participation elp join postal_locator pl on pl.postal_locator_uid=elp.locator_uid where pl.last_chg_time is not null and elp.entity_uid in (select person_parent_uid from person)
) tmptbl
group by person_parent_uid
```
Special notes on this query:
- We only index records where person_uid=person_parent_uid.  Any other records are snapshots/duplicates of ones propagated to the parent.
- Rather than doing one large join with lots of random io which could tax a large db it's broken up until full scans in each associated table and data is aggregated into a temp table.
- The associated records under person have an added check for last_chg_time not null.  This is because upon first indexing the db on a cleared state everything associated to persons gets indexed so we only ever need to check for last_chg_time not null on the associated records upon subsequent updates.  This has the added benefit that it minimizes the amount of records in our temp table. Also note in the db persons are initially created with their associated records as not null last_chg_times.
- New patient adds were already taken care of by a fast indexed db check on max of person_parent_uid.  Unfortunately both the new update checks and the old query couldn't coexist without preventing the patients from being indexed twice.  So I abandoned the new patient query in favor of this all encompassing query which does slower non-indexed db scan as light as possible.
- To be clear the information shown in our search screens which use elasticsearch always show the latest current information in the db.  This is because elasticsearch fields are searched but ultimately the patient ids are returned with a subsequent db lookup (ie the info shown to the user is not the info in elasticsearch).  This indexing is strictly so that fields that are updated or added are searchable.  In other words if you update an address on a patient and search for the patient by patient id in our api/which uses elasticsearch the most current address info will be shown to the user.  However, if you search by the new address you will only get the patient when the patient/address is indexed by these queries.

To test this:

1) stop the patient ingestion first processor and clear state. 
2) stop the last processor (elasticsearch update)
3) start the first processor
4) view the queue to elasticsearch update and note that it is bulk indexing the entire set of records in your local db (or whatever number of persons you have where person=person parent
5) let it run and note that it will not queue anymore because no changes are being made in the db
6) directly update a record in the db: ie 
`update person_name set last_chg_time=CURRENT_TIMESTAMP where person_uid=10000001 `
7) note that that record is now queued also
8) start the last processor and note how everything is indexed properly

NOTE~ have to check but I might need to coalesce add time and chg time on person and the associations to be sure new patients or new associations are updated.  Something like the following:
`IIF(COALESCE(add_time, '1/1/1973')>COALESCE(last_chg_time, '1/1/1973'),COALESCE(add_time, '1/1/1973'),COALESCE(last_chg_time, '1/1/1973')) last_chg_time`


adamloup-enquizit (approved):
👍

PR #305, opened 2023-06-16
CNFT2-473 Api Create Value Set

Emelias-Enquizit:


apps/question-bank/build.gradle
```diff
@@ -87,9 +87,13 @@ dependencies {
 
     // id generation
     implementation project(':id-generator')
+    
+    
+    implementation project(':database-entities')
 
 }
 
+compileJava.mustRunAfter(":database-entities:build")
```
mpeels, line 96: Since the only Entity that is being pulled in from the database-entities is the `CodeValueGeneral` entity, lets just copy it over.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetCreator.java
```diff
@@ -0,0 +1,173 @@
...
+				request.getValueSetNm(), 
+				request.getLdfPicklistIndCd(), 
+				request.getValueSetCode(),
+				request.getValueSetTypeCd(), 
+				request.getValueSetOid(), 
+				request.getValueSetStatusCd(),
+				request.getValueSetStatusTime(),
+				request.getParentIsCd(), 
+				request.getAddTime(), 
+				request.getAddUserId());
```
mpeels, line 137: Probably don't want to allow the UI to specify which user to set as the "added by" user. Can use the `UserDetailsProvider` component to retrieve the current user form the context

Emelias-Enquizit: @mpeels updated

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/response/CreateValueSetResponse.java
```diff
@@ -0,0 +1,34 @@
...
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public static class ValueSetCreateShort {
+	 private CodesetId id;
+	 private Instant addTime;
+	 private Long addUserId;
+	 private String valueSetNm;
+	
+}
```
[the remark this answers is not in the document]
Emelias-Enquizit: @mpeels updated

mpeels (approved):

PR #308, opened 2023-06-16
General patient info update integration

hclarkEnq:
api for update is not fully functional but the integration is set from the UI calling the API ,


adamloup-enquizit (approved):
👍

PR #312, opened 2023-06-20
CNFT2-477 Legacy Delete Value Set

Emelias-Enquizit:
Should be merged after: https://github.com/CDCgov/NEDSS-Modernization/pull/305 is merged.


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetController.java
```diff
@@ -0,0 +1,47 @@
...
+		 
+	 }
+	 
+
+	@DeleteMapping("/{codeSetNm}")
+	@PreAuthorize("hasAuthority('LDFADMINISTRATION-SYSTEM')")
+	public ResponseEntity<DeleteValueSetResponse> deleteValueSet(@PathVariable String codeSetNm) {
+		DeleteValueSetResponse response = valueSetDeletor.deleteValueSet(codeSetNm);
+		return new ResponseEntity<>(response, null, response.getStatus());
+	}
```
mpeels, line 45: Do we want to go ahead and add support for marking an inactive value set active? Could either do it as part of this story or create a new.

Emelias-Enquizit: @mpeels  it is a pretty quick update. I will include it in this story. Thanks.

Emelias-Enquizit: @mpeels  updated and included

mpeels (approved):

PR #324, opened 2023-06-23
CNFT2-556 Manage Pages view layout

jlee-enquizit:
This starts the PageBuilder suite of pages--context with fake API data, general component structure et al.
Note--I removed the weird 5px border that wrapped the app.

Hoping to get this merged before any other devs start jumping into PB frontend tix.


apps/modernization-ui/src/providers/PageBuilderContext.tsx
mpeels, line 1: I believe the `PageBuilderContext.tsx` can also be moved into `apps/page-builder`

jlee-enquizit: Oops--got it!

mpeels (approved):

PR #326, opened 2023-06-26
CNFT1-1055,1053,1052 GQL/API changes for StateCode and CountryCode, patient not redeletable

stevegsa:
Screenshot 2023-06-26 at 9 20 40 PM


apps/modernization-api/src/test/java/gov/cdc/nbs/patient/profile/CountryCodeResolverTest.java
```diff
@@ -0,0 +1,20 @@
...
+import gov.cdc.nbs.entity.srte.CountryCode;
+import org.junit.jupiter.api.Test;
+import java.util.Optional;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class CountryCodeResolverTest {
+    @Test
+    void should_resolve_state_code_for_postal_locator() {
```
adamloup-enquizit, line 13: This test and `StateCodeResolverTest` are failing with null pointers.  The repositories aren't being set.  We might be able to test these with Cucumber or can pass in mocked repositories.

adamloup-enquizit (changes requested):
Cucumber is reporting that `the patient has been deleted` is undefined.

The frontend build is failing due to eslint failing two files having formatting issues, if you're using VS Code there is a way to auto-format on save.  You can also run eslint from the command line.

From the `modernization-ui` folder you can run these commands so clear up the formatting

`npx eslint --fix src/pages/advancedSearch/AdvancedSearch.tsx`
`npx eslint --fix src/pages/advancedSearch/components/PatientResults.tsx`

adamloup-enquizit (approved):

PR #337, opened 2023-06-29
Resolves error from state and country lookups for Patient Search results

adamloup-enquizit:
Prevents a lookup of Country and State for Patient Search results without a country or state.

### DIFF
--- a/apps/modernization-api/src/test/java/gov/cdc/nbs/patient/profile/StateCodeResolverTest.java
+++ b/apps/modernization-api/src/test/java/gov/cdc/nbs/patient/profile/StateCodeResolverTest.java
@@ -1,0 +1,73 @@
+package gov.cdc.nbs.patient.profile;
+
+import gov.cdc.nbs.entity.odse.Locator;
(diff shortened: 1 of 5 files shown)


hclarkEnq (approved):

apps/modernization-api/src/test/java/gov/cdc/nbs/patient/profile/StateCodeResolverTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
+
+class StateCodeResolverTest {
+
+    @Test
+    void should_lookup_state_when_a_postal_locator_has_a_state() {
```
stevegsa, line 22: Nice... thanks!!!

PR #340, opened 2023-06-29
Cnft1 1091 event search bug fixes

hclarkEnq:
Event Method fields are now filtering
 coded values in Event and Lab search fields to resolve UI display for Codes instead of values
 Default Event checkboxes to NBS Standard
Correct "Date" fields for Laboratory Search.


apps/modernization-ui/src/pages/advancedSearch/components/eventSearch/GeneralSearch.tsx
```diff
@@ -21,6 +21,7 @@ type GeneralSearchProps = {
 
 export const GeneralSearch = ({ control, filter }: GeneralSearchProps) => {
     const [facilityType, setFacilityType] = useState(false);
+    const [eventDateType, setEventDateType] = useState<string>('');
```
adamloup-enquizit, line 24: Should be able to use that `useWatch` hook here

adamloup-enquizit (approved):

adamloup-enquizit (approved):

PR #347, opened 2023-07-05
CNFT2-554 sort page table

mpeels:
1. Creates docker container for page builder service
2. Creates traefik routing for new service
3. Updates modernization-ui proxy routing to allow multiple service connections
4. Updates modernization-ui routes to use nested routing for page-builder:
    ```javascript
    <Route path="/page-builder">
        <Route path="manage">
            <Route path="pages" element={<ManagePages />} />
        </Route>
        <Route path="add">
            <Route path="page" element={<AddNewPage />} />
        </Route>
    </Route>
    ```
5. Creates API method generation script for page-builder. Generated files are located in `modernization-ui/src/apps/page-builder/generated/`
6. Adds the `set-page-size` action to `usePage` to allow changing page size in the future


apps/modernization-ui/src/page/usePage.tsx
```diff
@@ -23,6 +27,8 @@ const pageReducer = (state: Page, action: Action) => {
             return { ...state, status: Status.Requested, current: action.page };
         case 'reload':
             return { ...state, status: Status.Requested };
+        case 'set-page-size':
+            return { ...state, pageSize: action.pageSize, page: 1, status: Status.Requested };
```
adamloup-enquizit, line 31: should be `current: 1` instead of `page: 1`.  The state stores the current page number in the `current` property.  The `go-to` action has the requested page number in the `page` property so that the action `{ type: 'go-to', page: 5 }` can be read as `go to page 5`

mpeels: 🏆

jlee-enquizit (approved):

PR #352, opened 2023-07-06
Cnft2 539 create condition

adasika-cdc:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/condition/request/CreateConditionRequest.java
```diff
@@ -0,0 +1,67 @@
...
+    private String contactTracingEnableInd;
+    private String vaccineEnableInd;
+    private String treatmentEnableInd;
+    private String labReportEnableInd;
+    private String morbReportEnableInd;
+    private String portReqIndCd;
+    private String familyCd;
+    private String coinfectionGrpCd;
+    private String rhapParseNbsInd;
+    private String rhapActionValue;
```
mpeels, line 47: Instead of accepting every field present in the database entity, it would be better to match the fields the user inputs into the UI and then populate the entity with data in the same manner as legacy.

image

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/condition/request/CreateConditionRequest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.time.Instant;
+import gov.cdc.nbs.questionbank.entity.condition.CondtionCode;
+import gov.cdc.nbs.questionbank.entity.condition.CodeToCondition;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+public class CreateConditionRequest {
```
mpeels, line 12: This could be a record

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/condition/CodeToCondition.java
```diff
@@ -0,0 +1,60 @@
...
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+@Setter
+@Entity
+@Table(name = "Code_to_condition", catalog="NBS_SRTE")
+public class CodeToCondition {
+    @EmbeddedId
+    private CodeToConditionId id;
```
mpeels, line 21: I don't see this object being defined anywhere. Also, how does the `Code_to_condition` table work? What is it used for? 

When I create a condition through the legacy UI, no entry is made into the `Code_to_condition` table.

mpeels (changes requested):
Will need to add BDD tests as well.

PR #354, opened 2023-07-06
CNFT1-1094 Logging calls to NBS Classic

adamloup-enquizit:
Adds very basic logging to the outgoing requests to NBS Classic so that communication from `modernization-api` to NBS Class can be verified using the log.

Any request from `modernization-api` directly to NBS Classic will be logged to the `nbs.classic.outgoing` logger as ` {http-method} {path}       Response: {response-status}`.  For example;

```
2023-07-06 13:23:08.973 DEBUG 10388 --- [nio-9080-exec-9] nbs.classic.outgoing                     : GET /nbs/HomePage.do       Response: 200 OK
2023-07-06 13:23:09.094 DEBUG 10388 --- [nio-9080-exec-9] nbs.classic.outgoing                     : GET /nbs/PatientSearchResults1.do  Response: 200 OK
```

The level of the logging can be changed using the standard logging properties, the default level has been set to `DEBUG` for the meantime.

```yaml
logging:
  level:
    nbs:
      classic:
        outgoing: DEBUG
```


mpeels (approved):
Look decent to me! 🥉

PR #357, opened 2023-07-07
Ab sandbox readme update

alaapbharadwaj:
Small addition to the README file within the cdc_sandbox, for better understanding of how to store the db restore files in the correct path.


mpeels (dismissed):

adamloup-enquizit:
Step 10 in the root level [README](https://github.com/CDCgov/NEDSS-Modernization/blob/main/README.md#mac-arm-system-m1m2)  provides an unzip command.  That might be useful to include in the cdc-sandbox README also.  

`unzip -j ~/Downloads/db-restore.zip -d {location}/NEDSS-Modernization/cdc-sandbox/db/restore/restore.d`

PR #359, opened 2023-07-07
- Alert component with context and hook

hclarkEnq:
* created new component for the alert toast notifications. 
* added styles. 
* created context with hook to communicate to components the alert dynamically. 
* implemented message notification to all tables and cards in demographics tab.


adamloup-enquizit (approved):
👍

PR #378, opened 2023-07-18
Cnft2 539 create condition

adasika-cdc:


apps/question-bank/src/test/resources/features/condition/CreateCondition.feature
```diff
@@ -0,0 +1,24 @@
...
+Feature: Create condition
+    When attempting to create a condition
+
+    Scenario Outline: ConditionCd already exists
+        Given ConditionCd already exists
+        When Create condition
+        Then A condition creation exception is thrown
+        Examples:
+            | condition |
+            | id      |
```
mpeels, line 11: the example is not being used

apps/question-bank/src/test/resources/features/condition/CreateCondition.feature
```diff
@@ -0,0 +1,24 @@
...
+
+    Scenario Outline: Condition name already exists
+        Given A condition name already exists
+        When Create condition
+        Then a condition creation exception is thrown
+        Examples:
+            | condition |
+            | name      |
```
mpeels, line 19: same as above. the example is not being used

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/condition/ConditionCode.java
```diff
@@ -107,13 +113,33 @@ public class ConditionCode implements Serializable {
     private Character morbReportEnableInd;
 
     @Column(name = "port_req_ind_cd")
-    private Character portReqIndCd;
+    private String portReqIndCd;
```
mpeels, line 116: The database defines this as a char(1). 
image

we should change the input to a Character instead of changing the database entity to hold a String

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/condition/ConditionCode.java
```diff
@@ -107,13 +113,33 @@ public class ConditionCode implements Serializable {
...
+        this.codeSystemDescTxt = request.codeSystemDescTxt();
+        this.conditionShortNm = request.conditionShortNm();
+        this.nndInd = request.nndInd();
+        this.progAreaCd = request.progAreaCd();
+        this.reportableMorbidityInd = request.reportableMorbidityInd();
+        this.reportableSummaryInd = request.reportableSummaryInd();
+        this.contactTracingEnableInd = request.contactTracingEnableInd();
+        this.familyCd = request.familyCd();
+        this.coinfectionGrpCd = request.coinfectionGrpCd();
+    }
```
mpeels, line 144: I believe an entry in the `LDF_page_set` table is created when creating a new condition through legacy. We need to verify that happens in our new API as well

PR #396, opened 2023-07-21
BBJ12-14_Demo_add_new_patient_Incorrect_warning_message

JimBarrows-Enquizit:


apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -221,13 +221,13 @@ export default function AddPatient() {
                                     <div className="width-full text-bold flex-row display-flex flex-align-center flex-justify">
                                         New patient
                                         <div className="button-group">
-                                            {!isValid && (
+                                            {isValid && (
```
adamloup-enquizit, line 224: The `isValid` flag is coming from form validation, this change would allow adding a new Patient with invalid values.  The real issue is there is no definition of what "missing values" means.

adamloup-enquizit (changes requested):
I don't know that this bug is workable without a definition of what "missing values" means.

PR #397, opened 2023-07-21
CNFT2-614-Add tab to page

massad22:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/addtab/model/CreateTabRequest.java
mpeels, line 1: The create request should simply be:
```java
public record CreateTabRequest(long page, String name, boolean visibile) {}
```

All the other fields should be set in the same manner as legacy.

The legacy UI only has these fields:

image

apps/question-bank/src/test/java/gov/cdc/nbs/questionbank/addtab/AddTabServiceTest.java
```diff
@@ -0,0 +1,43 @@
+package gov.cdc.nbs.questionbank.addtab;
+
+import gov.cdc.nbs.questionbank.addtab.exceptions.AddTabException;
+import gov.cdc.nbs.questionbank.addtab.model.CreateTabRequest;
+import gov.cdc.nbs.questionbank.addtab.repository.WaUiMetaDataRepository;
+import org.junit.Test;
```
mpeels, line 6: think this should be import `org.junit.jupiter.api.Test;`

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/addtab/CreateTabService.java
```diff
@@ -0,0 +1,72 @@
...
+        waUiMetadata.setAddUserId(uid);
+        waUiMetadata.setWaTemplateUid(request.page());
+        waUiMetadata.setNbsUiComponentUid(1010L);
+        Long nbsUiComponentUid = waUiMetadata.getNbsUiComponentUid();
+        Long nextOrderNumber = null;
+
+        if (nbsUiComponentUid.equals(1010L)) {
+            // Get the next order number for NBS_UI component with nbsUiComponentUid 1010
+            nextOrderNumber = getCurrentHighestOrderNumber(1010L) + 1;
+        }
```
mpeels, line 47: Does this need to find the highest `1010` component or just the highest `orderNbr`?

mpeels:
Would be nice to have a few BDD tests.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/addtab/CreateTabService.java
massad22: Resolved

massad22:
> Would be nice to have a few BDD tests.

apps/question-bank/src/test/java/gov/cdc/nbs/questionbank/addtab/AddTab.scenario
mpeels, line 1: This file can be removed as it was replace by the `AddTab.feature` files

mpeels (changes requested):

mpeels (approved):

PR #401, opened 2023-07-24
Add condition Id to Elasticsearch investigation ingestion

mpeels:
Revert change to search condition by condition name as these can be changed. 

Updates NiFi investigation ingestion to include the `public_health_case.cd` column to the Elasticsearch index.


stevegsa (approved):
Looks good!

PR #404, opened 2023-07-25
Cnft1 1162 patient profile geocoding and address verification UI

GangulaMounika:


apps/modernization-ui/src/pages/patient/profile/addresses/AddressEntryForm.tsx
```diff
@@ -49,256 +162,406 @@ export const AddressEntryForm = ({ action, entry, onChange, onCancel }: EntryPro
...
+                                                <li key={idx} className="usa-nav__submenu-item">
+                                                    <SearchCriteriaContext.Consumer>
+                                                        {({ searchCriteria }) => {
+                                                            return (
+                                                                <Button
+                                                                    key={idx}
+                                                                    onClick={() =>
+                                                                        handleSuggestionSelection(
+                                                                            idx,
+                                                                            searchCriteria.states
```
adamloup-enquizit, line 271: The state list is available in the component in `location.states`.  Trying to keep `SearchCriteriaContext` use limited in favor of the hooks that pull multiple lists at once.

adamloup-enquizit (changes requested):
`const location = useLocationCodedValues();` this hook loads the `states` and `countries` lists and should be used.  The Patient Profile isn't wrapped in a `SearchCriteriaContext` so any lists pulled from there will be empty.

apps/modernization-ui/src/pages/patient/profile/addresses/AddressEntryForm.tsx
GangulaMounika: Sounds good.

adamloup-enquizit (approved):
Looks good! 👍

PR #405, opened 2023-07-25
adding scroll to bottom of event search and added validation country code

hclarkEnq:
https://cdc-nbs.atlassian.net/browse/BBJ12-29-Notification status drop-down field inoperable
Added style for notification status, if its open then it will open towards the top instead of opening below the viewer point

https://cdc-nbs.atlassian.net/browse/BBJ12-25-Added validation for this field to not allow special characters

### DIFF
--- a/apps/modernization-ui/src/pages/patient/profile/phoneEmail/PhoneEmailEntryForm.tsx
+++ b/apps/modernization-ui/src/pages/patient/profile/phoneEmail/PhoneEmailEntryForm.tsx
@@ -101,7 +101,7 @@ export const PhoneEmailEntryForm = ({ action, entry, onChange, onCancel }: Entry
                             rules={{
                                 pattern: {
                                     value: /^\+?\d{1,6}$/,
-                                    message: 'Invalid country code'
+                                    message: 'A country code should be 1 to 3 digits"'
                                 }
                             }}
                             render={({ field: { onChange, value }, fieldState: { error } }) => {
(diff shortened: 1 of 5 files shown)


apps/modernization-ui/src/pages/patient/profile/phoneEmail/PhoneEmailEntryForm.tsx
```diff
@@ -97,18 +98,30 @@ export const PhoneEmailEntryForm = ({ action, entry, onChange, onCancel }: Entry
...
-                                    label="Country code"
-                                    name="countryCode"
-                                    htmlFor="countryCode"
-                                    id="countryCode"
-                                />
-                            )}
+                            rules={{
+                                pattern: {
+                                    value: /^\+?\d{1,6}$/,
+                                    message: 'Invalid country code'
```
adamloup-enquizit, line 104: Two things here, when using the rules the `message` with be available in `error.message` so for the Input property you can use `error={error?.message}`.

For the message let's try to be proactive for 508 compliance,  can the validation message be changed to "A country code should be 1 to 3 digits"

adamloup-enquizit (changes requested):

apps/modernization-ui/src/pages/patient/profile/phoneEmail/PhoneEmailEntryForm.tsx
hclarkEnq: sounds good, thanks

adamloup-enquizit (approved):

PR #406, opened 2023-07-25
Bug bash fixes as of 20230725

adamloup-enquizit:
Merges bug fixes as of 07/25/2023.

- adding scroll to bottom of event search and added validation country code (#405)
- Rename field cd -> condition (#403)
- switches the state list to the unbounded list (#402)
- Add condition Id to Elasticsearch investigation ingestion (#401)
- Update to use condition name in search instead of Id (#400)
- ensure identifications are indexed when added to a patient (#399)
- BBJ12-33 zip code removal (#398)
- fixing clearing all button issues for event search and phone number field (#395)

This cause a dependency with version `1.19.0-SNAPSHOT.ff8163b` of the NiFi container


achapm (approved):
👍

PR #414, opened 2023-07-28
CNFT2-617 Add section to page

massad22:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/addsection/CreateSectionService.java
```diff
@@ -0,0 +1,67 @@
...
+public class CreateSectionService implements CreateSectionInterface {
+
+    @Autowired
+    private WaUiMetaDataRepository waUiMetaDataRepository;
+
+    @Override
+    public CreateSectionResponse createSection(Long userId, CreateSectionRequest request) throws AddSectionException {
+        try {
+            WaUiMetadata waUiMetadata = createWaUiMetadata(userId, request);
+            log.info("Saving Rule to DB");
```
mpeels, line 26: Log message needs updated

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/addsection/model/CreateSectionRequest.java
```diff
@@ -0,0 +1,5 @@
+package gov.cdc.nbs.questionbank.addsection.model;
+
+public record CreateSectionRequest(long wa_ui_metadata_uid, long page, String name, String visible) {
```
mpeels, line 3: the generic `wa_ui_metadata_uid` field could be changed to something like `tabId` to make it a bit clearer on what type of component it is referencing

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/addtab/repository/WaUiMetaDataRepository.java
```diff
@@ -10,4 +10,7 @@ public interface WaUiMetaDataRepository extends JpaRepository<WaUiMetadata, Long
 
     @Query(value = "SELECT MAX(w.order_nbr) FROM WA_Ui_metadata w WHERE w.wa_template_uid = ?1 ", nativeQuery = true)
     Long findMaxOrderNumberByTemplateUid(Long waTemplateId);
+    @Query(value = "SELECT w.order_nbr FROM WA_Ui_metadata w WHERE w.wa_ui_metadata_uid = ?1 ", nativeQuery = true)
+    Long findOrderNbrForSection(Long waUiMetadataUid);
+
```
mpeels, line 15: Its better to avoid nativeQuery's wherever possible. Hibernate can do some optimizations on JPQL queries that it cant on native queries. -- That being said, this query is so simple it won't make much difference but its still nice to try to avoid nativeQueries when you can.

mpeels: Oh, and this method name could be a little more generic as well. Since it doesn't use a `nbs_ui_component_uid` in the query, it could just be `findOrderNbr`. This would make it look cleaner when re-using this to find the order nbr for other wa_ui_metatadata objects.

apps/question-bank/src/test/java/gov/cdc/nbs/questionbank/addsection/AddSectionServiceTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    void createSectionServiceTestException() throws AddSectionException {
+
+        AddSectionException exception =
+                assertThrows(AddSectionException.class, () -> createSectionService.createSection(123L, null));
+
+        assertEquals(
+                "gov.cdc.nbs.questionbank.addsection.exception.AddSectionException: Add section exception",
+                exception.toString());
```
mpeels, line 43: the assertEquals is redundant. The assertThrows will cause the test to fail if the exception isn't thrown.

PR #417, opened 2023-07-28
CNFT2-624 Save Page As Draft

Emelias-Enquizit:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageStateChanger.java
```diff
@@ -0,0 +1,50 @@
...
+				response.setStatus(HttpStatus.NOT_FOUND);
+				response.setTemplateId(id);
+			}
+		} catch (Exception e) {
+			response.setMessage(e.getMessage());
+			response.setStatus(HttpStatus.INTERNAL_SERVER_ERROR);
+			response.setTemplateId(id);
+
+		}
+		return response;
```
mpeels, line 47: Legacy does not allow converting an existing "published" page back into a draft. It creates a new page from the existing one without modifying the published page. (Plz double check and validate this though).

Workflow that I have seen:
1. Create a new page (`template_type` is set to "Draft", `publish_version_nbr` of NULL)
2. Publish that page (`template_type` is set to "Published")
3. Create a draft from a published page - creates a new entry in the `wa_template` table with status of "Draft" and a `publish_version_nbr` of 0, the published page has its `template_type` set to `Published With Draft`.

The query that fetches the pages to display their summaries in table format only selects `template_type` of "Draft" or "Published", this prevents the duplicates from showing up in the summary table.

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageStateChanger.java
Emelias-Enquizit: @mpeels  sure I will confirm

Emelias-Enquizit: Hi @mpeels  updated please take a look:

Screenshot 2023-08-02 at 2 29 18 PM

Screenshot 2023-08-02 at 2 47 58 PM

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageStateChanger.java
```diff
@@ -0,0 +1,86 @@
...
+	}
+	
+	public WaTemplate createDraftCopy(WaTemplate oldPage) {
+		if (oldPage.getTemplateType().equals("Draft")) {
+			return oldPage;
+		}
+		Long id = templateRepository.getMaxTemplateID();
```
mpeels, line 54: The id column for `Wa_template` will auto generate an Id.
```java
@Id
@GeneratedValue(strategy = GenerationType.IDENTITY)
@Column(name = "wa_template_uid", nullable = false)
private Long id;
```

mpeels (approved):

PR #436, opened 2023-08-07
Read Condition

adasika-cdc:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/condition/request/ReadConditionRequest.java
```diff
@@ -0,0 +1,24 @@
...
+
+    private String coinfectionGrpCd;
+
+    private Character nndInd;
+
+    private String investigationFormCd;
+
+    private Character statusCd;
+
+}
```
mpeels, line 24: What will the UI look like for this search? Will it be a table with a search box at the top? If so, it should the request should have a single input parameter. Usually named "searchText" or similar.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/condition/repository/ConditionCodeRepository.java
```diff
@@ -12,4 +14,9 @@ public interface ConditionCodeRepository extends JpaRepository<ConditionCode, St
     long checkId(@Param("id")String id);
     @Query("SELECT count(*) FROM ConditionCode c WHERE c.conditionShortNm=:name")
     long checkConditionName(@Param("name") String name);
+
+    @Query("SELECT c FROM ConditionCode c WHERE c.id LIKE %:id% OR c.conditionShortNm LIKE %:conditionShortNm% OR c.progAreaCd LIKE %:progAreaCd% OR c.familyCd LIKE %:familyCd% OR c.coinfectionGrpCd LIKE %:coinfectionGrpCd% OR c.nndInd =:nndInd OR c.investigationFormCd LIKE %:investigationFormCd% OR c.statusCd =:statusCd")
+    Page<ConditionCode> findByField(@Param("id")String id, @Param("conditionShortNm") String conditionShortNm, @Param("progAreaCd") String progAreaCd,
+                                    @Param("familyCd")String familyCd, @Param("coinfectionGrpCd")String coinfectionGrpCd, @Param("nndInd")Character nndInd, @Param("investigationFormCd") String investigationFormCd,
+                                    @Param("statusCd")Character statusCd, Pageable pageable);
 }
```
mpeels, line 22: If the UI is going to be a single search box, this would be changed to have a single parameter "searchText" or "searchQuery" and use that value in each of the `LIKE`s

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/condition/request/ReadConditionRequest.java
mpeels: Hmm looking at: 
image

It seems we will need to support both a generic search input as well as individual field filtering where if both are supplied they would be `AND`ed together

So we may need to change the input for the search to be a single text entry as well as a list of filters.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/condition/request/ReadConditionRequest.java
```diff
@@ -0,0 +1,18 @@
...
+import lombok.Data;
+
+@Data
+public class ReadConditionRequest {
+
+    private String searchText;
+
+    private String filterField;
+
+    private String filterValue;
```
mpeels, line 12: can they only send 1 filter at a time or should this be a list of filter objects?

mpeels (approved):

PR #437, opened 2023-08-07
CNFT2-478 Legacy Create Page

Emelias-Enquizit:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageController.java
```diff
@@ -78,6 +83,13 @@ public Page<PageSummary> search(
         return results;
     }
     
+    @PostMapping("/add")
+    public PageCreateResponse createPage(@RequestBody PageCreateRequest request) {
+    	Long userId = userDetailsProvider.getCurrentUserDetails().getId();
+    	return creator.createPage(request,userId);
+    }
```
mpeels, line 90: Hmm looks like I painted you into a corner for the mapping. The @PostMapping above this one that is used for search should really be updated to `@PostMapping("/search")` so that this one could be `@PostMapping`.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/response/PageCreateResponse.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+public class PageCreateResponse {
+		private Long pageId;
+		private String pageName;
+		private String message;
+		private HttpStatus status;
```
mpeels, line 14: Could be a record

apps/question-bank/src/test/resources/features/page/PageCreator.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Add Page
+
+
+	Scenario: As an admin I can create a Page
+         Given I am an admin user make an add page request
+         When I make a request to create a Page
+         Then A page is created
```
mpeels, line 7: Should be able to add some permission BDDs without adding any more code:

```cucumber
    Scenario: I cannot create a page without being logged in
        Given I am not logged in
        When I make a request to create a Page
        Then a no credentials found exception is thrown

    Scenario: I cannot create a page without proper permissions
        Given I am a user without permissions
        When I make a request to create a Page
        Then an accessdenied exception is thrown
```

mpeels (changes requested):

mpeels (approved):

PR #448, opened 2023-08-09
CNFT2-619 revisions

mpeels:


massad22 (approved):
All changes look good as discussed earlier and currently on chat as the orderNbr part was important. Thank you Michael

PR #452, opened 2023-08-11
CNFT2-829: Add Concept Code to Response Body

Emelias-Enquizit:
Screenshot 2023-08-11 at 12 21 11 PM

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetReader.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetReader.java
@@ -99,6 +99,7 @@ public class ValueSetReader {
                 codeValueGeneral.getConceptPreferredNm(),
                 codeValueGeneral.getCodeSystemDescTxt(),
                 codeValueGeneral.getConceptStatusCd(),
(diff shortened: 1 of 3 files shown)


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetReader.java
```diff
@@ -99,6 +99,7 @@ public Concept toConcept(CodeValueGeneral codeValueGeneral) {
                 codeValueGeneral.getConceptPreferredNm(),
                 codeValueGeneral.getCodeSystemDescTxt(),
                 codeValueGeneral.getConceptStatusCd(),
+                codeValueGeneral.getConceptCode(),
```
mpeels, line 102: It looks like `getConceptCode()` is already being called on line 98, and corresponds to the `value` field in the `Concept` object.

Emelias-Enquizit: @mpeels  sure I removed the additional call to getConceptCode() but renamed value => conceptCode in the response for better readability.

mpeels (approved):

lramakr (approved):
LGTM

PR #464, opened 2023-08-17
Added sonar_scan reusable workflow

njaved:
Updated logic on when SonarCloud gets triggered:

1) Sonar scan will run on ALL PR
2) Sonar scan will run on a PUSH only when Build-and-deploy-modernization-api Workflow runs. It will then trigger steps of Sonar Scanner workflow.


szamfir (approved):
LGTM, if we wanted to be extra safe we can add a needs: sonar_scan to the call-build-microservice-container-workflow step, but I think it would be extra.

PR #469, opened 2023-08-18
CNFT2-741-Section+SubSection-Update/Delete

massad22:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/tab/repository/WaUiMetaDataRepository.java
```diff
@@ -73,4 +73,29 @@ Long findOrderNbrOfNextSectionOrTab(
...
+    @Query(value = "update wa_ui_metadata set order_nbr = order_nbr - 1 where order_nbr >= ?1 and wa_ui_metadata_uid!=?2", nativeQuery = true)
+    void updateOrderNumberByDecreasing(Integer orderNbr, Long waUiMetadataUid);
+
+
+    @Query(value = "select order_nbr from WA_Ui_metadata w where w.wa_ui_metadata_uid=?1", nativeQuery = true)
+    Integer getOrderNumber(Long waUiMetadataUid);
+
+    @Modifying
+    @Query(value = "delete from WA_Ui_metadata where wa_ui_metadata_uid=?1", nativeQuery = true)
+    void deletefromTable(Long waUiMetadataUid);
```
mpeels, line 99: the built in `delete` function should do the same thing

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/subsection/SubSectionCreator.java
```diff
@@ -37,6 +43,42 @@ public CreateSubSectionResponse createSubSection(long pageId, Long userId, Creat
...
+                return new UpdateSubSectionResponse(request.subSectionId(), "SubSection Updated Successfully");
+            } else if ( request.questionLabel() != null ) {
+                waUiMetaDataRepository.updateQuestionLabel(request.questionLabel(), request.subSectionId());
+                return new UpdateSubSectionResponse(request.subSectionId(), "SubSection Updated Successfully");
+            } else if ( request.visible() != null ) {
+                waUiMetaDataRepository.updateVisibility(request.visible(), request.subSectionId());
+                return new UpdateSubSectionResponse(request.subSectionId(), "SubSection Updated Successfully");
+            } else {
+                return new UpdateSubSectionResponse(request.subSectionId(), "questionLabel or Visible is required to update subsection");
+            }
```
mpeels, line 74: I think we should require both the label and the `visible` values to be present. Then if either are null, throw an exception. Then this becomes something like:

```java
if (request.questionLabel() == null || request.visible() == null) {
    throw new UpdateSubSectionException("Label and visibility fields are required");
}
waUiMetaDataRepository.updateQuestionLabelAndVisibility(request.questionLabel(), request.visible(), request.subSectionId());
return new UpdateSubSectionResponse(request.subSectionId(), "SubSection Updated Successfully");
```

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/section/exception/DeleteSectionException.java
```diff
@@ -0,0 +1,18 @@
...
+        super(message);
+        this.errorCode = errorCode;
+
+    }
+
+    public int getErrorCode() {
+        return errorCode;
+    }
+
+}
```
mpeels, line 17: If these new custom exceptions extend either `BadRequestException` or `NotFoundException`, then the `PageBuilderExceptionHandler` will return the "message" to the user and set the appropriate status code on the response.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/section/SectionCreator.java
```diff
@@ -37,6 +46,35 @@ public CreateSectionResponse createSection(long pageId, Long userId, CreateSecti
...
+
+    public DeleteSectionResponse deleteSection(DeleteSectionRequest request) {
+        try {
+            log.info("Deleting section");
+            Integer orderNbr = waUiMetaDataRepository.getOrderNumber(request.sectionId());
+            waUiMetaDataRepository.deleteById(request.sectionId());
+            waUiMetaDataRepository.updateOrderNumberByDecreasing(orderNbr, request.sectionId());
+            return new DeleteSectionResponse(request.sectionId(), "Section Deleted Successfully");
+        } catch(Exception exception) {
+            throw new DeleteSectionException(exception.toString());
```
mpeels, line 58: I don't believe you can delete a section if it has any questions in it. We would need to verify that the next orderNumber after the section is either: 
1. another section
2. a tab
3. null 

if one of those 3 is true, then we can do the delete, else throw a `DeleteSectionException`

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/subsection/SubSectionCreator.java
```diff
@@ -37,6 +44,35 @@ public CreateSubSectionResponse createSubSection(long pageId, Long userId, Creat
...
     }
 
+
+    public DeleteSubSectionResponse deleteSubSection(DeleteSubSectionRequest request) {
+        try {
+            log.info("Deleting section");
+            Integer orderNbr = waUiMetaDataRepository.getOrderNumber(request.subSectionId());
+            waUiMetaDataRepository.deleteById(request.subSectionId());
+            waUiMetaDataRepository.updateOrderNumberByDecreasing(orderNbr, request.subSectionId());
+            return new DeleteSubSectionResponse(request.subSectionId(), "Sub Section Deleted Successfully");
```
mpeels, line 54: I believe the same logic applies here as above, you cannot delete a subsection if it has any content.

so I think the check would be "is the wa_ui_metadata entry with the next order number: "
1. null
2. another subsection
3. another section
4. a tab

if one of those 4 are true, then allow the delete.

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/subsection/SubSectionCreator.java
```diff
@@ -37,6 +44,41 @@ public CreateSubSectionResponse createSubSection(long pageId, Long userId, Creat
...
     }
 
+
+    public DeleteSubSectionResponse deleteSubSection(DeleteSubSectionRequest request) {
+        try {
+            log.info("Deleting sub section");
+            Integer orderNbr = waUiMetaDataRepository.getOrderNumber(request.subSectionId());
+            Long pageNumber = waUiMetaDataRepository.findPageNumber(request.subSectionId());
+            Long nbsComponentUid = waUiMetaDataRepository.findNextNbsUiComponentUid(orderNbr+1, pageNumber);
+            if(nbsComponentUid == 1010L || nbsComponentUid == 1015L || nbsComponentUid == null || nbsComponentUid == 1016L) {
```
mpeels, line 54: Similar to above, can we define some constants for these Ids to improve readability?

mpeels (approved):

massad22:
SectionCreatorTest

massad22:
SubSectionControllerTest

massad22:
SubSectionCreatorTest

massad22:
SectionControllerTest

PR #477, opened 2023-08-21
CNFT2-854 Attach Create New Condition modal

jlee-enquizit:
Here we add the CreateCondition modal to Add New Page.
A refactor that uses our native components for the CreateCondition form is forthcoming.


apps/modernization-ui/src/apps/page-builder/components/CreateCondition/CreateCondition.tsx
```diff
@@ -37,8 +41,8 @@ export const CreateCondition = () => {
         }).then((response: any) => {
             const data = response || [];
             const codingSystemList: never[] = [];
-            data.map((each: { value: never }) => {
-                codingSystemList.push(each.value);
+            data.map((each: { conceptCode: never }) => {
+                codingSystemList.push(each.conceptCode);
             });
             setSystemOptions(codingSystemList);
```
mpeels, line 47: We should avoid using `:any` and `never[]`. The proper types can be seen in the generated API calls. I think this is also causing the drop down selections to be empty:
image



I think this fixes it for the `Coding System`

```typescript
const fetchCodingSystemOptions = () => {
    ValueSetControllerService.findConceptsByCodeSetNameUsingGet({
        authorization: `Bearer ${state.getToken()}`,
        codeSetNm: 'CODE_SYSTEM'
    }).then((response: Concept[]) => {
        setSystemOptions(response);
    });
};
```

```typescript
<select
    className="field-space"
    name="codingSystem"
    value={system}
    onChange={(e: any) => setSystem(e.target.value)}>
    <option>- Select -</option>
    {systemOptions.map((o, i) => (
        <option key={i} value={o.conceptCode}>
            {o.display}
        </option>
    ))}
</select>
```

mpeels:
Clicking the `Cancel` button does not close the modal

image

apps/modernization-ui/src/apps/page-builder/components/CreateCondition/CreateCondition.tsx
jlee-enquizit: Done

mpeels: Would you mind fixing the other drop down values that list `never` as their type? 

1. `fetchProgramAreaOptions`
2. `fetchGroupOptions`
3. `fetchProgramAreaOptions`

mpeels:
Completely hijacked this PR since John was out today. Fixed an issue where the newly created condition was lacking the `condition_cd` specified. Also updated it so the newly created condition is returned and selected in the "Add new page" form


Test execution:
image


Code Coverage:
image

Emelias-Enquizit (approved):

Emelias-Enquizit (approved):

hclarkEnq (approved):

PR #482, opened 2023-08-23
CNFT1-1449 Resolve event filters and forward to event search

mpeels:
[CNFT1-1449](https://cdc-nbs.atlassian.net/browse/CNFT1-1449)

1. Adds support for resolving Investigation and LabReport filters. 
2. Attempts to alleviate the "you must enter at least one item to search" while displaying search records by only adding first name and last name to the PatientFilter if they are not empty. When a user selected an unsupported `Event ID Type`, the filter received by the UI contained empty strings for first and last name, it would then perform the search while simultaneously displaying the error

image

***Notes***

Modernized search does not currently support searching by `Document ID` or `Morbidity Report ID`. If a user selects these event id types, they will be sent to the modernized advanced search with no search parameters set

When both Patient Filter and Event filter parameters are set, event search "wins"

[CNFT1-1449]: https://cdc-nbs.atlassian.net/browse/CNFT1-1449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adamloup-enquizit (approved):
👍

PR #492, opened 2023-08-25
fix build pipeline and create dependency on sonar running first

szamfir:
This change is similar to the previous one that passes in secrets that are otherwise missing. This should be the last of them.

One more significant change is to solidify the intention behind this change. If sonar does not run or fails, then the container build will fail.


adamloup-enquizit (approved):
Great!  I like that the building of the container depends on the outcome of sonar.

PR #498, opened 2023-08-28
CNFT2-923 CreateCondition refactor

jlee-enquizit:
Here we refactor CreateCondition to use our native components and react-hook-form


lramakr (approved):
LGTM

PR #502, opened 2023-08-29
Cnft1 1186 error validation bug fix

hclarkEnq:


apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -100,8 +100,26 @@ const AddPatient = () => {
...
+                    return false; // At least one property has no value
+                }
+            }
+        }
+        return true;
+    }
+
     const evaluateMissingFields = (entry: NewPatientEntry) => {
-        setEntryState({ step: 'verify-missing-fields', entry });
+        console.log('entry:', entry);
```
adamloup-enquizit, line 121: console log!

apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -100,8 +100,26 @@ const AddPatient = () => {
         setEntryState({ step: 'entry' });
     };
 
+    function isValidForm(obj: any) {
```
adamloup-enquizit, line 103: Two things here;  
1. This function is not checking the validity of the form only that missing fields exists in the new patient.  At this point the form is already valid.  It may benefit from a different name. 
2. The function is always given a type of `NewPatientEntry` so `any` can be removed.

adamloup-enquizit (changes requested):

apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
hclarkEnq: Understood

apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -100,8 +100,25 @@ const AddPatient = () => {
         setEntryState({ step: 'entry' });
     };
 
+    function isMissingFields(entry: NewPatientEntry) {
```
adamloup-enquizit, line 103: The name is a little confusing because `isMissingFields` return `false` when when fields are missing and `true` when they all contain data.

apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -100,8 +100,25 @@ const AddPatient = () => {
...
+                    return false; // At least one property has no value
+                }
+            }
+        }
+        return true;
+    }
+
     const evaluateMissingFields = (entry: NewPatientEntry) => {
-        setEntryState({ step: 'verify-missing-fields', entry });
+        setEntryState({ step: isMissingFields(entry) ? 'create' : 'verify-missing-fields', entry });
```
adamloup-enquizit, line 121: This is by passing the Address Verification. 

If the `entry` is missing fields it should call `setEntryState({step: 'verify-missing-fields', entry})` to prompt the user if they would like to continue.

If the `entry` is not missing fields it should skip the prompt and just call `setEntryState({step: 'verify-address', entry})`.

adamloup-enquizit (changes requested):

apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
hclarkEnq: just replaced the boolean values as per the function name so it makes sense now

PR #506, opened 2023-08-30
Merge cnft1 1437 and CNFT1-1456

mdavis-enquizit:
Cnft1 1437 [Demo] Able to add a new patient with validation error
-added unified validation for errors also.

CNFT1-1456Ensure Event Search Color Consitency in Select Boxes.

### DIFF
--- a/apps/modernization-ui/src/components/FormInputs/DatePickerInput.tsx
+++ b/apps/modernization-ui/src/components/FormInputs/DatePickerInput.tsx
@@ -62,9 +62,10 @@ export const DatePickerInput = ({
         const currentVal = (event.target as HTMLInputElement).value;
         const valid = isValid(currentVal);
         const validYear = isValidYear(currentVal.substr(currentVal.length - 4));
+
         setError(!valid);
         setError(!validYear);
-        
+
         onBlur && onBlur(event);
     };
 
--- a/apps/modernization-ui/src/components/FormInputs/DatePickerInput.tsx
+++ b/apps/modernization-ui/src/components/FormInputs/DatePickerInput.tsx
@@ -25,7 +25,13 @@ const inputFormat = /^[0-3]?[0-9]\/[0-3]?[0-9]\/[0-9]{4}$/;
 
 const matches = (value: string) => inputFormat.test(value);
 
+const yearLength = /^(19|20)\d{2}$/;
+
 const isValid = (value?: string) => !value || matches(value);
+
+const matchesYear = (value: string) => yearLength.test(value);
+
+const isValidYear = (value?: string) => !value || matchesYear(value);
 
 const interalize = (value: string) => {
     const [month, day, year] = value.split('/');
@@ -55,7 +61,10 @@ export const DatePickerInput = ({
     const checkValidity = (event: React.FocusEvent<HTMLInputElement> | React.FocusEvent<HTMLDivElement>) => {
         const currentVal = (event.target as HTMLInputElement).value;
         const valid = isValid(currentVal);
+        const validYear = isValidYear(currentVal.substr(currentVal.length - 4));
         setError(!valid);
+        setError(!validYear);
+        
         onBlur && onBlur(event);
     };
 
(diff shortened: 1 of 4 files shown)


apps/modernization-ui/src/components/FormInputs/DatePickerInput.tsx
```diff
@@ -25,8 +25,14 @@ const inputFormat = /^[0-3]?[0-9]\/[0-3]?[0-9]\/[0-9]{4}$/;
 
 const matches = (value: string) => inputFormat.test(value);
 
+const yearLength = /^(19|20)\d{2}$/;
+
 const isValid = (value?: string) => !value || matches(value);
 
+const matchesYear = (value: string) => yearLength.test(value);
+
```
mdavis-enquizit, line 33: Could be cleaner but not sure what makes an invalid date. This allows for 1890-2999 year value

apps/modernization-ui/src/pages/addPatient/components/otherInfoFields/OtherInfoFields.tsx
```diff
@@ -28,15 +28,19 @@ export default function OtherInfoFields({ id, title, control, coded }: Props) {
                 <Grid row>
                     <Grid col={6}>
                         <Controller
+                            rules={{
+                                required: { value: true, message: 'This field is required' }
+                            }}
                             control={control}
```
mdavis-enquizit, line 34: This was accidentally overwritten.

apps/modernization-ui/src/pages/addPatient/components/otherInfoFields/OtherInfoFields.tsx
```diff
@@ -28,15 +28,19 @@ export default function OtherInfoFields({ id, title, control, coded }: Props) {
...
-                            name="dateOfBirth"
-                            render={({ field: { onChange, value, name } }) => (
+                            name="dob"
```
adamloup-enquizit: There may have been a few merge conflicts in this file resulting in changes being reverted.  The names of the form properties were aligned with the names expected by the API.  

The `render` property of the [Controller](https://react-hook-form.com/docs/usecontroller/controller) exposes two pretty useful objects `field` and `fieldState`.  

`field` is a render prop that provides state and functions to the component rendering the controlled form field.  The `name` from the `Controller` component is available here which prevents having to make sure that the `Controller` and component have the same name and id's.  

[fieldState](https://react-hook-form.com/docs/useform/getfieldstate) contains `error` and `invalid` properties that are set based on the `rules` from the `Controller`.  This means that the `render` property can provide the error state of the controlled form field without the need to pass down the errors from the `form`.  

`render={({ field: { onChange, value, name }, fieldState:{ error } }) => (` will contain any errors for this field only (if present).  There are a couple of examples of it being used [here](https://github.com/CDCgov/NEDSS-Modernization/blob/665b5074d024f534841898d2fef695add26bf429/apps/modernization-ui/src/pages/patient/profile/addresses/AddressEntryForm.tsx#L74)

adamloup-enquizit (changes requested):
`OtherInfoFields` has a few lines that are reverting a recent change to the names of the form fields.  

CNFT1-1437 was referring to the `asOf` date field in `GeneralInformation`.  This issue seems to have been resolved in `main` as the issue is not present in the DEV environment.  

Since CNFT-1437 seems to be resolved through other means you can mark that one as done and change this PR to just address CNFT1-1456.  That part looks great.

adamloup-enquizit (approved):

adamloup-enquizit (commented):
The Sonar Scan is complaining about a few things.

```
 src/pages/addPatient/components/otherInfoFields/OtherInfoFields.tsx
  Line 17:75:  Replace `,·errors:·any;` with `;·errors:·any`  prettier/prettier
  Line 19:70:  'errors' is defined but never used             @typescript-eslint/no-unused-vars
```

For the prettier issue if you don't have the plugin installed in VS Code it can be formatted from the command line using
`npx prettier --write src/pages/addPatient/components/otherInfoFields/OtherInfoFields.tsx --config ./.prettierrc`

mdavis-enquizit:
closed for https://github.com/CDCgov/NEDSS-Modernization/pull/522/files

PR #509, opened 2023-08-31
Add test run to ui build

mpeels:
Causes UI tests to be executed as part of the build. If a UI test fails, the build will now fail.

[Failing Build](https://github.com/CDCgov/NEDSS-Modernization/actions/runs/6112416260/job/16589831914) due to UI test failure


lramakr (approved):
LGTM

PR #525, opened 2023-09-07
sidebar fixed

mdavis-enquizit:


apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -303,7 +303,7 @@ const AddPatient = () => {
                                             assigningAuthorities: coded.assigningAuthorities
                                         }}
                                     />
-                                    <div style={{ height: `calc(20%)` }} />
+                                    <div style={{ height: `calc(30%)` }} />
```
hclarkEnq, line 306: It seems that 40 % instead of 30% solves the issue 
<div style={{ height: `calc(40%)` }} />

adamloup-enquizit (approved):

PR #527, opened 2023-09-08
first name and last name check for null to not appear in search results

hclarkEnq:


apps/modernization-ui/src/pages/advancedSearch/components/InvestigationResults.tsx
```diff
@@ -70,14 +70,24 @@ export const InvestigationResults = ({
         }
     };
 
+    const formattedName = (lastNm = '', firstNm = '') => {
```
adamloup-enquizit, line 73: Can this function be pulled out into it's own file and imported when needed?  Then a unit test can be created to verify the output of the function

adamloup-enquizit (changes requested):
Extract the `formattedName` function into it's own file

adamloup-enquizit (changes requested):

adamloup-enquizit (approved):

PR #534, opened 2023-09-11
CNFT2-961 Edit Page unit tests

jlee-enquizit:
Unit tests for the Edit Page family of components.


rkrusselenq (approved):
just a suggestion, to wrap all. the tests in a describe() that tells what component it is.  For example on EditPage, do describe('Edit Page',() => ... just kinda makes it easier to read the tests when debugging no biggie though

PR #544, opened 2023-09-14
Show an Alert when a patient is deleted

GangulaMounika:
Show an Alert when a patient is deleted


adamloup-enquizit (approved):
The acceptance criteria states that the Patient Name should be shown in the alert.  There doesn't seem to be a straightforward way of doing this right know though.

GangulaMounika:
> The acceptance criteria states that the Patient Name should be shown in the alert. There doesn't seem to be a straightforward way of doing this right know though.

Good catch about the name Adam. There is a way. Just updated the code. Thanks!

adamloup-enquizit (approved):

PR #547, opened 2023-09-14
adding refetch to the patient profile and calling it based on event c…

hclarkEnq:
…hanges across the different tables in demographics tab


adamloup-enquizit (approved):
Good job!

PR #549, opened 2023-09-14
CNFT1-1498 there are lot of records with zip code 5 digits and UI does not let us search with zip code less than 5 digits

adamloup-enquizit:
Resolves [CNFT1-1498](https://cdc-nbs.atlassian.net/browse/CNFT1-1498) by allowing Patient Searches with partial zip codes.

- Relaxes the validation of zip codes on Patient Search to allow for partial searches
- Refactors the Patient Search test steps to be more generic
- Patient Search Tests now clears any data added to conduct each test
- Resolved a discrepancy between how the `ElasticsearchPerson` was being created in `test` vs `src`.  They use the same method now.
- Removed some unused code and accompanying tests 

[CNFT1-1498]: https://cdc-nbs.atlassian.net/browse/CNFT1-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/test/java/gov/cdc/nbs/patient/profile/address/change/PatientProfileAddressChangeConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Bean
+    TestActive<UpdatePatientAddressInput> activeUpdatePatientAddressInput() {
+        return new TestActive<>();
+    }
+
+    @Bean
+    TestActive<DeletePatientAddressInput> activeDeletePatientAddressInput() {
+        return new TestActive<>();
+    }
```
mpeels, line 24: Thoughts on making these a single generic method? I think that works

adamloup-enquizit: Spring gets confused and starts injecting the wrong types.  I had it working at one point without any `@Bean` methods then it broke down as I added more types.

PR #555, opened 2023-09-15
Remove sonar cache 

mpeels:
Removes the sonar cache form the workflow. 
Removes the non-working exclusion from the `sonar.gradle`
Adds `clean` to the gradle command just in case


adamloup-enquizit (approved):
🚋

PR #558, opened 2023-09-18
Disable edit buttons for deleted patient in demographics tab

GangulaMounika:


apps/modernization-ui/src/components/EditableCard/EditableCard.tsx
```diff
@@ -12,15 +12,16 @@ export type Props = {
     editing?: boolean;
     onEdit: () => void;
     children: ReactNode;
+    disableEdit?: boolean;
```
adamloup-enquizit, line 15: It's good to keep boolean variables affirmative, can `disableEdit` be renamed?  Maybe something like `editable` but then you have to flip the value when passed to `disabled`.  

Does `readOnly` make sense?  it retains the same boolean value as `disableEdit` and is affirmative.

apps/modernization-ui/src/pages/patient/profile/addresses/AddressesTable.tsx
```diff
@@ -79,6 +81,8 @@ type Props = {
 
 export const AddressesTable = ({ patient }: Props) => {
     const { showAlert } = useAlert();
+    const { id } = useParams();
```
adamloup-enquizit, line 84: Can a property be added to the table to like the `EditableCard`?  The actual identifier of the patient is already passed as the `patient` property but that' can't be used with the `usePatientProfile` hook because it expects the short-id.  Grabbing the short id from the params is placing a dependency on the URL.

adamloup-enquizit (changes requested):
Just a couple of suggestions that I can be probably be talked out of 😄 

1.  Renaming `disableEdit` to make it affirmative
2. Possibly adding a similar properties to the table.

One bigger thing is that these tables also have an `Actions` menu on each row that allows `Details`, `Edit`, and `Delete`.  The `Actions` component will also have to be updated to also not allow changes.  


`Actions` has the extra requirement where the `Delete` option can be turned off  (I see that Actions has a negative boolean property named `notDeletable` 😢 )

apps/modernization-ui/src/components/EditableCard/EditableCard.tsx
GangulaMounika: I like the idea of renaming while keeping it affirmative. Nice thoughts , thanks Adam!

GangulaMounika:
> Just a couple of suggestions that I can be probably be talked out of 😄
> 
> 1. Renaming `disableEdit` to make it affirmative
> 2. Possibly adding a similar properties to the table.
> 
> One bigger thing is that these tables also have an `Actions` menu on each row that allows `Details`, `Edit`, and `Delete`. The `Actions` component will also have to be updated to also not allow changes.
> 
> `Actions` has the extra requirement where the `Delete` option can be turned off (I see that Actions has a negative boolean property named `notDeletable` 😢 )

Fortunately the button that opens the actions is not within the actions component. I just had to disable that button instead of the whole actions component.

apps/modernization-ui/src/pages/patient/profile/addresses/AddressesTable.tsx
GangulaMounika: So the EditableCard always has an Edit button on its top right. 
But that's not the case for a Table component. You don't always have an Edit button in a table component level.

adamloup-enquizit (approved):

adamloup-enquizit (approved):

PR #560, opened 2023-09-19
CNFT2-956 Save Page as draft

jlee-enquizit:
Here we allow users to Save Page as Draft.

A new AlertBanner component is also added here.


apps/modernization-ui/src/apps/page-builder/pages/EditPage/EditPage.tsx
```diff
@@ -46,29 +50,51 @@ export const EditPage = () => {
         }
     };
 
+    const handleSaveDraft = async () => {
+        await savePageAsDraft(token, Number(pageId))
+            .then((response) => {
+                console.log(response);
```
rkrusselenq, line 56: Left a console.log, also you put an 'async' on the function but then you're using .then.catch, you can use await if using async

apps/modernization-ui/src/apps/page-builder/pages/EditPage/EditPage.tsx
```diff
@@ -46,29 +50,51 @@ export const EditPage = () => {
...
-                        <EditPageHeader page={page} />
+                        <EditPageHeader page={page} handleSaveDraft={handleSaveDraft} />
                         <EditPageTabs tabs={tabs} active={active} setActive={setActive} />
                     </div>
                     <div className="edit-page__container">
-                        {page.pageTabs[active] ? (
-                            <EditPageContentComponent content={page.pageTabs[active]} onAddSection={handleAddSection} />
-                        ) : null}
+                        <div className="edit-page__content">
+                            {alertMessage !== '' ? <AlertBanner type={alertType}>{alertMessage}</AlertBanner> : null}
```
rkrusselenq, line 77: i dont think you need !== '', as an empty string evaluates to falsy

jlee-enquizit: alertMessage default is '', only displaying when there is a message to display

apps/modernization-ui/src/apps/page-builder/components/EditPageHeader/EditPageHeader.tsx
```diff
@@ -32,7 +33,7 @@ export const EditPageHeader = ({ page }: PageProps) => {
                         </Button>
                     </>
                 ) : (
-                    <Button type="button" outline>
+                    <Button type="button" outline onClick={() => handleSaveDraft()}>
```
rkrusselenq, line 36: You dont need the call back if you're not passing arguements, you can just do `onClick={handleSaveDraft}`

rkrusselenq (dismissed):

rkrusselenq (approved):

PR #562, opened 2023-09-19
CNFT-1507 investigation status

adamloup-enquizit:
Resolves [CNFT-1507](https://cdc-nbs.atlassian.net/browse/CNFT1-1507) by ensuring that NiFi properly links a Notification to an Investigation.  Furthermore, the query generated by the API to search over Investigations was creating conditions that resulted in all Investigations being returned when a Notification Statues, Processing Status, or Case Status is present in the `InvestigationFilter`.

The `Investigation Ingestion` flow will now set a `null` value of Notification Statues, Processing Status, or Case Status to `UNASSIGNED` so that the API can use a `terms` query to search over the status values.

```json
{
  "terms": {
    "notification_record_status_cd": ["UNASSIGNED", "APPROVED"]
}
```

This will require a re-index of Investigations.


apps/modernization-api/src/main/java/gov/cdc/nbs/event/search/InvestigationFilter.java
```diff
@@ -66,31 +67,61 @@ public enum ReportingEntityType {
...
-        NO_FOLLOW_UP,
-        OPEN_CASE,
-        SURVEILLANCE_FOLLOW_UP
+        UNASSIGNED("UNASSIGNED"),
+        AWAITING_INTERVIEW("AI"),
+        CLOSED_CASE("CC"),
+        FIELD_FOLLOW_UP("FF"),
+        NO_FOLLOW_UP("NF"),
+        OPEN_CASE("OC"),
+        SURVEILLANCE_FOLLOW_UP("SF");
```
mpeels, line 76: lol why didn't i do this to begin with?
🌟🌟🌟🌟🌟

mpeels (approved):

PR #564, opened 2023-09-21
CNFT2-1095 Update Page Condition List Copy

Emelias-Enquizit:


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageStateChanger.java
```diff
@@ -51,13 +58,20 @@ public PageStateResponse savePageAsDraft(Long id) {
             Optional<WaTemplate> result = templateRepository.findById(id);
             if (result.isPresent()) {
                 WaTemplate page = result.get();
+                if(page.getTemplateType().equals(PageConstants.PUBLISHED_WITH_DRAFT))  {
+                response.setMessage(PageConstants.SAVE_DRAFT_NOCHANGE);
+                response.setTemplateId(page.getId());	
+                }
```
mpeels, line 64: I think we would want to throw a  `BadRequest` exception of some kind here so that the status code returned by the API is non 200.  Maybe a `PageUpdateException`?

Emelias-Enquizit: Okay sure  I will update

Emelias-Enquizit: Hi @mpeels  updated.

mpeels (approved):

PR #570, opened 2023-09-22
sorting by date and isolating logic with its test file and sharing 

hclarkEnq:
sorting by date and isolating logic with its test file and sharing across tables


adamloup-enquizit (approved):
📆 ⬆️

PR #579, opened 2023-09-26
- adding autocomplete input component and implementing into event search

hclarkEnq:


apps/modernization-ui/src/components/AutoCompleteInput/AutoCompleteInput.tsx
```diff
@@ -0,0 +1,26 @@
+import { ComboBox, Label } from '@trussworks/react-uswds';
+import { InputMaybe } from 'generated/graphql/schema';
+
+type AutoCompleteProps = {
+    label: string;
+    name: string;
+    value: InputMaybe<string> | undefined;
+    onChange: (event: any) => void;
+    options: any;
```
adamloup-enquizit, line 9: `SelectInput`  defines options as `{ name: string; value: string }`, let's try to keep them the same

adamloup-enquizit (changes requested):

adamloup-enquizit (approved):

PR #581, opened 2023-09-26
Project configuration

adamloup-enquizit:
Configures the gradle projects to ensure that Java 17 is used in local development.

### DIFF
--- a/.github/workflows/Build-and-deploy-modernization-api.yaml
+++ b/.github/workflows/Build-and-deploy-modernization-api.yaml
@@ -25,6 +25,7 @@ jobs:
       microservice_name: modernization-api
       dockerfile_relative_path: -f ./apps/modernization-api/Dockerfile .
       environment_classifier: SNAPSHOT
+      java_version: 17
     secrets:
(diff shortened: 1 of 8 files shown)


.github/workflows/Build-and-deploy-modernization-api.yaml
```diff
@@ -25,6 +25,7 @@ jobs:
       microservice_name: modernization-api
       dockerfile_relative_path: -f ./apps/modernization-api/Dockerfile .
       environment_classifier: SNAPSHOT
+      java_version: 17
```
szamfir, line 28: java_version should be a string. 
      java_version: '17'

.github/workflows/Build-and-deploy-nbs-gateway.yaml
```diff
@@ -13,6 +13,7 @@ jobs:
       microservice_name: nbs-gateway
       dockerfile_relative_path: -f ./apps/nbs-gateway/Dockerfile .
       environment_classifier: SNAPSHOT
+      java_version: 17
```
szamfir, line 16: java_version should be a string. 
      java_version: '17'

szamfir (changes requested):
Meant to do this first, see comments about java_version. In short it should be in quotes, since the value is a number (may or may not make a difference)

szamfir (approved):
LGTM. Not sure if Michael needs to approve as well

PR #590, opened 2023-09-28
[FIX] MessageModal causes blank page

adamloup-enquizit:
## Description

When attempting to delete a patient from the Patient Profile a message appears in a modal if the Patient cannot be deleted.  A change to the `.usa-modal__close` style was causing an error that lead to a blank page being shown.  The cause of the error was that there was no component in the modal to focus on since the only component in the `MessageModal` that could have focus was the `X` icon in the upper right.

## Steps to verify

1. Search for a Patient associated with events
1. Go to the Patient Profile
1. Attempt to delete the Patient


adamloup-enquizit:
@rkrusselenq I am not sure of the impact of removing this style on Page Builder so including you on here.  The `X` can be removed from the `Modal` component by setting the `forceAction` property.

PR #598, opened 2023-10-02
CNFT1-1532 added auto slash for date

GangulaMounika:
## Description

Added auto slash for date using JS since the trussworks didn't provide any masked input for dates

## Tickets

* https://cdc-nbs.atlassian.net/browse/CNFT1-1533

## Steps to verify

1. Manually type numbers in date picker it should automatically add slashes to validate date.


adamloup-enquizit (changes requested):
The `DatePickerInput component tests › when a valid date value is entered › should clear the previous error state` test is failing.

The `DatePickerInput` allows a date to be type without leading zeros for example, `1/1/2021` and `01/01/2021` are both valid.  The mask added is adding an extra `/` when typing date with leading zeros.  

result of typing `1/1/2022`

Screen Recording 2023-10-02 at 5 35 10 PM

GangulaMounika:
> The `DatePickerInput component tests › when a valid date value is entered › should clear the previous error state` test is failing.
> 
> The `DatePickerInput` allows a date to be type without leading zeros for example, `1/1/2021` and `01/01/2021` are both valid. The mask added is adding an extra `/` when typing date with leading zeros.
> 
> result of typing `1/1/2022`
> 
> Screen Recording 2023-10-02 at 5 35 10 PM [ Screen Recording 2023-10-02 at 5 35 10 PM ](https://user-images.githubusercontent.com/124325935/272104935-c200e0e7-f178-46e5-b0d0-8fed8c5687ea.gif) [ ](https://user-images.githubusercontent.com/124325935/272104935-c200e0e7-f178-46e5-b0d0-8fed8c5687ea.gif)

Done @adamloup-enquizit . Good catch. Thank you!

adamloup-enquizit (approved):

PR #606, opened 2023-10-05
remove PHIN API call

mpeels:
## Description

The PHIN VADS API tests are calling the external API. If the API is down, this will cause the build to fail. As a temporary fix, the test is being removed

## Tickets

N/A


adamloup-enquizit (approved):

mpeels:
Something goofy going on with the merge. Creating a new PR

PR #616, opened 2023-10-11
Disable Compare Investigations button based on permissions

GangulaMounika:
## Description

Disable/Enable Compare Investigations button based on permissions

## Tickets

* [[Jira Ticket](url)](https://cdc-nbs.atlassian.net/browse/CNFT1-1499)

## Steps to verify

 Open the NBS test env url in safari browseR: 

Enter user name as state and click on submit button

on the main header section on Home Screen, click on the open investigations link

Click on any of the patient name that is [Dupont, Ishmael](https://app.test-cdc-nbs.eqsandbox.com/nbs/LoadNavbar.do?ContextAction=GlobalInvestigations&initLoad=true#)

click on events tab

select the checkboxes beside 2 investigations and observe that compare investigations button should be disabled now if user doesn't have permissions.


adamloup-enquizit (changes requested):
Added a comment to Jira ticket about a rule of the design system to only show actions when a user is allowed to do that action.  This was missed when reviewing the bug in refinement.

Please change it to only show the button if the user has the appropriate permissions.

adamloup-enquizit (approved):

adamloup-enquizit (approved):

PR #642, opened 2023-10-19
Update deployment from dev to int1

mpeels:
## Description

Int1 will now be replacing the DEV environment. This updates the workflows to deploy there instead.

## Tickets

NA

## Steps to verify

NA


szamfir (approved):
LGTM, note nifi-efs will not effect until we update the environment to point to this helm chart.

PR #649, opened 2023-10-23
Cnft1 1490 system management permissions

hclarkEnq:
Added the system management access hook to hide system management based on the user permission rules

https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT1/boards/112?selectedIssue=CNFT1-1490

## Steps to verify

1. Log in using MSA as user and create a new user with only "view only" permissions.
2. Login locally with that new user you just created
3. Observe how "system management" is hidden
4. Another way of observing the permissions is by looking at the "me" endpoint and seeing the list of permissions based on the user


apps/modernization-ui/src/permissions/useSystemManagementAccess.ts
```diff
@@ -0,0 +1,21 @@
...
+    'EPILINKADMIN-SYSTEM',
+    'VIEWELRACTIVITY-OBSERVATIONLABREPORT',
+    'SRTADMIN-SYSTEM',
+    'VIEWPHCRACTIVITY-CASEREPORTING',
+    'IMPORTEXPORTADMIN-SYSTEM',
+    'REPORTADMIN-SYSTEM',
+    'ALERTADMIN-SYSTEM'
+];
+
+const useSystemManagementAccess = (permissions: string[] | undefined | null) => {
```
adamloup-enquizit, line 13: This should be a hook that returns the permissions for the navigation bar, not just for System Management Access.  For now it should return an object with only the boolean property `systemManagementAccess`.

apps/modernization-ui/src/permissions/useSystemManagementAccess.ts
```diff
@@ -0,0 +1,21 @@
...
+    const hasAccess = useMemo(() => {
+        return permissions && permissions.some((permission) => SYSTEM_MANAGEMENT_PERMISSIONS.includes(permission));
```
adamloup-enquizit, line 15: The `some` and the `includes` causes multiple iterations over the `SYSTEM_MANAGEMENT_PERMISSIONS` array.  The `reduce` used by `usePatientProfilePermissions` is done to prevent that.

apps/modernization-ui/src/permissions/useSystemManagementAccess.ts
adamloup-enquizit, line 13: I'm on the fence out passing the permissions vs using `useUser`

adamloup-enquizit (changes requested):
The permissions hook should cover all of the permissions for the navigation bar, the only permission we are worrying about right now is `systemManagementAccess`.

apps/modernization-ui/src/permissions/useSystemManagementAccess.ts
hclarkEnq: will use it

apps/modernization-ui/src/permissions/useSystemManagementAccess.ts
hclarkEnq: your right we can directly use the useUser hook inside the custom hook

adamloup-enquizit (approved):

PR #656, opened 2023-10-26
Cnft2 1300/create condition library page

rkrusselenq:
## Description

This adds the conditions library page at /page-builder/manage/condition-library.  The table is searchable, sortable and has pagination along with the ability to create a new condition. (Note: the search api has a bug and is not returning the conditions)
## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT2/boards/99?assignee=712020%3A591f5469-427d-49e9-87dc-8b4df72a02b4&selectedIssue=CNFT2-1300)

## Steps to verify

1. Pull this branch.
2. navigate to page-builder
3. click 'Conditions' on the left
4. Verify functions of table sorting and pagination work


apps/modernization-ui/src/apps/page-builder/pages/ConditionLibrary/ConditionLibraryTable.tsx
```diff
@@ -0,0 +1,149 @@
...
+            { id: 3, title: condition.progAreaCd },
+            {
+                id: 4,
+                title: condition.familyCd
+            },
+            { id: 5, title: condition.coinfectionGrpCd },
+
+            { id: 6, title: condition.nndInd },
+            { id: 7, title: condition.investigationFormCd },
+            { id: 8, title: condition.statusCd }
```
mpeels, line 114: The status values are showing as `I` or `A`, should these be `Active` or `Inactive`? 
image

apps/modernization-ui/src/apps/page-builder/services/conditionAPI.ts
```diff
@@ -1,12 +1,38 @@
...
+) => {
+    const response = await ConditionControllerService.searchConditionsUsingPost({
+        authorization: token,
+        search,
+        page,
+        size,
+        sort
+    });
+    return response;
 };
```
mpeels, line 36: The auto generated method is returning `Page_Condition_ | any`, can we have this API method call return `Page_Condition_`. This way we can use the type provided. I am not sure why the generated code is using any. I think its an issue with the Java `Page` interface.

mpeels:
Looks like the last page is empty. Do we know if this is an off by 1 issue with the UI component or if the API is returning an inaccurate total count?

rkrusselenq:
> Looks like the last page is empty. Do we know if this is an off by 1 issue with the UI component or if the API is returning an inaccurate total count?

This is the api. Returns an extra page

apps/modernization-ui/src/apps/page-builder/context/ConditionsContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+    searchQuery: '',
+    setSearchQuery: () => {},
+    currentPage: 1,
+    setCurrentPage: () => {},
+    sortBy: 'conditionShortNm',
+    setSortBy: () => {},
+    sortDirection: 'asc',
+    setSortDirection: () => {},
+    pageSize: 10,
+    setPageSize: () => {},
```
mpeels, line 33: Page size doesn't seem to be hooked up to the selector in the UI

mpeels (changes requested):

apps/modernization-ui/src/apps/page-builder/context/ConditionsContext.tsx
rkrusselenq: So this is kind of an issue, i have a fix pushed here for now but we may want to look into it... or not?  For every table we have created a context for the page size, current page etc etc.  well all these contexts are just copied and pasted from each other with only a couple of different minor configurations.  The problem was that the page size selector was only looking at one of those contexts.  So i created a prop we can pass in from each table which fixes the issue.  But I still dont like how we have the same context copy/pasted 5 times

mpeels (approved):

PR #658, opened 2023-10-26
Workflows Modernization

adamloup-enquizit:
## Description

Updates the trigger for the Build and Deploy modernization api work flow to trigger only for changes to the `modernization-api` or any dependencies.

Updates the trigger for the sonar workflow to not trigger for PRs that only affect the `modernization-ui` due to our Sonar only being able to scan Java projects.


szamfir (approved):
LGTM

PR #662, opened 2023-10-27
Cnft1 1615 add patient button alignments

hclarkEnq:
Aligning button as per newest Design review changes 

CNFT1-1610 https://cdc-nbs.atlassian.net/browse/CNFT1-1610

Screenshot 2023-10-27 at 7 48 41 AM

1. Navigate to add new patient page
2. Observe Add patient button aligned correctly in center of component 
[](url)


apps/modernization-ui/src/pages/advancedSearch/AdvancedSearch.scss
```diff
@@ -49,6 +49,7 @@ button {
     border-bottom: 1px solid #e7e9e9;
     .add-patient-button {
         height: fit-content;
+        margin: 0 !important;
```
adamloup-enquizit, line 52: question:  Should this style be in the Advanced Search styles?  It's called `add-patient-button` but it seems more like a general button style.  Should it be in the `shared.scss` file?

apps/modernization-ui/src/pages/addPatient/AddPatient.tsx
```diff
@@ -245,11 +245,9 @@ const AddPatient = () => {
                             }}>
                             <div className="width-full text-bold flex-row display-flex flex-align-center flex-justify">
                                 New patient
-                                <div className="button-group">
-                                    <Button className="padding-x-3 add-patient-button" type={'submit'}>
-                                        Save changes
-                                    </Button>
-                                </div>
+                                <Button className="padding-x-3 add-patient-button" type={'submit'}>
```
adamloup-enquizit, line 248: nit: can you remove `padding-x-3`?  I am not sure where it came from, it's not defined in our styles nor the USWDS styles.

adamloup-enquizit (changes requested):
The alignment looks good, just a few suggestions to clean it up a bit.

apps/modernization-ui/src/pages/advancedSearch/AdvancedSearch.scss
hclarkEnq: Yeah you are right we can move it to the shared.scss because it is a general button

apps/modernization-ui/src/assets/shared.scss
```diff
@@ -19,3 +19,7 @@ a {
         text-decoration: underline;
     }
 }
+
+.usa-button {
```
adamloup-enquizit, line 23: issue: This is changing every button, checkout out Advanced Search.  What I meant by the previous comment that the `add-patient-button` exited in the Advanced Search styles but was used in another page.  This signals to me that this class is a global style that should be available to the entire application so it should exist in the `share.scss` file.

question: What is the reason for the `margin: 0`?

adamloup-enquizit (changes requested):
Checkout the issue raise above where this change is affecting all buttons.

apps/modernization-ui/src/assets/shared.scss
hclarkEnq: Yes so those styles are used in advanceSearch and in the addPatient page, I will move both those styles to the shared.scss instead of AdvancedSearch.scss. 

margin: 0 is needed to remove spaces from top and bottom to make it horizontally and vertically aligned as it has margin from the uswds class so Im overriding it using margin: 0

adamloup-enquizit (approved):
🥇 Nice!

PR #666, opened 2023-10-30
CNFT1-1529 Shrink DB to an arbitrary number of patients

stevegsa:
## Description

The physical size of the current production anonymized masters are too unwieldy (restore time, size, etc.).  This script pares down any database to an arbitrary size with the most performant method.  Since tables are very large any method using deletes would take forever to run. So we chose to rebuild from the ground up while satisfying any constraints (that the subset of patients requires) in the most performant manner as possible as follows:

1) analyze all schemas and identify largest schemas/tables and start with biggest and work our way down
2) for each schema remove all database constraints
3) rebuild tables without indexes (while maintaining copies of the old and creating new unique key names to prevent collisions)
4) bulk insert with select intos from old tables (fast since no indexes)
5) re-add/build indexes (with new unique names to avoid collisions)
6) rename tables (switch old and new)
7) remove old tables

## Tickets

* [CNFT1-1529](https://cdc-nbs.atlassian.net/browse/CNFT1-1529)

## Steps to verify

1. Run the SQL on the db and follow directions for manual steps
2. Remove all elastic search indexes
3. Reindex with nifi
4. Test front end

[CNFT1-1529]: https://cdc-nbs.atlassian.net/browse/CNFT1-1529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cdc-sandbox/db/shrink.sql
+++ b/cdc-sandbox/db/shrink.sql
@@ -1,0 +1,831 @@
+-- THE FOLLOWING SQL WILL SHRINK ANY DATABASE TO AN ARBITRARY NUMBER OF PATIENTS
+
+-- THESE WERE THE SIZES FOR EACH SCHEMA BASED ON A FULL CDC ANONYMIZED DB:
+
+-- NBS_DataIngest 16M
+-- NBS_MSGOUTE 153M
+-- NBS_ODSE 1.6T <- EXTREMELY LARGE
+-- NBS_SRTE 285M
+-- RDB 1.6T <- EXTREMELY LARGE
+
+-- FOLLOWING QUERY LISTS LARGEST TABLE SIZES IN THE SCHEMA:
+  select schema_name(tab.schema_id) + '.' + tab.name as [table],
+    cast(sum(spc.used_pages * 8)/1024.00 as numeric(36, 2)) as used_mb,
+    cast(sum(spc.total_pages * 8)/1024.00 as numeric(36, 2)) as allocated_mb
+from sys.tables tab
+    inner join sys.indexes ind 
+        on tab.object_id = ind.object_id
+    inner join sys.partitions part 
+        on ind.object_id = part.object_id and ind.index_id = part.index_id
+    inner join sys.allocation_units spc
+        on part.partition_id = spc.container_id
+group by schema_name(tab.schema_id) + '.' + tab.name
+order by sum(spc.used_pages) desc
+
+-- LARGEST TABLES IN NBS_ODSE:
+-- |table                                  |used_mb  |allocated_mb|
+-- |---------------------------------------|---------|------------|
+-- |dbo.Observation                        |66,942.38|66,943.86   |
+-- |dbo.Participation                      |54,708.93|54,729.25   |
+-- |dbo.Entity_id                          |49,674.71|49,676.54   |
+-- |dbo.Person                             |46,704.04|46,707.88   |
+-- |dbo.Entity_locator_participation       |39,826.88|39,828.16   |
+-- |dbo.Act_id                             |33,323.13|33,323.6    |
+-- |dbo.Entity_loc_participation_hist      |28,392.43|28,392.55   |
+-- |dbo.Act_relationship                   |27,324.23|27,327.36   |
+-- |dbo.Person_name                        |23,338.17|25,214.56   |
+
+USE NBS_ODSE;
+
+    SELECT  'alter table  [' + tab1.name  + '] drop constraint [' + obj.name  + '];' + char(13)
+FROM sys.foreign_key_columns fkc
+INNER JOIN sys.objects obj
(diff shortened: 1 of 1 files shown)


cdc-sandbox/db/shrink.sql
```diff
@@ -0,0 +1,845 @@
...
+	status_time datetime NOT NULL,
+	to_time datetime NULL,
+	use_cd varchar(20) COLLATE SQL_Latin1_General_CP1_CI_AS NULL,
+	user_affiliation_txt varchar(20) COLLATE SQL_Latin1_General_CP1_CI_AS NULL,
+	valid_time_txt varchar(100) COLLATE SQL_Latin1_General_CP1_CI_AS NULL,
+	as_of_date datetime NULL,
+	CONSTRAINT PK_Entity_loc_particip_hist2 PRIMARY KEY (entity_uid,locator_uid,version_ctrl_nbr)
+);
+
+INSERT INTO Entity_loc_participation_hist2
```
adamloup-enquizit, line 777: question: Does it make sense to name the `Person2` something like `Person_shrunken` and `Person3` table as `Person_original`?

adamloup-enquizit (commented):
If the intention of this is to shrink a database for our uses, should we just truncate all of the `_hist` tables?

cdc-sandbox/db/shrink.sql
stevegsa: only reason i kept it short is because the suffix is used for the new constraint/indexes and key names and they have a size limit but we should be ok

cdc-sandbox/db/shrink.sql
```diff
@@ -0,0 +1,845 @@
...
+-- |dbo.Person                             |46,704.04|46,707.88   |
+-- |dbo.Entity_locator_participation       |39,826.88|39,828.16   |
+-- |dbo.Act_id                             |33,323.13|33,323.6    |
+-- |dbo.Entity_loc_participation_hist      |28,392.43|28,392.55   |
+-- |dbo.Act_relationship                   |27,324.23|27,327.36   |
+-- |dbo.Person_name                        |23,338.17|25,214.56   |
+
+USE NBS_ODSE;
+
+    SELECT  'alter table  [' + tab1.name  + '] drop constraint [' + obj.name  + '];' + char(13)
```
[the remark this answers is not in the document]
stevegsa: I was never a fan of having these constraints in the db to begin with.  So one of my assumptions is that the system still behaved as normal when they are removed.  I guess a good check would be to see if all the references are correct after the data is culled.  I'm on the fence with this.

stevegsa:
> If the intention of this is to shrink a database for our uses, should we just truncate all of the `_hist` tables?  I wasn't 100% on the second order effects of doing this.  Are you 100% we won't break something?

cdc-sandbox/db/shrink.sql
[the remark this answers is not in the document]
mpeels: I like constraints :) keeps me from doing dumb things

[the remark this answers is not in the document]
stevegsa: Ok the ayes have it ;-)

cdc-sandbox/db/shrink.sql
```diff
@@ -0,0 +1,885 @@
...
+-- |dbo.Entity_loc_participation_hist      |28,392.43|28,392.55   |
+-- |dbo.Act_relationship                   |27,324.23|27,327.36   |
+-- |dbo.Person_name                        |23,338.17|25,214.56   |
+
+USE NBS_ODSE;
+
+-- Disable constraints for all tables in the database:
+EXEC sp_msforeachtable 'ALTER TABLE ? NOCHECK CONSTRAINT ALL'
+
+CREATE TABLE NBS_ODSE.dbo.Person_shrunken (
```
adamloup-enquizit, line 57: Each Patient will have a row in the `Entity` table with a `class_cd` of `PSN` and an `entity_uid` equal to  `person.person_uid`.

adamloup-enquizit (commented):
A Patient can also be referenced in the `ct_contact` table as either a `contact_entity_uid` or `subject_entity_uid`

cdc-sandbox/db/shrink.sql
stevegsa: Ok is there something I need to do on this one?  Entity table wasn't shrunken because it was about 26th in the list of table sizes and not too large.

adamloup-enquizit: There is an `Entity` row for each Patient, if we are removing patients we should remove the `Entity` also.  Not sure if leaving the rows in the `Entity` table will have an impact or not.

stevegsa:
Shrinking db put on hold.  Closing this PR for now

PR #669, opened 2023-10-30
Cnft1 1588 1584 1585 1593 patient design UI fixes

GangulaMounika:
## Description
Updated UI inconsistencies based on design requirements

## Tickets

* https://cdc-nbs.atlassian.net/browse/CNFT1-1588
* https://cdc-nbs.atlassian.net/browse/CNFT1-1584
* https://cdc-nbs.atlassian.net/browse/CNFT1-1585
* https://cdc-nbs.atlassian.net/browse/CNFT1-1593

## Steps to verify

1) CNFT1-1593- In Partient search Sort BY button to not be shown ”when there are no results”. So Sort By won’t be shown in following scenarios:
* When you click clear all 
* When there are no results after a search 
* When the results are loading when switching between pages - even in that case there are no results and hence nothing to sort.

2) CNFT1-1584- In patient search Padding should be consistent on all for sides and match the design.

3) CNFT1-1588- show paginated count correctly (i.e. showing 1-25 of 10000, showing 26-50 of 10000)

4) CNFT1-1585- In Patient Search - Patient ID should be uppercase since it's an acronym.


apps/modernization-ui/src/pages/advancedSearch/AdvancedSearch.spec.tsx
```diff
@@ -25,19 +25,4 @@ describe('AdvancedSearch component tests', () => {
             expect(btn.hasAttribute('disabled'));
         });
     });
-
```
GangulaMounika, line 28: Looks like some dummy unit test case without any expectations. Deleted it.

PR #670, opened 2023-10-30
Cnft2 1202 reorder drag and drop

jlee-enquizit:
Here we have implemented the drag drop functionality in the Edit Page > Reorder Modal.

You will have to install the `react-beautiful-dnd` library.

You should be able to reorder items within pages, sections, subsections as well as drag questions, subsections, sections into other sections.

DragDropProvider has some methods for detecting collisions and dragging for styling to come.
The context provider will allow us to implement drag drop into the Edit Page views quickly.


mpeels:
There is a warning from the dnd library when reordering:
image


Not sure if this is an issue.

jlee-enquizit:
> There is a warning from the dnd library when reordering: image
> 
> Not sure if this is an issue.

I believe a lot of these dnd logs are hidden on production builds. If it works ok, we're probably fine.

mpeels (approved):

PR #676, opened 2023-11-01
INT1 Remove sort filter Page Library

jlee-enquizit:
Here we fix the search in Pages Library:
now when you clear the search terms, the page list resets

### DIFF
--- a/apps/modernization-ui/src/apps/page-builder/components/TableMenu/TableMenu.tsx
+++ b/apps/modernization-ui/src/apps/page-builder/components/TableMenu/TableMenu.tsx
@@ -34,6 +34,12 @@ export const TableMenu = ({ tableType, searchQuery, setSearchQuery, onDownloadIc
         }
     }, [searchQuery]);
 
+    useEffect(() => {
+        if (!keywords || keywords === '') {
+            setSearchQuery('');
+        }
+    }, [keywords]);
+
     return (
         <div className="table-menu">
             <Input


apps/modernization-ui/src/apps/page-builder/components/TableMenu/TableMenu.tsx
```diff
@@ -34,6 +34,12 @@ export const TableMenu = ({ tableType, searchQuery, setSearchQuery, onDownloadIc
         }
     }, [searchQuery]);
 
+    useEffect(() => {
+        if (!keywords || keywords === '') {
+            setSearchQuery('');
+        }
+    }, [keywords]);
```
adamloup-enquizit, line 41: question: Should this case be handled in the on-click?  it's changing search query without having to hit submit, not sure if that is desired behavior or not.

mpeels:
> Here we fix the search in Pages Library: now when you clear the search terms, the page list resets

PR is missing 
1. Link to Jira ticket
2. Steps to verify

jlee-enquizit:
[CNFT2-1336](https://cdc-nbs.atlassian.net/browse/CNFT2-1336)

apps/modernization-ui/src/apps/page-builder/components/TableMenu/TableMenu.tsx
jlee-enquizit: it's setting the query, but it won't search until 'Search' is clicke.
It searches for '' when the user clicks 'X' to clear the searchQuery, search is run automatically only then.

mpeels:
Should the `onClick` for the `X` just re-run the search?

jlee-enquizit:
> Should the `onClick` for the `X` just re-run the search?

The X is inserted via a prop, we don't have access to add an `onClick` function to it. 
We had to find another way to run it which, in this case, is when the search field is empty.

adamloup-enquizit (approved):

PR #677, opened 2023-11-01
CNFT2-1177-delete-static-element

alaapbharadwaj:
## Description

Delete Static Element API, given a specific ID for a static element on the page.

## Tickets

* [CNFT2-1177](https://cdc-nbs.atlassian.net/browse/CNFT2-1177)

## Steps to verify

1. Using swagger, create a static element of each type
2. Then use that componentID returned as the response to create a request for the delete static element 
3. If you have NBS running locally, you can manually check the database or launch the NBS classic UI and edit the corresponding page to then see whether the static elements created exist

[CNFT2-1177]: https://cdc-nbs.atlassian.net/browse/CNFT2-1177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/staticelement/PageStaticDeletor.java
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+
+        WaTemplate template = entityManager.find(WaTemplate.class, pageId);
+
+        if (template == null) {
+            throw new DeleteStaticElementException("Please provide valid page ID");
+        }
+
+        if (!template.getTemplateType().equals("Draft")) {
```
mpeels, line 37: nit (non-blocking): its good to do the comparison in the reverse order just in case the templateType is null.
`"Draft".equals(template.getTemplateType());`

mpeels, line 37: Issue: Thrown exceptions are missing test coverage:

image

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/staticelement/PageStaticDeletor.java
alaapbharadwaj: Can pages only have a type of "Drafts"?

mpeels (approved):

PR #678, opened 2023-11-02
Page-report report query

sizatenquizit:
## Description

Updated the query in the template repository (Modern) so that the PDF and CSV downloads match those in Classic.

## Tickets

* https://cdc-nbs.atlassian.net/browse/CNFT2-1225
* https://cdc-nbs.atlassian.net/browse/CNFT2-1224

## Steps to verify

1. Create Page Library PDF and CSV documents from Classic.
2. Create the same using Modern API.
3. Both PDFs and both CSVs should match.

### DIFF
--- a/apps/modernization-ui/src/utils/ExportUtil.ts
+++ b/apps/modernization-ui/src/utils/ExportUtil.ts
@@ -76,7 +76,7 @@ export const downloadLabReportSearchResultPdf = (labReportFilter: LabReportFilte
 
 export const downloadPageLibraryPdf = (token: string) => {
     // auto generated methods dont allow direct conversion to blob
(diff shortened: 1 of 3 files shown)


mpeels (changes requested):
1. The UI path for the download button is incorrect after the addition of `/nbs` to the pagebuilder path. Can you update `ExportUtil.ts` to have the path as follows:
image


2. The current export is showing the time in the improper format for the PDF version
Modernized export PDF: 
image

Classic export PDF:
image


The formatting should be updated to `MM/dd/yyyy`

mpeels (changes requested):
[4 Code Smells](https://sonarcloud.io/project/issues?id=CDCgov_NEDSS-Modernization&pullRequest=678&resolved=false&types=CODE_SMELL)

mpeels (approved):

PR #679, opened 2023-11-02
add-error-handling-to-page-library

rkrusselenq:
## Description
This PR adds error handling to the page library api requests.  It also cleans up some of the copy/paste related to the pagination contexts for each table, and cleans up some errors from using a hook inside of a hook within the RangeToggle component.

## Steps to verify

1. Use the page size toggle on any table and ensure there are no errors in the console.
2. Manually hard code something into the page search api call that will cause an error. Ensure that the page does not crash and the error is logged in the console.  We will need to add a banner or something in the future for displaying an error to the user.

### DIFF
--- a/apps/modernization-ui/src/apps/page-builder/pages/ManagePages/ManagePagesTable.tsx
+++ b/apps/modernization-ui/src/apps/page-builder/pages/ManagePages/ManagePagesTable.tsx
@@ -15,7 +15,7 @@ import { downloadPageLibraryPdf } from 'utils/ExportUtil';
 export enum Column {
     PageName = 'Page name',
     EventType = 'Event name',
-    RelatedConditions = 'Related conditions',
+    RelatedConditions = 'Related condition(s)',
     Status = 'Status',
     LastUpdate = 'Last updated',
     LastUpdatedBy = 'Last updated by'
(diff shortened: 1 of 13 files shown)


mpeels:
Build failure

apps/modernization-ui/src/apps/page-builder/pages/ManagePages/ManagePagesTable.tsx
```diff
@@ -91,6 +91,9 @@ export const ManagePagesTable = ({ summaries, currentPage, pageSize, totalElemen
                 case Column.Status:
                     setSortBy('status');
                     break;
+                case Column.RelatedConditions:
+                    setSortBy('conditions');
+                    break;
```
mpeels, line 96: Sorting by conditions is not yet supported on the API. We need to have a discussion with the design team to ensure this is something we want to do. 

If a page is associated with condition `aaa` and condition `zzz`. No matter the direction of the sort, this page will be first. I have concerns over the usefulness of sorting on a multi-valued field.

If you have time, you could connect with Chang or Jared and verify the expected behavior

rkrusselenq: Yeah, i can remove this if we decide not to allow the sorting. for now I have the config set to false so that the sorting arrow will not appear.

rkrusselenq: I will follow up with designers

mpeels (approved):

PR #686, opened 2023-11-02
CNFT2-1329 pagination lost on navigation

rkrusselenq:
## Description

On the pages table, if you use the pagination, and refresh or click into a page and use the breadcrumbs to navigate back, the page number is lost and you are returned to the first page.  I used url search params to hold the current page so that when refreshing or navigating back, the page number is restored to the last page viewed.
## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT2/boards/99?assignee=712020%3A591f5469-427d-49e9-87dc-8b4df72a02b4&selectedIssue=CNFT2-1329)

## Steps to verify

2. Use the pagination links to navigate to any page > 1.
3. Verify the page number is stored in the URL.
4. Click a page from the table.
5. At the top, click the  <- Page Library link.
6. Verify the last page was restored and the URL still holds the last page number.
7. Refresh browser and verify the url and current page still persist.


mpeels:
Build is failing

apps/modernization-ui/src/apps/page-builder/pages/ManagePages/ManagePagesTable.tsx
```diff
@@ -41,6 +41,8 @@ export const ManagePagesTable = ({ summaries, currentPage, pageSize, totalElemen
     const [tableRows, setTableRows] = useState<TableBody[]>([]);
     const { searchQuery, setSearchQuery, setCurrentPage, setSortBy, setSortDirection } = useContext(PagesContext);
     const { state } = useContext(UserContext);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [searchParams, setSearchParams] = useSearchParams();
```
mpeels, line 45: Question: I don't really like disabling eslint. Does `const setSearchParams = useSearchParams()[1]` work?  Or is there a more elegant solution

mpeels (changes requested):

apps/modernization-ui/src/apps/page-builder/pages/ManagePages/ManagePagesTable.tsx
rkrusselenq: Unfortunately that doesn't work. I can do this though...
```
const searchParams = useSearchParams();
const setSearchParams = searchParams[1];
```

mpeels (approved):

PR #692, opened 2023-11-06
Cnft1 1583 1612

hclarkEnq:
https://cdc-nbs.atlassian.net/browse/CNFT1-1583
In advanced search page the patient search and event search tabs for visual consistency the highlight should not lead to the text being lifted, but instead the text should remain consistent across the line
Increasing transparent border in Advanced Search to ensure the switching of tabs does not cause a movement of the rest of the elements, that is because by increasing the transparent border that was previously 5px to 7px it ensures that when you click between patient search and event search that no visible movement is created.

Aligning Data Entry - Fields should be horizontally aligned -> https://cdc-nbs.atlassian.net/browse/CNFT1-1612

Steps to verify
1 .Login
2. Navigate to advanced search
3. Click repeatedly between Patient search and Event search tabs
4. Observe no movement with the change of the two tabs


apps/modernization-ui/src/pages/patient/search/result/PatientResult.tsx
```diff
@@ -119,7 +119,7 @@ const ResultItem = ({ label, orientation = 'horizontal', children }: ResultItemP
 type ResultItemLabelProps = { children: string };
 
 const ResultItemLabel = ({ children }: ResultItemLabelProps) => (
-    <span className="patient-search-result-item-label">{children}</span>
+    <span className="patient-search-result-item-label font-sans-3xs">{children}</span>
```
adamloup-enquizit, line 122: question:  Why was the font size here changed?  this value seems small.

apps/modernization-ui/src/pages/advancedSearch/AdvancedSearch.scss
```diff
@@ -23,7 +23,7 @@ button {
 }
 
 .type {
-    border-bottom: 5px solid transparent;
+    border-bottom: 7px solid transparent;
```
adamloup-enquizit, line 26: issue:  this height should be `height: 0.4375rem;`

apps/modernization-ui/src/pages/addPatient/components/contactFields/ContactFields.tsx
```diff
@@ -144,7 +144,7 @@ export default function ContactFields({ id, title }: Props) {
                             }
                             className="text-bold"
                             unstyled
-                            style={{ margin: '0', padding: '0' }}>
+                            style={{ margin: '10px 0 0 0', padding: '0' }}>
```
adamloup-enquizit, line 147: issue:  This gap should be set to 1.125rem.  Also, this is a great candidate for a refactor.  All of the elements within the card should have a gap of 1.125rem, I think the use of `Grid` is complicating the styles.

adamloup-enquizit (changes requested):

apps/modernization-ui/src/pages/patient/search/result/PatientResult.tsx
hclarkEnq: I changed that font size to ensure for the children that we get the same font size across the summary card for the headers making it smaller but visually thought it looked good, will play around with a bigger size to see if doable.

apps/modernization-ui/src/pages/addPatient/components/contactFields/ContactFields.tsx
hclarkEnq: okay will look into it thanks

adamloup-enquizit (approved):

PR #693, opened 2023-11-06
defining profile in useProfileContext to get the latest profile data …

hclarkEnq:
Description 
Fixing a bug in the patient profile where the patients data was not being reflected immediately upon edits or changes done in the demographics card using any table or card where the information is changeable.

## Steps to verify

1. Login
1. Navigate to any patient profile after searching for a patient
2. navigate to demographics and edit to change the "name" of the patient
3. observe the changes reflected immediately in the summary card in the patient profile at the top of the page


apps/modernization-ui/src/pages/patient/profile/ProfileContext.tsx
```diff
@@ -19,9 +20,9 @@ export function useProfileContext() {
 }
 
 export const ProfileProvider = ({ children, id }: any) => {
-    const { refetch } = usePatientProfile(id);
+    const { refetch, profile } = usePatientProfile(id);
 
     const changed = () => refetch();
 
-    return <ProfileContext.Provider value={{ changed }}>{children}</ProfileContext.Provider>;
+    return <ProfileContext.Provider value={{ changed, profile }}>{children}</ProfileContext.Provider>;
```
adamloup-enquizit, line 27: issue:  combining `patient` and `summary` into one object will cause the entire page to re-render.  An alternative approach would be to have `usePatientProfile`return the `patient` and `summary` as two separate items in `PatientProfile.tsx`

`const {patient, summary} = usePatientProfile()`

adamloup-enquizit (changes requested):
Putting the `patient` in the ProfileContext is okay however, adding the profile will cause the entire page to re-render when the summary changes.

apps/modernization-ui/src/pages/patient/profile/ProfileContext.tsx
hclarkEnq: got it

apps/modernization-ui/src/pages/patient/profile/addresses/AddressesTable.tsx
```diff
@@ -129,17 +127,18 @@ export const AddressesTable = ({ patient }: Props) => {
...
+                variables: {
+                    patient: patient?.id,
+                    page: {
+                        pageNumber: currentPage - 1,
+                        pageSize: TOTAL_TABLE_DATA
+                    }
+                },
+                notifyOnNetworkStatusChange: true
+            });
+    }, [currentPage, patient]);
```
adamloup-enquizit, line 141: question: Is `patient` needed as a dependency? It's passed in and not part of the state of this component

adamloup-enquizit (approved):

apps/modernization-ui/src/pages/patient/profile/addresses/AddressesTable.tsx
hclarkEnq: Yes because patient.ID is being passed into the API

PR #688, opened 2023-11-03
[CNFT2-1163] Page Summary Search API

adamloup-enquizit:
## Description

Adds filter functionality to the POST `/page-builder/api/v1/pages/search` endpoint.  Filters can be added in the `filters` array of the request.  The properties `name`, `eventType`, `condition`, and `status` are currently supported.

### Single Value Filters

A filters the results based on an operator using a single `value`.  Accepts an `operator` value of either `equals`, `not-equal-to`, `contains`, or `starts-with`.

```json
{
    "filters":[{"property":"condition","operator":"contains","value":"type"}]
}
```

### Multi Value Filters

A filters the results based on an operator using a multiple `values`.  Accepts an `operator` value of either `equals`, `not-equal-to`, `contains`, or `starts-with`.

```json
{
    "filters":[{"property":"condition","operator":"contains","values":["type", "sec"]}]
}
```

## Tickets

* [CNFT2-1163](https://cdc-nbs.atlassian.net/browse/CNFT2-1163)

## Steps to verify

1. Import Postman collection with new Requests
2. Execute the requests in `Question Bank > Pages > Search`

[CNFT2-1163]: https://cdc-nbs.atlassian.net/browse/CNFT2-1163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #701, opened 2023-11-08
Resolve sorting issues on status column for pages library

rkrusselenq:
## Description

The sort direction was not being set properly when clicking the Status column header.  Also adds a bit of type safety for retrieving pages.
## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT2/boards/99?assignee=712020%3A591f5469-427d-49e9-87dc-8b4df72a02b4&selectedIssue=CNFT2-1397)

## Steps to verify

1. Navigate to page-builder/manage/pages
2. Sort by the status column.
3. Verify the fetch goes from desc, to asc, and back to none.
4. Switch pagination to higher pages and verify still functions properly

### DIFF
--- a/apps/modernization-ui/src/apps/page-builder/generated/models/Page_PageSummary_.ts
+++ b/apps/modernization-ui/src/apps/page-builder/generated/models/Page_PageSummary_.ts
@@ -7,16 +7,15 @@ import type { PageSummary } from './PageSummary';
 import type { Sort } from './Sort';
 
 export type Page_PageSummary_ = {
-    content?: Array<PageSummary>;
-    empty?: boolean;
-    first?: boolean;
-    last?: boolean;
-    number?: number;
-    numberOfElements?: number;
-    pageable?: Pageable;
-    size?: number;
-    sort?: Sort;
-    totalElements?: number;
-    totalPages?: number;
+    content: Array<PageSummary>;
+    empty: boolean;
+    first: boolean;
+    last: boolean;
+    number: number;
+    numberOfElements: number;
+    pageable: Pageable;
+    size: number;
+    sort: Sort;
+    totalElements: number;
+    totalPages: number;
 };
-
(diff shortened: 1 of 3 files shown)


apps/modernization-ui/src/apps/page-builder/generated/models/Page_PageSummary_.ts
```diff
@@ -7,16 +7,15 @@ import type { PageSummary } from './PageSummary';
...
+    empty: boolean;
+    first: boolean;
+    last: boolean;
+    number: number;
+    numberOfElements: number;
+    pageable: Pageable;
+    size: number;
+    sort: Sort;
+    totalElements: number;
+    totalPages: number;
```
mpeels, line 20: Info: Was this a manual change? If so this file is an auto-generated file. We would need to add `@ApiModelProperty(required = true)` to the corresponding java object and trigger the generation.

mpeels:
The underlying issue is due to the `Status` values in the database being stored as `IXS`, `INV`, 'CON' etc. The sort is being applied to the "Short" names which do not match the displayed names in the UI. 

I'm fairly certain Adam has resolved this issue as part of the Filter story.

apps/modernization-ui/src/apps/page-builder/generated/models/Page_PageSummary_.ts
rkrusselenq: Yeah this was a manual change, thanks for the info. I just kinda figured these values would always be returned.  Should I revert or is it ok to change the @apiModelProperty to true?

mpeels:
> > The underlying issue is due to the `Status` values in the database being stored as `IXS`, `INV`, 'CON' etc. The sort is being applied to the "Short" names which do not match the displayed names in the UI.
> > I'm fairly certain Adam has resolved this issue as part of the Filter story.
> 
> If thats the case, should we just scrap this PR and close the story?

I believe so, but let me confirm that this is fixed

mpeels:
Confirmed sorting is resolved as part of https://github.com/CDCgov/NEDSS-Modernization/pull/688. Closing this PR

PR #714, opened 2023-11-13
Page Info Panel History

sizatenquizit:
## Description

An API needs to be created to provide the history for a page.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1125

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageController.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageController.java
@@ -2,9 +2,11 @@ package gov.cdc.nbs.questionbank.page;
 
 import com.itextpdf.text.DocumentException;
 import gov.cdc.nbs.authentication.UserDetailsProvider;
+import gov.cdc.nbs.questionbank.page.model.PageHistory;
 import gov.cdc.nbs.questionbank.page.request.PageCreateRequest;
 import gov.cdc.nbs.questionbank.page.response.PageCreateResponse;
 import gov.cdc.nbs.questionbank.page.response.PageStateResponse;
+import gov.cdc.nbs.questionbank.page.service.PageHistoryFinder;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.core.io.InputStreamResource;
 import org.springframework.core.io.Resource;
@@ -21,8 +23,8 @@ import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
-
 import java.io.IOException;
+import java.util.List;
 
 @Slf4j
 @RestController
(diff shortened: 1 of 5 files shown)


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageController.java
```diff
@@ -86,4 +92,9 @@ public PageStateResponse deletePageDraft(@PathVariable("id") Long pageId) {
   }
 
 
+  @GetMapping("{template_nm}/page-history")
```
mpeels, line 95: Can we use {id} instead of `template_nm` to be consistent.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/service/PageHistoryFinder.java
```diff
@@ -0,0 +1,44 @@
...
+            " hist.version_note AS "+NOTES+" " +
+            " FROM WA_template_hist hist LEFT OUTER JOIN user_profile userProfile ON hist.last_chg_user_id = userProfile.nedss_entry_id " +
+            " WHERE hist.template_nm = ? " +
+            " UNION " +
+            " SELECT tem.publish_version_nbr AS "+PUBLISH_VERSION_NBR+"," +
+            " CONVERT(varchar, tem.last_chg_time, 101) AS "+LAST_UPDATE_DATE+"," +
+            " userProfile.first_nm || ' ' || userProfile.last_nm AS "+LAST_UPDATE_BY+"," +
+            " tem.version_note AS "+NOTES+" " +
+            " FROM WA_template tem LEFT OUTER JOIN user_profile userProfile ON tem.last_chg_user_id = userProfile.nedss_entry_id " +
+            " WHERE tem.template_nm = ? AND tem.template_type IN ('Published With Draft','Published')";
```
mpeels, line 31: Issues (blocking):
1. Lets use a multiline String here. Makes it a lot easier to read and to copy/paste into a db tool.
```java
"""
<your query here>
"""
```
2. What is the purpose of the `UNION`, do we need to convert time in this manner?  
3. Can we update the query to use a template Id instead of name to be consistent with other queries?

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/service/PageHistoryFinder.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public List<PageHistory> getPageHistory(String waTemplateName) {
+        Object[] args = new Object[]{waTemplateName, waTemplateName};
+        int[] argTypes = new int[]{Types.VARCHAR, Types.VARCHAR};
```
mpeels, line 35: Nit (non-blocking): using the `setter -> setter.setLong(1, wa_template_id),` is the preferred manner to pass arguments to a query.

mpeels (changes requested):

mpeels:
Duplicate of https://github.com/CDCgov/NEDSS-Modernization/pull/718

PR #718, opened 2023-11-14
Page Info Panel History

sizatenquizit:
## Description

An API needs to be created to provide the history for a page. 

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1125


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/service/PageHistoryFinder.java
```diff
@@ -0,0 +1,59 @@
...
+        WA_template_hist hist
+        LEFT OUTER JOIN user_profile userProfile ON hist.last_chg_user_id = userProfile.nedss_entry_id
+      WHERE
+        hist.wa_template_uid = ?
+      UNION
+      SELECT
+        tem.publish_version_nbr AS publishVersionNbr,
+        tem.last_chg_time AS lastUpdatedDate,
+        userProfile.first_nm || ' ' || userProfile.last_nm AS lastUpdatedBy,
+        tem.version_note AS notes
```
mpeels, line 37: Getting a syntax error for the `||`

image

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/service/PageHistoryFinder.java
```diff
@@ -0,0 +1,59 @@
...
+                    SELECT
+                      tem.publish_version_nbr AS publishVersionNbr,
+                      tem.last_chg_time AS lastUpdatedDate,
+                      CONCAT(userProfile.first_nm, ' ', userProfile.last_nm) AS lastUpdatedBy,
+                      tem.version_note AS notes
+                    FROM
+                      WA_template tem
+                      LEFT OUTER JOIN user_profile userProfile ON tem.last_chg_user_id = userProfile.nedss_entry_id
+                    WHERE
+                      tem.wa_template_uid = ?
```
adamloup-enquizit, line 42: issue:  When a page is published and a draft is created there are essentially two versions of the same Page, a published and a draft, each having their own identifier.  This means that `wa_template_uid` cannot be used to get all of the history.  The `template_nm` is the most unique common field.  

Something like...

```sql
with [target] ([name]) as (
    select 
        [target].template_nm
    from WA_template [target]
    where [target].[wa_template_uid] = ?
),
select
...
from [target]

    join wa_template_hist [history] on
        [target].[name] = [history].template_nm
...
UNION
select
...
from [target]

    join wa_template [page] on
        [page].template_nm = [target].[name]
```

suggestion:  The interesting part is that the API is passing only the identifier of the Page we want history for.  This can be handled in SQL using a CTE that grabs the `template_nm` of the Page using the `wa_template_uid` and then join the other two tables to it using the name.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/service/PageHistoryFinder.java
```diff
@@ -0,0 +1,59 @@
...
+                      AND tem.template_type IN('Published With Draft', 'Published');
+                        """;
+
+    public List<PageHistory> getPageHistory(Long pageId) {
+        return jdbcTemplate.query(QUERY, setter -> {
+            setter.setLong(1, pageId);
+            setter.setLong(2, pageId);
+        }, (rs, rowNum) -> new PageHistory(
+                rs.getString("publishVersionNbr"),
+                rs.getTimestamp("lastUpdatedDate").toLocalDateTime().format(dateFormatter),
```
adamloup-enquizit, line 52: suggestion (non-blocking):  This can also be done in the SQL statement to avoid having to transform in the mapper.  MS SQL has a `format` function available to format any data type.

adamloup-enquizit (commented):
Added a few SQL suggestions

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/service/PageHistoryFinder.java
```diff
@@ -0,0 +1,58 @@
...
+      with target as (
+                       SELECT
+                           template_nm as template_name
+                       FROM WA_template
+                       WHERE wa_template_uid = ?
+                   )
+      SELECT
+        hist.publish_version_nbr AS publishVersionNbr,
+        CONVERT(varchar, hist.last_chg_time, 101) AS lastUpdatedDate,
+        userProfile.first_nm || ' ' || userProfile.last_nm AS lastUpdatedBy,
```
mpeels, line 26: Issue: the `||` syntax is not supported in MSSQL. API fails to return any data due to `com.microsoft.sqlserver.jdbc.SQLServerException`

mpeels (changes requested):

mpeels (approved):

lramakr (approved):
LGTM

PR #722, opened 2023-11-14
Page Info Panel Metadata Button

sizatenquizit:
## Description

In the Classic NBS, the Excel File gets downloaded when the user taps on the “Page Metadata” button. The new API must follow the same interaction, and the same data and column should be generated in our modernized version. 

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1127


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageMetaDataDownloader.java
```diff
@@ -0,0 +1,195 @@
...
+            List<Object[]> pageMetadata = findPageMetadataByWaTemplateUid(waTemplateUid);
+            for (Object[] data : pageMetadata) {
+                csvPrinter.printRecord(data);
+            }
+            csvPrinter.flush();
+            return new ByteArrayInputStream(out.toByteArray());
+        } catch (IOException e) {
+            throw new IOException("Error downloading Page Metadata:" + e.getMessage());
+        }
+    }
```
mpeels, line 178: The `PageMetadata` created by Classic is a `.xls` file with multiple tabs of data.

image


The `.csv` produced by this endpoint is `108 KB` whereas `PageMetadata.xls` is 1.5 MB

image

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageController.java
```diff
@@ -61,26 +70,36 @@ public ResponseEntity<Resource> downloadPageLibrary() throws IOException {
...
   @DeleteMapping("{id}/delete-draft")
-  public PageStateResponse deletePageDraft(@PathVariable("id") Long pageId) {
-    return stateChange.deletePageDraft(pageId);
+  public PageDeleteResponse deletePageDraft(@PathVariable("id") Long pageId) {
+    return pageDeletor.deletePageDraft(pageId);
+  }
+
+  @GetMapping("{waTemplateUid}/download-metadata")
+  public ResponseEntity<Resource> downloadPageMetadata(@PathVariable("waTemplateUid") Long waTemplateUid) throws IOException {
+    String fileName = "PageMetadata.xls";
```
mpeels, line 97: Suggestion: NBS Classic created an `xls` while the newer library is creating a `xlsx` file.  Since the file the API is creating is `xlsx` but has `xls` extension, I am getting a warning when opening it. 

I recommend changing the file name to `xlsx`. This will differ from Classic but makes sense as that is the newer format.

mpeels (approved):

PR #726, opened 2023-11-15
Bug/cnft 1423/add sorting to quick condition modal

rkrusselenq:
## Description

Adds the ability to sort by column on the quick conditions modal.  Also had to do a little refactoring to keep QuickConditions and ConditionsLibraryTable more DRY
## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT2/boards/99?assignee=712020%3A591f5469-427d-49e9-87dc-8b4df72a02b4&selectedIssue=CNFT2-1423)

## Steps to verify

1. Navigate to page-builder/manage/pages
2. click Create new page.
3. click "search for condition(s)"
4. Verify sorting by columns works properly.


mpeels (commented):
Question: When clicking a header to sort, the modal shrinks and then grows as the content is removed and then re-added. 

Any good way we can prevent this?
annoyingSizeChange

rkrusselenq:
> Question: When clicking a header to sort, the modal shrinks and then grows as the content is removed and then re-added.
> 
> Any good way we can prevent this? 

Good catch. I gave it a scroll for the overflow and a fixed height.

mpeels:
> > Question: When clicking a header to sort, the modal shrinks and then grows as the content is removed and then re-added.
> > Any good way we can prevent this?
> 
> Good catch. I gave it a scroll for the overflow and a fixed height.

Better but now I'm getting double scroll bars :)

image

rkrusselenq:
> Better but now I'm getting double scroll bars :)
> 

Can't believe i missed that. fixed now.

mpeels (changes requested):
`Cancel` and `Create new condition` buttons are now inside the table
image

rkrusselenq:
> `Cancel` and `Create new condition` buttons are now inside the table 

Ok sorry had to do some rework but finally got it figured out i think.

apps/modernization-ui/src/apps/page-builder/components/QuickConditionLookup/QuickConditionLookup.tsx
```diff
@@ -58,61 +52,57 @@ export const QuickConditionLookup = ({ modal, addConditions }: Props) => {
...
+        const search: ReadConditionRequest = { searchText };
 
-            ConditionControllerService.searchConditionsUsingPost({
-                authorization,
-                search,
-                page: page,
-                size: pageSize
+        searchConditions(authorization, currentPage, pageSize, `${sortBy},${sortDirection}`, {
+            searchText: search.searchText
+        })
```
mpeels, line 62: Issue: the `currentPage` starts at 1 and corresponds to what the user sees. The API expects pages to start at 0. I think we need a `currentPage - 1` when performing a search

mpeels (changes requested):
Issue: Return of the double scroll bar

image



Issue: No conditions are selectable. This brings up a requirement that was never implemented. Only conditions that are **NOT** associated with a page should be selectable. So I believe we can update the `asTableRow` to have `selectable: condition.investigationFormCd == undefined,` 

image

rkrusselenq:
> Issue: Return of the double scroll bar
Weird im not seeing it on my screen. I'll keep hacking at it

rkrusselenq:
Bump

PR #740, opened 2023-11-20
CNFT2-1425-numeric-special-characters-fix-page-name

alaapbharadwaj:
## Description

Entering a full string of numbers and special characters as a Page name are accepted when clicking Create Page.

## Tickets

* [CNFT2-1425](https://cdc-nbs.atlassian.net/browse/CNFT2-1425)

## Steps to verify

2. Then click on Create new page
3. Scroll down to page name and verify only some special character work and not all

[CNFT2-1425]: https://cdc-nbs.atlassian.net/browse/CNFT2-1425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-ui/src/apps/page-builder/pages/AddNewPage/AddNewPageFields.tsx
```diff
@@ -50,12 +51,16 @@ export const AddNewPageFields = (props: AddNewPageFieldProps) => {
...
+                rules={{
+                    required: { value: true, message: 'Name is required.' },
+                    ...validPageNameRule
+                }}
                 render={({ field: { onChange, value }, fieldState: { error } }) => (
                     <Input
                         onChange={onChange}
                         defaultValue={value}
                         label="Page name"
+                        className="pageName"
```
mpeels, line 63: Issue: You will need to pass the `onBlur` callback into the Input so that the validation will trigger when the user tabs away from the input.

```typescript
 render={({ field: { onChange, onBlur, value }, fieldState: { error } }) => (
                    <Input
                        onBlur={onBlur}
                        onChange={onChange}
                        defaultValue={value}


```

apps/modernization-ui/src/validation/entry/validPageNameRule.ts
```diff
@@ -0,0 +1,14 @@
+import { maxLengthRule } from './maxLengthRule';
+
+/**
+ * Creates a standardized react-hook-form rule for the entry of name values.
+ */
+const validPageNameRule = {
+    ...maxLengthRule(),
+    pattern: {
+        value: /^[\w*()_+\-=;:/.,\d\s]+$/gm,
+        message: 'Only alphabets, digits, and some special character allowed such as ( ) _ + - = ; : / . ,'
```
mpeels, line 10: Nit/question (non-blocking): what do you think about: `Valid characters are A-Z, a-z, 0-9, or ()_+-=;:/.,`

apps/modernization-ui/src/validation/entry/validPageNameRule.ts
```diff
@@ -0,0 +1,14 @@
+import { maxLengthRule } from './maxLengthRule';
+
+/**
+ * Creates a standardized react-hook-form rule for the entry of name values.
+ */
+const validPageNameRule = {
+    ...maxLengthRule(),
+    pattern: {
+        value: /^[\w*()_+\-=;:/.,\d\s]+$/gm,
```
mpeels, line 9: Suggestion: i think this can be simplified slightly.
`\d` is equivalent to [0-9]
`\w` is equal to:  letters (A–Z, a–z), numbers (0–9), and underscore (_). [source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_class_escape#w).

So using both `\w` and `\d` is redundant. And `\w` includes underscore.

With that we could update the regex to be: `\w*()+\-=;:/.,\s`

mpeels (changes requested):

mpeels:
Issue: it also appears the `Create Page` button is not being disabled when the validation fails. May take a little digging to get this to work.

image

mpeels (approved):
🌟

PR #744, opened 2023-11-21
CNFT2-1460 Persist page size and pagination on pages library

rkrusselenq:
## Description

Previously we added query params to track page number, this PR adds the same functionality for the page size, but also moves it to the range toggle so that all tables with pagination will have this functionality.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT2/boards/99?assignee=712020%3A591f5469-427d-49e9-87dc-8b4df72a02b4)

## Steps to verify

2. Use the pagination alongside the page size selector.
3. Ensure when navigating between pages that the size persists.
4. Ensure when refreshing the page that both the size and page number persist.
5. Click into a page, and use the back button, ensure the previous page number and size remain.
6. When changing the size, the page number should revert to '1'


apps/modernization-ui/src/apps/page-builder/components/QuickConditionLookup/QuickConditionLookup.tsx
```diff
@@ -41,6 +41,7 @@ export const QuickConditionLookup = ({ modal, addConditions }: Props) => {
     const { state } = useContext(UserContext);
     const { currentPage, setCurrentPage, pageSize, setPageSize } = useContext(ConditionsContext);
 
+    console.log('test', conditions);
```
mpeels, line 44: Issue: `console.log`

mpeels (changes requested):
Issue: If we are using the query parameters to handle the paging, then there is a problem when switch to `page 2` and clicking the back button without leaving the page library. The query params are updated to show `page=1` but the table is not refreshed to match the parameter.

apps/modernization-ui/src/apps/page-builder/components/QuickConditionLookup/QuickConditionLookup.tsx
rkrusselenq: fixed

rkrusselenq: Ok i think i have it fixed with adding location to the useEffect, now we update the page number and page size whenever the query params change.

mpeels (changes requested):
Issue: When clicking the `Create new page` the query parameters are being added to the URL causing some issue with the back button.

image


Issue: When navigating to the 2nd page on the `Page Library`. 2 `back` button presses are required to go to page 1.

rkrusselenq:
> Issue: When clicking the `Create new page` the query parameters are being added to the URL causing some issue with the back button.

> Issue: When navigating to the 2nd page on the `Page Library`. 2 `back` button presses are required to go to page 1.

This was really tricky but I think I got it figured out finally.  I won't be surprised if you find another side effect lol

mpeels:
Issue (blocking): Clicking to change page does not load the page

rkrusselenq:
> Couple issues.
> 
> 1. can remove the commented out code
> 2. Clicking the page button doesn't actually change the page

Fixed, the commented out code was the issue.

achapm (changes requested):
Please merge with main and resolve any conflics

mpeels:
Closing PR, moved existing bug ticket to OBE as the functionality does not have formal requirements and will need to be added.

PR #745, opened 2023-11-21
Cnft1 749 skip link context hook

hclarkEnq:
## Description

Adding a context for the Skip Link, this is a invisible link that should wrap up the app and be the first tabbing option for the user 
The goal of this skip link is to navigate the user to main content of the page.

Note: There is an example of the usage of the skip link in advanced search for testing purposes
Proper usage of main content will be determined later

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT1-749

## Steps to verify

1.Login
2. Click on tab button
3. Observe the skip link focused
4. navigate to main content


apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const [id, setId] = useState<string | undefined>(undefined);
+
+    // Setting up the id to pass it to the anchor tag
+    const contextValue: SkipLinkContextType = {
+        setId
+    };
+
+    return (
+        <SkipLinkContext.Provider value={contextValue}>
+            <a href={'#' + id}>Skip to main content</a>
```
adamloup-enquizit, line 33: suggestion:  only display the skip link if `id` exists to avoid any possible confusion when the link doesn't actually go anywhere

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+        throw new Error('useSkipLink must be used within a SkipLinkProvider');
+    }
+    return context;
+}
+
+interface SkipLinkProviderProps {
+    children: React.ReactNode;
+}
+
+export const SkipLinkProvider: React.FC<SkipLinkProviderProps> = ({ children }: SkipLinkProviderProps) => {
```
adamloup-enquizit, line 23: question: What is `React.FC`?

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const [id, setId] = useState<string | undefined>(undefined);
+
+    // Setting up the id to pass it to the anchor tag
+    const contextValue: SkipLinkContextType = {
+        setId
```
adamloup-enquizit, line 28: question:  The value is an `id` but `setId` doesn't really express the intention of it.  What name would express the intention more?  `skipTo`?

hclarkEnq: Yeah skipTo makes more sense for this case.

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
hclarkEnq: okay will add the check

apps/modernization-ui/src/SkipLink/SkipLink.scss
```diff
@@ -0,0 +1,6 @@
+.hidden-link {
```
adamloup-enquizit, line 1: suggestion: should the class name match the Component name to keep things consistent?

apps/modernization-ui/src/SkipLink/SkipLink.scss
```diff
@@ -0,0 +1,6 @@
...
+    color: transparent;
```
adamloup-enquizit, line 2: issue:  The element is still visible as a giant bar on the top of the screen.  `display: none` might be enough

adamloup-enquizit (changes requested):

apps/modernization-ui/src/SkipLink/SkipLink.scss
hclarkEnq: I also deleted the styles as we don't need them

adamloup-enquizit (approved):

PR #746, opened 2023-11-22
List Valid Display Types

sizatenquizit:
## Description

The UI needs to be notified of what values are valid for the displayControl field.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1182

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/request/CreateQuestionRequest.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/request/CreateQuestionRequest.java
@@ -3,135 +3,45 @@ package gov.cdc.nbs.questionbank.question.request;
 import gov.cdc.nbs.questionbank.entity.question.CodeSet;
 import lombok.Getter;
 import lombok.Setter;
-import gov.cdc.nbs.questionbank.question.model.DisplayOption;
-import java.util.Arrays;
-import java.util.List;
 
 @Getter
 @Setter
 public abstract class CreateQuestionRequest {
-        private CodeSet codeSet;
-        private String uniqueId;
-        private String uniqueName;
-        private String subgroup;
-        private String description;
-        private String label;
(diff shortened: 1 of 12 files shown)


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/request/CreateQuestionRequest.java
```diff
@@ -35,4 +38,100 @@ public record MessagingInfo(
...
+                READONLY_MULTI_SELECT_NO_SAVE(1028, "Readonly multi-select no save"),
+                CODE_LOOKUP(1031, "Code Lookup");
+
+                private final long key;
+                private final String value;
+
+                CodedQuestionType(long key, String value) {
+                        this.key = key;
+                        this.value = value;
+                }
```
mpeels, line 56: Suggestion: If we change this to `long value, String name` then the UI can use the options directly in the `SelectInput` component without need to map them

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/request/CreateQuestionRequest.java
```diff
@@ -35,4 +38,100 @@ public record MessagingInfo(
...
+                public static List<DisplayOption> getDisplayOptions() {
+                        return Arrays.stream(values()).map(DateQuestionType::toDisplayOption).toList();
+                }
+
+                private DisplayOption toDisplayOption() {
+                        return new DisplayOption(key,value);
+                }
+        }
+
+        public enum NumericQuestionType {
```
mpeels, line 88: Suggestion: What do you think about renaming these to `NumericDisplayControl` as the value maps to the `displayControl` field.

Also, what do you think about moving the various enum's into the question requests? For example, moving this `NumericQuestionType` into the `CreateNumericQuestionRequest.java` class

mpeels (approved):

PR #760, opened 2023-11-30
[Workflow] Release canidate containers build and deploymeny preparation

adamloup-enquizit:
An attempt at automating the creation of the individual containers that make up a release candidate for the NBS Modernization.  This workflow is meant to automate the tedious steps of building/tagging each release candidate container.

Includes the container build and helm chart update for each application that is part of the release candidate.

- NBS Gateway
- Modernization API
- Elasticsearch
- Nifi

Limits the environment selection to just TEST (test2) for now.

### DIFF
--- a/.github/workflows/Release-candidate-containers.yml
+++ b/.github/workflows/Release-candidate-containers.yml
@@ -3,15 +3,30 @@ on:
   workflow_dispatch:
     inputs: 
         environment:
+          description: 'The environment to tag the resulting images to.  The resulting container would be tagged as <application-name>:<application-version>-<environment>.<githubsha>'
           type: choice
           required: true
+          default: 'TEST'
           options:
           - TEST
         values:
+          description: 'The name of the Helm chart values file to update with the resulting tag name.'
           type: choice
           required: true
+          default: 'test2'
           options:
-          - test2          
+          - test2
+        build-elasticsearch:
+          description: 'Release a new ElasticSearch Container?'
+          required: true
+          type: boolean
+          default: false    
+        build-nifi:
+          description: 'Release a new NiFi Container?'
+          required: true
+          type: boolean
+          default: false            
+
 jobs:
   build-nbs-gateway-release-candidate:
     name: Build NBS Gateway Release Candidate
@@ -66,6 +81,7 @@ jobs:
       GIT_USER_NAME: ${{secrets.GIT_USER_NAME}}
       HELM_TOKEN: ${{secrets.HELM_TOKEN}}      
   build-elasticsearch-release-candidate:
+    if: ${{ inputs.build-elasticsearch == 'true' }}
     name: Build Elasticsearch Release Candidate
     uses: CDCgov/NEDSS-Workflows/.github/workflows/Release-gradle-microservice-container.yaml@main
     with:
@@ -92,6 +108,7 @@ jobs:
       GIT_USER_NAME: ${{secrets.GIT_USER_NAME}}
       HELM_TOKEN: ${{secrets.HELM_TOKEN}}        
   build-nifi-release-candidate:
+    if: ${{ inputs.build-nifi == 'true' }}
     name: Build NiFi Release Candidate
     uses: CDCgov/NEDSS-Workflows/.github/workflows/Release-gradle-microservice-container.yaml@main
     with:
--- a/.github/workflows/Release-candidate-containers.yml
+++ b/.github/workflows/Release-candidate-containers.yml
@@ -87,7 +87,7 @@ jobs:
     uses: CDCgov/NEDSS-Workflows/.github/workflows/Update-helm-charts.yaml@main
     with:
       microservice_name: elasticsearch
-      values_file_with_path: charts/elasticsearch/values_${{inputs.environment}}.yaml
+      values_file_with_path: charts/elasticsearch-efs/values_${{inputs.environment}}.yaml
       new_image_tag: ${{ needs.call-build-microservice-container-workflow.outputs.output_image_tag }}
     secrets:
       GIT_USER_EMAIL: ${{secrets.GIT_USER_EMAIL}}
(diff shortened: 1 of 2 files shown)


.github/workflows/Release-candidate-containers.yml
```diff
@@ -0,0 +1,119 @@
+name: Release Candidate Container Build and Deployment Preparation
+on:
+  workflow_dispatch:
+    inputs: 
+        environment:
+          type: choice
+          required: true
+          options:
```
szamfir, line 8: To reduce button click you can add a line after line 7-- default: 'TEST'

And  after line 12 --  default: 'test2'

.github/workflows/Release-candidate-containers.yml
```diff
@@ -0,0 +1,119 @@
...
+    uses: CDCgov/NEDSS-Workflows/.github/workflows/Update-helm-charts.yaml@main
+    with:
+      microservice_name: modernization-api
+      values_file_with_path: charts/modernization-api/values_${{inputs.values}}.yaml
+      new_image_tag: ${{ needs.call-build-microservice-container-workflow.outputs.output_image_tag }}
+    secrets:
+      GIT_USER_EMAIL: ${{secrets.GIT_USER_EMAIL}}
+      GIT_USER_NAME: ${{secrets.GIT_USER_NAME}}
+      HELM_TOKEN: ${{secrets.HELM_TOKEN}}      
+  build-elasticsearch-release-candidate:
```
szamfir, line 68: Is there always a need to create a new Elasticsearch (or even NiFi) release candidate? In many cases we can reuse the same one. 

One way to customize is with boolean inputs for each microservice. In the case of elasticsearch you would have an input like the one below. And add a new line after line 68 and after line 82 that says -- if: ${{ inputs.build-elasticsearch == 'true' }}

build-elasticsearch:
   description: 'Release a new ElasticSearch Container?'
   required: true
   type: boolean
   default: true

szamfir (approved):
1. I realize the scope of this is to reduce the prep work needed for releasing containers to test. Since this is test it is probably ok to leave as is, but will you Always be building a new container or is there a case where you will be "promoting" a container from int?

Overall looks good and approving (needs testing though). My suggestions are general potential improvements and are not required, but may be desired.

adamloup-enquizit:
> 1. I realize the scope of this is to reduce the prep work needed for releasing containers to test. Since this is test it is probably ok to leave as is, but will you Always be building a new container or is there a case where you will be "promoting" a container from int?
> 
> 
> Overall looks good and approving (needs testing though). My suggestions are general potential improvements and are not required, but may be desired.

I had been creating a new container with each release branch because we were not including pagebuilder for the first few release candidates.  It's really not needed any more, i'll think about how to fix this in the future

.github/workflows/Release-candidate-containers.yml
adamloup-enquizit: implemented this, thanks

.github/workflows/Release-candidate-containers.yml
```diff
@@ -0,0 +1,136 @@
...
+      dockerfile_relative_path: -f ./cdc-sandbox/elasticsearch/Dockerfile .
+      environment_classifier: ${{inputs.environment}}
+    secrets:
+      CDC_NBS_SANDBOX_SHARED_SERVICES_ACCOUNTID: ${{secrets.CDC_NBS_SANDBOX_SHARED_SERVICES_ACCOUNTID}}
+      ECR_REPO_BASE_NAME: ${{secrets.ECR_REPO_BASE_NAME}}
+      GIT_USER_EMAIL: ${{secrets.GIT_USER_EMAIL}}
+      GIT_USER_NAME: ${{secrets.GIT_USER_NAME}}
+      HELM_TOKEN: ${{secrets.HELM_TOKEN}}
+  prepare-elasticsearch-release-candidate:
+    name: Prepare Elasticsearch Release Candidate Deployment
```
szamfir, line 99: This needs the same if statement as line 84 or it will fail the pipeline.

.github/workflows/Release-candidate-containers.yml
```diff
@@ -0,0 +1,136 @@
...
+      dockerfile_relative_path: -f ./cdc-sandbox/nifi/Dockerfile .
+      environment_classifier: ${{inputs.environment}}
+    secrets:
+      CDC_NBS_SANDBOX_SHARED_SERVICES_ACCOUNTID: ${{secrets.CDC_NBS_SANDBOX_SHARED_SERVICES_ACCOUNTID}}
+      ECR_REPO_BASE_NAME: ${{secrets.ECR_REPO_BASE_NAME}}
+      GIT_USER_EMAIL: ${{secrets.GIT_USER_EMAIL}}
+      GIT_USER_NAME: ${{secrets.GIT_USER_NAME}}
+      HELM_TOKEN: ${{secrets.HELM_TOKEN}}
+  prepare-nifi-release-candidate:
+    name: Prepare NiFi Release Candidate Deployment
```
szamfir, line 126: This needs the same if statement as line 111 or it will fail the pipeline.

szamfir (changes requested):
2 changes to prevent pipeline failing. See in-line comments

.github/workflows/Release-candidate-containers.yml
szamfir: I stand corrected based on definition of the needs block.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

.github/workflows/Release-candidate-containers.yml
szamfir: I stand corrected based on definition of the needs block.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

szamfir (approved):
LGTM.

PR #763, opened 2023-12-01
Changes that are related to group/ungroup subsection

sizatenquizit:
## Description

In NBS Classic Page Builder, a user is able to mark a subsection as grouped or ungrouped. Modernized APIs needs to be created to perform these actions.

The modernized APIs should leave the database in a state that is exactly the same as if the group or ungroup was handled by NBS Classic. 

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1527

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaUiMetadata.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaUiMetadata.java
@@ -18,6 +18,7 @@ import gov.cdc.nbs.questionbank.page.command.PageContentCommand;
 import gov.cdc.nbs.questionbank.page.command.PageContentCommand.UpdateSection;
 import gov.cdc.nbs.questionbank.page.command.PageContentCommand.UpdateSubsection;
 import gov.cdc.nbs.questionbank.page.command.PageContentCommand.UpdateTab;
+import gov.cdc.nbs.questionbank.page.content.subsection.request.GroupSubSectionRequest;
 import gov.cdc.nbs.questionbank.page.exception.AddQuestionException;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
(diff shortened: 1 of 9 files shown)


achapm (approved):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaUiMetadata.java
```diff
@@ -524,6 +525,33 @@ private void setVisible(boolean visible) {
     this.displayInd = visible ? "T" : "F";
   }
 
+  public void update(PageContentCommand.GroupSubsection command) {
+    setBlockNm(command.blockName());
```
adamloup-enquizit, line 529: nit: if you're within the object itself it's better to prefer  assigning the  value to the field directly.  

Just a bit of context: When using this pattern with the commands the entities should not really have setters as that allows the entity state to be changed from anywhere.  The purpose of the `command` is to encapsulate all values needed to carryout a change forcing the business logic to be contained within the entity and not spread around the application.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/subsection/SubSectionGrouper.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        if (totalPercentage != 100) {
+            throw new UpdateSubSectionException("the total of batchTableColumnWidth must calculate to 100");
+        }
+        WaTemplate page = entityManager.find(WaTemplate.class, pageId);
+        if (page == null) {
+            throw new UpdateSubSectionException("Unable to find page with id: " + pageId);
+        }
+        WaUiMetadata section = page.groupSubSection(asCommand(userId, request));
+        entityManager.flush();
```
adamloup-enquizit, line 46: issue:  we shouldn't have to explicitly flush the entitymanager.  Interactions with an entity should be `@Transactional`.  This can be accomplished with the [PageService](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/PageService.java), for examples of it's usage see [PageInformationChanger](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/information/change/PageInformationChanger.java)

apps/question-bank/src/test/java/gov/cdc/nbs/questionbank/page/content/subsection/SubSectionControllerTest.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class SubSectionControllerTest {
```
adamloup-enquizit, line 23: nit: Feature tests using cucumber are more effective at testing Contollers over mocks.

apps/question-bank/src/test/java/gov/cdc/nbs/questionbank/page/content/subsection/SubsectionGrouperTest.java
```diff
@@ -0,0 +1,250 @@
...
+        ResponseEntity<String> result = grouper.unGroup(1l, request, userId);
+        assertEquals("Subsection " + waUiMetadata.getId() + " is UnGrouped Successfully ", result.getBody());
+    }
+
+    @Test
+    void testGroupSubSection() {
+        // Create a mock for WaUiMetadata
+        Long subSectionId = 1L;
+        WaUiMetadata subsection = mock(WaUiMetadata.class);
+        when(subsection.getId()).thenReturn(subSectionId);
```
adamloup-enquizit, line 117: nit:  Not sure if there is a need for all of the mocking, these objects aren't very large

PR #768, opened 2023-12-05
Moves page library

adamloup-enquizit:
## Description

Moves the components of Page Library to the `page/library` folder.

## Steps to verify

1. Ensure the tests pass and the application can be built using `npm run build`


mpeels (approved):
⧢

PR #769, opened 2023-12-05
Fixes that are related to group/ungroup subsection

sizatenquizit:
## Description

In NBS Classic Page Builder, a user is able to mark a subsection as grouped or ungrouped. Modernized APIs needs to be created to perform these actions.

The modernized APIs should leave the database in a state that is exactly the same as if the group or ungroup was handled by NBS Classic. 

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1527


achapm (approved):
👍

apps/question-bank/src/test/resources/features/page/content/subsection/GroupSubsection.feature
```diff
@@ -0,0 +1,23 @@
+@group_subsection
+Feature: Group Subsection
+
+    Background:
+        Given I have a page
+        And the page has a tab
+        And the page has a section in the 1st tab
+        And the page has a sub-section named "whatever" in the 1st section
+
```
mpeels, line 9: Issue: Test does not add any questions to the subsection and therefore does not verify that the question is updated to reflect that it has been grouped / ungrouped.

The method for grouping also filters out non questions which i assume is the same behavior as Classic NBS,
```java
.filter(ui -> batchIds.contains(ui.getId()) && ui.getNbsUiComponentUid() == QUESTION)
```
It would be good to add a static element or other entry to the subsection and verify that it is not modified as part of the grouping / ungrouping.

mpeels (changes requested):

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaTemplate.java
```diff
@@ -46,7 +48,7 @@ public class WaTemplate {
   private static final long TAB = 1010l;
   private static final long SECTION = 1015l;
   private static final long SUB_SECTION = 1016l;
-  private static final long QUESTION = 1007l;
+  private static final long QUESTION = 1008l;
```
mpeels, line 51: Potential Issue (needs checked): I think this might be wrong. Several of the page components have an id that wont change (such as TAB, SECTION, SUB_SECTION). A question could be one of several component 
|id | component type|
|---|----|
|1006 | radio button|
|1007 | Single select drop down|
|1008 | user entered text|
|1009 | multiline text|

This list is only partial. It might be a good idea to find the group / ungroup logic within Classic and try to verify how its working.  It might be the case that the check for the component type of question is either not needed, or can be inverted to check that the component is not one of a set of values.

mpeels (changes requested):

mpeels (approved):

alaapbharadwaj (approved):

PR #801, opened 2023-12-14
Implement feature to remove a question from the page

sizatenquizit:
## Description

An API needs to be created that will allow removing a question from a Draft page when the question is not standard.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1523

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaTemplate.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaTemplate.java
@@ -41,697 +41,713 @@ import java.util.concurrent.atomic.AtomicInteger;
 @Entity
 @Table(name = "WA_template", catalog = "NBS_ODSE")
 public class WaTemplate {
-  private static final String DRAFT = "Draft";
-  private static final long TAB = 1010l;
-  private static final long SECTION = 1015l;
-  private static final long SUB_SECTION = 1016l;
-  private static final long QUESTION = 1007l;
+    private static final String DRAFT = "Draft";
+    private static final long TAB = 1010l;
+    private static final long SECTION = 1015l;
+    private static final long SUB_SECTION = 1016l;
+    private static final long QUESTION = 1007l;
 
-  @Id
-  @GeneratedValue(strategy = GenerationType.IDENTITY)
-  @Column(name = "wa_template_uid", nullable = false)
-  private Long id;
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "wa_template_uid", nullable = false)
+    private Long id;
 
-  @Column(name = "template_type", nullable = false, length = 50)
-  private String templateType;
+    @Column(name = "template_type", nullable = false, length = 50)
+    private String templateType;
 
-  @Lob
-  @Basic(fetch = FetchType.LAZY)
-  @Column(name = "xml_payload", columnDefinition = "TEXT")
-  private String xmlPayload;
+    @Lob
+    @Basic(fetch = FetchType.LAZY)
+    @Column(name = "xml_payload", columnDefinition = "TEXT")
+    private String xmlPayload;
 
-  @Column(name = "publish_version_nbr")
-  private Integer publishVersionNbr;
+    @Column(name = "publish_version_nbr")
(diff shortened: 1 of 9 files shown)


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/question/WaQuestion.java
```diff
@@ -24,6 +26,7 @@
 
 @NoArgsConstructor
 @Getter
+@Setter
```
mpeels, line 29: Issue: We want to avoid `Setter`s on the entities if at all possible. Any change should be handled through a command. This way all business rules can be handled when processing those commands.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaTemplate.java
```diff
@@ -41,697 +41,713 @@
...
+        //can not delete standard questions
+        if (command.question().isStandard()) {
+            throw new PageContentModificationException("Unable to delete standard question");
+        }
+        // ensure page already contain question
+        WaUiMetadata waUiMetadata = uiMetadata.stream()
+                .filter(e -> e.getQuestionIdentifier() != null
+                        && e.getQuestionIdentifier().equals(command.question().getQuestionIdentifier())).findFirst().orElseThrow(() ->
+                        new PageContentModificationException("Unable to delete a question from a page, the page does not contain the question"));
+        this.uiMetadata.remove(waUiMetadata);
```
mpeels, line 527: Issue (blocking): Each question has an `orderNbr` field. These `MUST` stay in sequential order for the page to correctly render in classic.

There is a helper function in `WaTemplate` to help re-adjust these `orderNbr` after something is added or removed. 
`adjustingComponentsFrom`.

Example:
|question Id | orderNbr|
|----|---- |
|A | 1|
|B | 2 |
|C|3|
|D|4|

if we delete question `C`, then the table will be
|question Id | orderNbr|
|----|---- |
|A | 1|
|B | 2 |
|D|4|


The `orderNbr`s are no longer sequential. but calling `adjustingComponentsFrom(waUiMetadata.getOrderNbr());` would update the questions to be:

|question Id | orderNbr|
|----|---- |
|A | 1|
|B | 2 |
|D|3|

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaTemplate.java
```diff
@@ -41,697 +41,713 @@
...
-  }, orphanRemoval = true)
-  private List<WaRuleMetadatum> waRuleMetadatums;
+    public void deleteQuestion(PageContentCommand.DeleteQuestion command) {
+        // Can only modify Draft pages
+        verifyDraftType();
+
+        //can not delete standard questions
+        if (command.question().isStandard()) {
+            throw new PageContentModificationException("Unable to delete standard question");
+        }
```
mpeels, line 521: Issue: I would not trust the command to have the correct `isStandard`. Instead I would perform this check on the `WaUiMetadata` entity itself

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/question/PageQuestionDeleter.java
```diff
@@ -0,0 +1,38 @@
...
+    public void deleteQuestion(Long pageId,Long questionId, Long user) {
+
+        WaTemplate page = entityManager.find(WaTemplate.class, pageId);
+        if (page == null) {
+            throw new DeleteQuestionException("Failed to find page with id: " + pageId);
+        }
+        WaQuestion question = entityManager.find(WaQuestion.class,questionId);
+        if (question == null) {
+            throw new QuestionNotFoundException(questionId);
+        }
```
mpeels, line 33: Suggestion: This check probably isn't necessary since the same check is done within `WaTemplate`

mpeels (changes requested):
Suggestion: There is a code formatting file `eclipse-java-google-style.xml` that if imported, should greatly reduce the changes in your PRs

Formatting [README](https://github.com/CDCgov/NEDSS-Modernization/blob/main/documentation/Code-Formatting.md)

mpeels (approved):

PR #870, opened 2024-01-05
adding focus to result lists count label

hclarkEnq:
## Description

A user using screen reader should be able to see the focus set on the resultsCount which is the results count label that displays the total results found using the search criteria with a button showing what was searched upon any search made in the patient search. 

If there are results found or not but a search. has been made the focus of the main content should direct towards the results count label.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT1-1844

Steps To Verify 

1. Navigate to Advanced search
2. Make a search from any field in the patient search
3. The focus has been set to the results count but not visible for regular users and only for users with screen readers therefore the focus of the element itself is hidden. 


mdavis-enquizit (dismissed):

apps/modernization-ui/src/pages/advancedSearch/AdvancedSearch.scss
```diff
@@ -194,3 +194,7 @@ button {
     text-transform: uppercase;
     color: colors.$base;
 }
+
+.results-for:focus {
```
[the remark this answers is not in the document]
hclarkEnq: The issue with this is that the mouse click would not activate the style and regular users would see the results count outlined potentially causing a greater UX issue

mdavis-enquizit (dismissed):

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -22,16 +24,37 @@ interface SkipLinkProviderProps {
...
+        skipTo,
+        focusedTo,
+        currentFocusTarget
     };
+
+    useEffect(() => {
+        const targetElement = document.getElementById(currentFocusTarget);
+        if (targetElement && isActionFocused) {
+            targetElement.focus();
+            setCurrentFocusTarget('');
```
adamloup-enquizit, line 50: question:  Should the focus target be cleared at this point?  Does it need to remain as the main content?

mdavis-enquizit:
apps/modernization-ui/src/apps/search/event/components/labReportSearch does not like switching branches since it can't tell the difference in labReportSearch vs  LabReportSearch
Screenshot 2024-01-09 at 3 43 13 PM

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
hclarkEnq: yes we need this because if we remove it then after the search is made if the user makes another search then the focus would stay at the last Name and not go the results count because the skipTo is already assigned to the results count so we need to clear it after the focus is made

mdavis-enquizit (dismissed):

adamloup-enquizit (approved):
👍 Looks good

mdavis-enquizit (approved):

PR #890, opened 2024-01-11
create endpoint for fetching race values

sizatenquizit:
## Description
create endpoint for fetching race values from the table Race_code with the filter on the code_set_nm

## Tickets
https://cdc-nbs.atlassian.net/browse/CNFT2-1687

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetController.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetController.java
@@ -85,6 +85,11 @@ public class ValueSetController {
     return valueSetReader.findConceptCodes(codeSetNm);
   }
 
+  @GetMapping("{codeSetNm}/race/concepts")
+  public List<RaceConcept> findRaceConceptsByCodeSetName(@PathVariable String codeSetNm) {
+    return valueSetReader.findRaceConceptCodes(codeSetNm);
+  }
+
   @PutMapping("/{valueSetCode}/concepts/{localCode}")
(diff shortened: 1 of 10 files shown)


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/valueset/ValueSetController.java
```diff
@@ -85,6 +85,11 @@ public List<Concept> findConceptsByCodeSetName(@PathVariable String codeSetNm) {
     return valueSetReader.findConceptCodes(codeSetNm);
   }
 
+  @GetMapping("{codeSetNm}/race/concepts")
+  public List<RaceConcept> findRaceConceptsByCodeSetName(@PathVariable String codeSetNm) {
+    return valueSetReader.findRaceConceptCodes(codeSetNm);
+  }
```
mpeels, line 91: Issue: To support the edit / preview pages, special handling needs to be added to the `ValueSetReader.findConceptCodes` method. If the `codeSetNm` provided is equal to `P_RACE_CAT` then it should return the values from the `race_code` table where code_set_nm = `P_RACE_CAT`.


Why: 
When the `edit` page loads the questions, the coded questions are just associated with a value set. The UI uses the same API to fetch these values regardless of what the codeset is. Further clarification can be provided by Alaap as he is the one that initially discovered the issue.

mpeels (changes requested):

mpeels (dismissed):

mpeels (approved):

alaapbharadwaj (approved):

PR #918, opened 2024-01-22
Cnft1 1920 alt x subsequent searches

hclarkEnq:
## Description

Once a search is made in the advanced search  page for patient searches the "Alt-X" hotkey is available for any user in order to return to the last name field. 
This is done with the intention of having subsequent searches in case the existing search results are not needed. 

## Tickets

* [CNFT1-1913)]([CNFT1-1913](https://cdc-nbs.atlassian.net/browse/CNFT1-1913))

[CNFT1-1913]: https://cdc-nbs.atlassian.net/browse/CNFT1-1913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jlee-enquizit (approved):

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -24,14 +26,30 @@ interface SkipLinkProviderProps {
 
 export const SkipLinkProvider = ({ children }: SkipLinkProviderProps) => {
     const [currentFocusTarget, setCurrentFocusTarget] = useState('');
+    const [previousFocusTarget, setPreviousFocusTarget] = useState('');
```
adamloup-enquizit, line 29: question:  How is `currentFocusTarget` different from `previousFocusTarget`?  It seems to be the same thing here

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -24,14 +26,30 @@ interface SkipLinkProviderProps {
...
     const skipTo = (id: string) => {
         setCurrentFocusTarget(id);
     };
 
+    const previousTarget = (id: string) => {
+        setPreviousFocusTarget(id);
+    };
+
+    useEffect(() => {
+        const handleKeyDown = (event: KeyboardEvent) => {
```
adamloup-enquizit, line 40: question: is `ALT-X` a global shortcut or is it just for searches?

mdavis-enquizit (approved):

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
```diff
@@ -24,14 +26,30 @@ interface SkipLinkProviderProps {
...
+            if (event.altKey && event.keyCode === 88) {
+                event.preventDefault();
+                focusedTarget(previousFocusTarget);
+            }
+        };
+        window.addEventListener('keydown', handleKeyDown);
+    }, [previousFocusTarget]);
```
adamloup-enquizit, line 47: issue (blocker):  This is going to add a _new_ event listener each time the `previousFocusTarget` is changed.  A clean up function should be returned whenever a hook is interacting with an external system, see [useEffect](https://react.dev/reference/react/useEffect)

adamloup-enquizit (changes requested):
It seems like `alt-x` might be a global shortcut for returning to the skip link.  If that is the case is a separate state required?  Should the target of the shortcut change based on the `currentFocusTarget` alone?

apps/modernization-ui/src/SkipLink/SkipLinkContext.tsx
hclarkEnq: just for searches for now. Event and Patient search,
It should only work on that page or another page that has the same buttons and function if there is another that works like the search page which in this case can be event search

adamloup-enquizit: Since this is just for the search pages shouldn't the functionality be on the search pages and not the skip link?  

The `alt-x` shortcut is specific to search features to allow a quick way to return to the criteria and is a distinct feature from the skip link.  In this case the target of the skip link is the same as the target for `alt-x` however, it doesn't mean the shortcut is part of the skip link feature.

adamloup-enquizit (approved):

mdavis-enquizit (dismissed):

mdavis-enquizit (approved):

adamloup-enquizit (approved):

PR #921, opened 2024-01-22
Standardize object structure for Create and View APIs for business rules

sizatenquizit:
## Description

The create and view APIs for business rules handle the sourceValue differently. To make the edit and create of a business rule easier, the objects should be updated to have the same structure.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1768


mpeels (changes requested):
I believe the `sourceValue` fields of `ViewRuleResponse` and `CreateRuleRequest` need to be changed to match (probably completely re-done to make more sense).

The `CreateRuleRequest.java` has a `sourceValue` field as defined:
image


The `ViewRuleResponse.java` has a `sourcevalue` field of:
image


I haven't looked deeply into how this field works but i think both of these should be updated to be:
`List<SourceValue> sourceValues;`

Where `SourceValue` would be:
```java
public record SourceValue(String id, String text) {
}
```


It might be a good idea to confirm with Richard that this would meet his needs.

mpeels:
Also probably need to add the `anySourceValue` to `ViewRuleResponse`

sizatenquizit:
A decision was made not to change anything in the API.

PR #939, opened 2024-01-25
replacing tags with proper headings for addPatient

hclarkEnq:
## Description

Added correct heading and size for the add patient page.

## Tickets

* [1979](https://cdc-nbs.atlassian.net/browse/CNFT1-1979)


apps/modernization-ui/src/styles/global.scss
```diff
@@ -64,3 +63,11 @@ a {
         display: none;
     }
 }
+
+.h1 {
```
adamloup-enquizit, line 67: question: Since these are global styles shouldn't they be applied at the element level and not through a class?  If there is a reason to use the class then these are already defined [here](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/modernization-ui/src/components/heading/heading.module.scss).

apps/modernization-ui/src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -32,7 +32,7 @@ const SideNavigation = ({ title, className, children = [] }: SideNavigationProps
 
     return (
         <div className={classNames('navigation', className)}>
-            <h2>{title}</h2>
+            <h2 className="margin-0">{title}</h2>
```
adamloup-enquizit, line 35: issue: The [navigation class](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/modernization-ui/src/components/SideNavigation/SideNavigation.scss) is already setting styles for an `h2`, that would be the place to apply margins to this specific header.

adamloup-enquizit (approved):

mdavis-enquizit (approved):

PR #954, opened 2024-01-30
Implement error display functionality for duplicate Unique ID

sizatenquizit:
## Description

When creating a question with the same Unique ID previously used, the Create and add to page button (when clicked) is inoperable. The application should return an error message similar to "Unique ID cannot be duplicated.”

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1815


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/question/request/QuestionValidationRequest.java
```diff
@@ -0,0 +1,5 @@
+package gov.cdc.nbs.questionbank.question.request;
+
+public record QuestionValidationRequest(String field, String value) {
```
mpeels, line 3: Thoughts on having an `enum` for the field? This will let the UI know exactly what values are valid here.

mpeels (approved):

PR #1002, opened 2024-02-09
Refactor of PageRuleFinderService

sizatenquizit:
## Description

The PageRuleFinderService is an interface with a PageRuleFinderServiceImpl currently leveraging the WaRuleMetaDataRepository to perform rule retrieval. This is not consistent with how our other APIs are implemented.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-1997


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/pagerules/PageRuleFinder.java
```diff
@@ -0,0 +1,140 @@
...
+       from WA_rule_metadata [rule]
+       where    UPPER(source_values) LIKE CONCAT('%', UPPER(:searchValue), '%') OR 
+                UPPER(target_question_identifier) LIKE CONCAT('%', UPPER(:searchValue), '%')
+       order by add_time
+       offset :offset rows
+       fetch next :pageSize rows only
+       """;
+
+  private final NamedParameterJdbcTemplate template;
+  private final RowMapper<RuleResponse> mapper;
```
mpeels, line 87: Question: Can we not map directly to a `Rule` instead of `RuleResponse` and then mapping it again to a `Rule` in the `PageRuleReader`?

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/pagerules/PageRuleFinder.java
```diff
@@ -0,0 +1,140 @@
...
+                     UPPER(target_question_identifier) LIKE CONCAT('%', UPPER(:searchValue), '%')
+           ) as [TotalCount] 
+                                                
+       from WA_rule_metadata [rule]
+       where    UPPER(source_values) LIKE CONCAT('%', UPPER(:searchValue), '%') OR 
+                UPPER(target_question_identifier) LIKE CONCAT('%', UPPER(:searchValue), '%')
+       order by add_time
+       offset :offset rows
+       fetch next :pageSize rows only
+       """;
```
mpeels, line 84: Issue: The search should also return only rules associated with a given page id. There is no plan to have a "global" rule search.

apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/pagerules/Rule.java
```diff
@@ -0,0 +1,20 @@
...
+    long id,
+    long template,
+    String function,
+    String description,
+    QuestionInfo sourceQuestion,
+    boolean anySourceValue,
+    List<String> sourceValues,
+    String comparator,
+    String targetType,
+    List<QuestionInfo> targets
```
mpeels, line 15: Does not match signature in story:
```java
public record Rule(
        @ApiModelProperty(required = true) long id,
        @ApiModelProperty(required = true) long template,
        @ApiModelProperty(required = true) Function function,
        String description,
        @ApiModelProperty(required = true) SourceQuestion sourceQuestion,
        @ApiModelProperty(required = true) boolean anySourceValue,
        List<SourceValue> sourceValues,
        @ApiModelProperty(required = true) Comparator comparator,
        @ApiModelProperty(required = true) TargetType targetType,
        @ApiModelProperty(required = true) List<Target> targets
){}
```

mpeels (approved):

rkrusselenq (approved):

PR #1013, opened 2024-02-11
Solve grouped flag issue

sizatenquizit:
## Description

After grouping questions in a Subsection, the API returns a success response. However, Subsection value for isGrouped does not change to true

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-2015


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/page/content/subsection/SubSectionGrouper.java
```diff
@@ -46,6 +47,7 @@ public void group(Long pageId, GroupSubSectionRequest request, Long userId) {
       throw new PageNotFoundException(pageId);
     }
     page.groupSubSection(asCommand(userId, request), questionManagementUtil.getQuestionNbsUiComponentUids());
+    entityManager.flush();
```
mpeels, line 50: Question: Is the `.flush()` necessary? I thought that if the class was `@Transactional` when the method completed any outstanding changes by the `entityManager` would then be committed.

mpeels (approved):

lramakr (approved):

PR #1031, opened 2024-02-14
Creating Wa_RDB_metadata table entry while adding question to page.

sizatenquizit:
## Description

When adding a question to a page using the modernized APIs, it appears no Wa_RDB_metadata table entry is being created.

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT2-2018

### DIFF
--- a/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaRdbMetadatum.java
+++ b/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/entity/WaRdbMetadatum.java
@@ -1,5 +1,7 @@
 package gov.cdc.nbs.questionbank.entity;
 
+import gov.cdc.nbs.questionbank.entity.question.WaQuestion;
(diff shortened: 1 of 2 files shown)


mpeels (approved):

alaapbharadwaj (commented):
Question: Are Wa_RDB_Metadata entries only created for questions when the user chooses to enable messaging? In classic only some of the questions have RDB entries and some questions have no entries, this could be because messaging has not been selected when creating the question. Might require more investigation into how classic is doing it.

sizatenquizit:
The reply to Alaap's question: "Are Wa_RDB_Metadata entries only created for questions when the user chooses to enable messaging?".

No. When adding a question to a page, if the question has data for RDB then a new record is added to RDB table. This story is missing this part! It adds a row in RDB table even if the question doesn't not have data for RDB. I will modify it to add a row in RDB table for only a question that has data for RDB.

mpeels (approved):

rkrusselenq (approved):

PR #1133, opened 2024-03-04
CNFT1-2082 Stored Procedure for Importation of data into CliniDeiD - ODSE

sveselev:


apps/modernization-api/src/main/resources/db/goldendb/01-pii-export-odse.sql
mpeels, line 1: I don't believe this should live in the `modernization-api`. 

Perhaps creating a dedicated folder would be better. `/db/golden/...`? Open to suggestions here

sveselev: Sure, no problem, that was a quick decision, it's easy to move it under new db sub-project

adamloup-enquizit (approved):
👍  much better location

adamloup-enquizit (approved):

mpeels (approved):

PR #1263, opened 2024-03-31
Target Question has to have Static elements also included

sizatenquizit:
## Description

Business Rules: Target Question has to have Static elements also included. When a business rule is added, it has to appear in the Business Rules Library screen with Static elements listed with the Unique ID

## Tickets
https://cdc-nbs.atlassian.net/browse/CNFT2-2254


alaapbharadwaj:
Comment(blocking): currently question identifiers are not being populated when static elements are created. So when we grab static elements for the target question we are not able to filter or display correctly because the question identifier for static elements are null. Other than that everything looks great

mpeels (approved):

PR #1314, opened 2024-04-23
Create Icon Button component

rkrusselenq:
## Description

Creates a component pre-styled for icon only buttons.
## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/jira/software/c/projects/CNFT1/boards/112?selectedIssue=CNFT1-2195)


rkrusselenq:
Questioning weather this component even needs to exist, the only change from a regular button is adding "size='big'" to the <Button>

adamloup-enquizit:
> Questioning weather this component even needs to exist, the only change from a regular button is adding "size='big'" to the

I had the same thought.  There are two kinds of buttons with icons, Button with icon and text and Button with only icon.  The sizing of the icon is different between the two.  Is there an easy way to distinguish these two in styles alone?

rkrusselenq:
> > Questioning weather this component even needs to exist, the only change from a regular button is adding "size='big'" to the
> 
> I had the same thought. There are two kinds of buttons with icons, Button with icon and text and Button with only icon. The sizing of the icon is different between the two. Is there an easy way to distinguish these two in styles alone?

I'm no css wizard, but i've been looking into it. Currently we select all buttons with svg's using the :has pseudo selector.  I thought i was close selecting buttons with svg's but no text using something like ' &:not(:has(*:not(svg)))' but i can't seem to figure out exactly what to use.  My other thought is to just use a class for icon buttons - we do the same for destructive buttons.

hclarkEnq:
> > Questioning weather this component even needs to exist, the only change from a regular button is adding "size='big'" to the
> 
> I had the same thought. There are two kinds of buttons with icons, Button with icon and text and Button with only icon. The sizing of the icon is different between the two. Is there an easy way to distinguish these two in styles alone?

I was thinking maybe we can create a button component that takes both the text and the icon as props and we set the styles for both, this way we can control its usage a little more with text with icon and only icon scenarios, also only text. Thoughts?

rkrusselenq:
> > > Questioning weather this component even needs to exist, the only change from a regular button is adding "size='big'" to the
> > 
> > 
> > I had the same thought. There are two kinds of buttons with icons, Button with icon and text and Button with only icon. The sizing of the icon is different between the two. Is there an easy way to distinguish these two in styles alone?
> 
> I was thinking maybe we can create a button component that takes both the text and the icon as props and we set the styles for both, this way we can control its usage a little more with text with icon and only icon scenarios, also only text. Thoughts?

Ok I've now updated so that we can have a button that takes a label, or an icon, or both.  I replaced 2 buttons in the codebase to test it out and seems to work fine.

Only issue was I could not name it just "Button" as it was causing conflicts with the props from importing "Button" from trussworks.  Went with NBSButton but i feel like we can do better maybe?

adamloup-enquizit:
> > > > Questioning weather this component even needs to exist, the only change from a regular button is adding "size='big'" to the
> > > 
> > > 
> > > I had the same thought. There are two kinds of buttons with icons, Button with icon and text and Button with only icon. The sizing of the icon is different between the two. Is there an easy way to distinguish these two in styles alone?
> > 
> > 
> > I was thinking maybe we can create a button component that takes both the text and the icon as props and we set the styles for both, this way we can control its usage a little more with text with icon and only icon scenarios, also only text. Thoughts?
> 
> Ok I've now updated so that we can have a button that takes a label, or an icon, or both. I replaced 2 buttons in the codebase to test it out and seems to work fine.
> 
> Only issue was I could not name it just "Button" as it was causing conflicts with the props from importing "Button" from trussworks. Went with NBSButton but i feel like we can do better maybe?

One reason to name it `Button` is the hope that it can be a direct replacement for the existing component that would allow just an import change to use the new component throughout the application.  

There are a couple of ways I can think to allow us to use the name `Button`.  

1. The import can be aliased within the component. `import { Button as TrussworksButton } from '@trussworks/react-uswds';`
2. We drop the Trussworks `Button` component completely. There is little extra functionality, it's mainly a wrapper of the html `button` element with specific USWDS styles.

apps/modernization-ui/src/components/button/NBSButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+}: Props) => {
+    const isIconOnly = icon && !label;
+    const classesArray = [];
+    if (destructive) {
+        classesArray.push(styles.destructive);
+    }
+
+    if (isIconOnly) {
+        classesArray.push(styles.icon);
+    }
```
adamloup-enquizit, line 35: suggestion (non-blocking):  The [classNames](https://github.com/JedWatson/classnames#readme) function can be used to handle conditionally applying classes to a component.  An object can be passed to the function with class names that are applied if the property value is `true`.

When using SASS modules bracket notation has to be used.  

```typescript

classNames({
  [styles.destructive] : styles.destructive,
  [styles.icon] : styles.icon
});

```

hclarkEnq (dismissed):

hclarkEnq (approved):

PR #1361, opened 2024-05-07
Remove deprecated constructor usage

mpeels:
## Description

Initialize `Locale` objects using static initializers instead of deprecated constructor.


adamloup-enquizit (approved):
👍 Great job making sure the quality stays up

PR #1432, opened 2024-05-28
Cypress e2e action for INT1 

mpeels:
## Description

Creates a GitHub Action to execute Cypress tests against the specified environment.

The tests will run nightly at 3am EST against the int1 environment. There is also a `workflow_dispatch` added to allow manual execution of the workflow. 

## Tickets

* [CNT-28](https://cdc-nbs.atlassian.net/browse/CNT-28)

[Workflow run](https://github.com/CDCgov/NEDSS-Modernization/actions/runs/9276096208)

[CNT-28]: https://cdc-nbs.atlassian.net/browse/CNT-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/Regression-tests.yaml
+++ b/.github/workflows/Regression-tests.yaml
@@ -1,0 +1,49 @@
+name: Regression tests
+on:
+  schedule:
+    # weekdays at 7am UTC -> 3am EST
+    - cron: "0 7 * * 1-5"
+  workflow_dispatch:
+    inputs:
+      spec:
+        description: "Spec to run"
+        required: false
+        type: string
+        default: "cypress/**/*.feature"
+      environment:
+        description: "Target environment"
+        required: true
+        type: choice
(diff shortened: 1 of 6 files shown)


.github/workflows/Cypress-Int1.yaml
szamfir, line 1: This looks good as a starting point and looks good overall. We would likely want to pull certain things like CYPRESS_BASE_URL into a variable that is passed in at runtime to reuse the same workflow against any environment.

szamfir (dismissed):

szamfir (approved):
Looks good. Keeping the scope limited to 2 environments for the time being makes sense. We can update this if the environments ever change.

PR #1520, opened 2024-07-03
Move secrets to env vars

mpeels:
## Description

There are some default values provided for local docker database connections, JWT encryption, and search parameter encryption. This removes those defaults and requires the developer to provide them as environment variables at build time or create non tracked `application-local.yml` files where applicable.


cdc-sandbox/db/Dockerfile
```diff
@@ -12,9 +12,10 @@ USER mssql
 ADD ./restore /var/opt/db-restore/
 
 # docker-compose environment variables are only visible at runtime
+ARG DB_PW
```
szamfir, line 15: This argument is something that must be passed in whenever the container is built, correct? Has this been tested with the GitHub pipeline? We may need to modify how containers are built.

apps/question-bank/src/main/resources/application.yml
```diff
@@ -6,7 +6,7 @@ nbs:
   datasource:
     server: localhost
   security:
-    tokenSecret: qJ5+oDurrJnxfaAz6h4Eg1saP7NSeHuQQ/Yq4vh+mbi4B4QYThc/M+33VtPJ5O2A
+    tokenSecret: ${TOKEN_SECRET}
```
szamfir, line 9: Will this require new secrets for helm charts? Assuming yes, if they do not already exist.

szamfir (commented):
Leaving a few comments. There are 2 main concerns:
1. The automated pipeline will need to be adjust to allow for container build time arguments
2. Helm charts will need to be adjusted and test to account for new secrets (as needed)

cdc-sandbox/db/Dockerfile
mpeels: ah, yes. Need to update the `Build-modernization-test-db.yaml`

apps/question-bank/src/main/resources/application.yml
mpeels: It appears we are not setting the `tokenSecret` or `parameterSecret` via the helm chart. We will need to do that.

mpeels:
Temp closing while i test some container builds

PR #1583, opened 2024-07-19
Add jdk.crypto.cryptoki module to ensure cipher availability"

mpeels:
## Description

Resolves (hopefully) the issue of the missing ciphers in the docker container when executed locally and pointed at the foundation team sandbox 2 ALB.

PR #1597, opened 2024-07-24
Adds the cipher dependency to JRE

mpeels:
## Description

Add additional SSL Cipher support to modernization-api and question-bank docker containers in the same manner as it was added to the nbs-gateway here:
https://github.com/CDCgov/NEDSS-Modernization/pull/1583


adamloup-enquizit (approved):
㊙️

alaapbharadwaj (approved):

PR #1663, opened 2024-08-07
Prevents sql int overflow when creating program-jurisdiciton oid

mpeels:
## Description

Resolves an issue within Inductive Health RI environment.


adamloup-enquizit (approved):
👍

stevegsa (approved):

PR #1766, opened 2024-09-04
[FIX] Patient sorting

adamloup-enquizit:
## Description

The recent changes to sorting broke a few of the sorting options on Patient Search. 

- Add support for sorting by `patientid` and `legalname` in Patient Search to match Investigation and Laboratory Search.
- Makes the `local` profile active for tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# local environment set up
 application-local.yml
-
-# merged .graphqls files used in code generation
-ui/src/generated/schema.graphqls
+local.env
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -28,7 +28,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
+# local environment set up
 application-local.yml
-
-# merged .graphqls files used in code generation
-ui/src/generated/schema.graphqls
\ No newline at end of file
+local.env
```
mpeels, line 33: Question: do we want to add any info about this to a README?

apps/modernization-api/src/test/java/gov/cdc/nbs/RunCucumber.java
```diff
@@ -30,7 +30,7 @@
 @CucumberContextConfiguration
 @SpringBootTest
 @Import(PatientLocalIdentifierGeneratorTestConfiguration.class)
-@ActiveProfiles({"default", "test"})
+@ActiveProfiles({"default", "test", "local"})
```
mpeels, line 33: 🙌

mpeels (approved):

hclarkEnq (approved):

.gitignore
adamloup-enquizit: YES!  I was going to get that written up soon

PR #1775, opened 2024-09-06
Create two env files

mpeels:
## Description

1. Adds a `check_env.sh` script that looks for `.env` files within `cdc-sandbox/` and at the root level. If no file is found, it creates them with current set values.
1. Consolidates the cdc-sandbox/.gitignore and the base .gitignore
1. Adds .env to .gitignore
1. docker compose will read a `.env` file found in the same directory as the docker-compose.yml by default.  So the build scripts create 2 `.env` files. One at the root level for the modernized services and one within `cdc-sandbox/`


adamloup-enquizit (approved):
👍 🍨 📁 📁

mdavis-enquizit (approved):

PR #1778, opened 2024-09-06
Delete .env

mpeels:
Since a .env was previously added, it needs to be removed before git will stop tracking changes.


mdavis-enquizit (approved):

adamloup-enquizit (approved):
👍

PR #1871, opened 2024-09-25
Fix panel appearing behind results

mpeels:
## Description

The panel that appears when clicking `Add new` after performing a patient search was appearing behind the search results display. This was probably due to some global styles being removed.

The designs also were not easily inspected and no example exists within the design system. A best effort was made to match the image

### Before
image

### After
image


adamloup-enquizit (approved):
👍 🧑‍🚒

hclarkEnq (approved):

PR #1894, opened 2024-10-02
Cnft1 3196 census tract validation

hclarkEnq:
See attached images observing the button enabled after correct format added
Screenshot 2024-10-02 at 7 57 40 AM
Screenshot 2024-10-02 at 7 57 54 AM
## Description
Census Tract should be in numeric XXXX or XXXX.xx format where XXXX is the basic tract and xx is the suffix. XXXX ranges from 0001 to 9999. The suffix is limited to a range between .01 and .98.

Must not allow nonNumeric input

examples of button disabled for invalid format
Screenshot 2024-10-02 at 8 00 43 AM

## Tickets

* [CNFT1-3196](https://cdc-nbs.atlassian.net/browse/CNFT1-3196)

[CNFT1-3196]: https://cdc-nbs.atlassian.net/browse/CNFT1-3196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/apps/patient/data/address/AddressEntryFields.spec.tsx
+++ b/apps/modernization-ui/src/apps/patient/data/address/AddressEntryFields.spec.tsx
@@ -134,46 +134,41 @@ describe('AddressEntryFields', () => {
         });
     });
 
-    it('should invalidate Census Tract format', async () => {
+    test.each([
+        { value: '0000.00', valid: false },
+        { value: '0001.00', valid: false },
+        { value: '0001.01', valid: true },
+        { value: '1000.00', valid: false },
+        { value: '9999.99', valid: false },
+        { value: '9999.98', valid: true },
+        { value: '0001', valid: true },
+        { value: '9999', valid: true },
+        { value: '0000', valid: false },
+        { value: '9999.00', valid: false },
+        { value: '0001.99', valid: false },
+        { value: '1234.56', valid: true }
+    ])('should validate Census Tract format for value: $value', async ({ value, valid }) => {
         const { getByLabelText, getByText, queryByText } = render(<Fixture />);
-
         const censusTractInput = getByLabelText('Census tract');
 
-        const testCases = [
-            { value: '0000.00', valid: false },
-            { value: '0001.00', valid: false },
-            { value: '0001.01', valid: true },
-            { value: '1000.00', valid: false },
-            { value: '9999.99', valid: false },
-            { value: '9999.98', valid: true },
-            { value: '0001', valid: true },
-            { value: '9999', valid: true },
-            { value: '0000', valid: false },
-            { value: '9999.00', valid: false },
-            { value: '0001.99', valid: false },
-            { value: '1234.56', valid: true }
-        ];
+        userEvent.clear(censusTractInput);
+        userEvent.paste(censusTractInput, value);
+        userEvent.tab();
 
-        for (const { value, valid } of testCases) {
-            userEvent.clear(censusTractInput);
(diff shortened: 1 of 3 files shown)


apps/modernization-ui/src/apps/patient/data/address/AddressEntryFields.tsx
```diff
@@ -206,19 +206,26 @@ export const AddressEntryFields = () => {
             <Controller
                 control={control}
                 name="censusTract"
-                rules={maxLengthRule(10)}
+                rules={{
+                    pattern: {
+                        value: /^\d{4}(\.\d{2})?$/,
```
mpeels, line 211: Issue: does not limit to `.01` - `.98`

mpeels (commented):
Suggestion: Talked to Adam some about the regex. The following was recommended for testing.
```regex
/^(?!0000)(\d{4})(?:\.(?!00|99)\d{2})?$/
```

adamloup-enquizit (changes requested):
The census tract should only allow numeric values.  The `Input` component is only going to apply the pattern if a `mask` is also present.  The mask in this case would be `___.__`

apps/modernization-ui/src/apps/patient/data/address/AddressEntryFields.spec.tsx
```diff
@@ -133,4 +133,23 @@ describe('AddressEntryFields', () => {
...
+    it('should invalidate Census Tract format', async () => {
+        const { getByLabelText, getByText } = render(<Fixture />);
+
+        const censusTractInput = getByLabelText('Census tract');
+
+        act(() => {
+            userEvent.clear(censusTractInput);
+            userEvent.type(censusTractInput, '12345');
+            userEvent.tab();
+        });
```
mpeels, line 146: Question: What are your thoughts on adding some additional test cases?

|Value|Valid|
|----|----|
|0000.00|false|
|0001.00|false|
|0001.01|true|
|1000.00|false|
|1000.01|true|
|9999.99|false|
|9999.98|true|

hclarkEnq: I have added most of these scenarios but the last two from this list as the user would not be permitted to enter more than 4 digits, similarly with the decimals it would not exceed the 2 decimals, the others were helpful though so thanks

apps/modernization-ui/src/apps/patient/data/address/AddressEntryFields.spec.tsx
```diff
@@ -133,4 +133,47 @@ describe('AddressEntryFields', () => {
...
+            { value: '1000.00', valid: false },
+            { value: '9999.99', valid: false },
+            { value: '9999.98', valid: true },
+            { value: '0001', valid: true },
+            { value: '9999', valid: true },
+            { value: '0000', valid: false },
+            { value: '9999.00', valid: false },
+            { value: '0001.99', valid: false },
+            { value: '1234.56', valid: true }
+        ];
```
adamloup-enquizit, line 155: suggestion (non-blocking):  this kind of test can be simplified into a Parameterized test using [test.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout)

adamloup-enquizit (approved):
👍

mpeels (approved):

adamloup-enquizit (approved):
👍

PR #1912, opened 2024-10-09
[CNFT1]-passing display type with name and id for profileLink to change table view for firstColumn

hclarkEnq:
## Description

Adding Patient ID as the first column in the Patient search results - Table view.
Adds hyperlink to Patient ID which routes to patient profile 
Screenshot 2024-10-09 at 8 14 59 AM

## Tickets

* [CNFT1-3290](https://cdc-nbs.atlassian.net/browse/CNFT1-3290)

[CNFT1-3290]: https://cdc-nbs.atlassian.net/browse/CNFT1-3290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
+++ b/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
@@ -3,13 +3,6 @@ import { BrowserRouter } from 'react-router-dom';
 import { PatientSearchResult } from 'generated/graphql/schema';
 import { displayProfileLink, displayPhones, displayEmails } from './patientSearchResult';
 
-jest.mock('name', () => ({
-    displayName: jest.fn(() => jest.fn((name) => `${name.last}, ${name.first}`))
-}));
-jest.mock('address/display', () => ({
-    displayAddress: jest.fn((address) => `${address.street}, ${address.city}`)
-}));
-
 describe('patientSearchResult functions', () => {
     const mockPatient: PatientSearchResult = {
         patient: 10043290,
@@ -57,17 +50,17 @@ describe('patientSearchResult functions', () => {
         emails: ['emily.reynolds@owensborohealth.org']
     };
 
-    test('displayPhones returns correct string', () => {
+    it('should displayPhones returns correct string', () => {
         const result = displayPhones(mockPatient);
         expect(result).toBe('270-685-4067');
     });
 
-    test('displayEmails returns correct string', () => {
+    it('should displayEmails returns correct string', () => {
         const result = displayEmails(mockPatient);
         expect(result).toBe('emily.reynolds@owensborohealth.org');
     });
 
-    test('displayProfileLink renders correctly with shortId', () => {
+    it('should displayProfileLink renders correctly with shortId', () => {
         const { getByText } = render(
             <BrowserRouter>{displayProfileLink(mockPatient.shortId, 'John Doe')}</BrowserRouter>
         );
--- a/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
+++ b/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
@@ -1,0 +1,77 @@
+import { render } from '@testing-library/react';
+import { BrowserRouter } from 'react-router-dom';
+import { PatientSearchResult } from 'generated/graphql/schema';
+import { displayProfileLink, displayPhones, displayEmails } from './patientSearchResult';
+
+jest.mock('name', () => ({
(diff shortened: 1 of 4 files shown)


mpeels (dismissed):

apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.tsx
```diff
@@ -3,11 +3,17 @@ import { displayName } from 'name';
 import { displayAddress } from 'address/display';
 import { Link } from 'react-router-dom';
 
-const displayProfileLink = (result: PatientSearchResult) => (
-    <Link to={`/patient-profile/${result.shortId}/summary`}>
-        {(result.legalName && displayName('fullLastFirst')(result.legalName)) || 'No Data'}
-    </Link>
-);
+const displayProfileLink = (result: PatientSearchResult) => {
+    return <Link to={`/patient-profile/${result.shortId}/summary`}>{result.shortId || 'No Data'}</Link>;
```
adamloup-enquizit, line 7: nit: a patient will always have a `shortId` so the "No Data" should not be required.

suggestion (non-blocking): The `displayProfileLink` function could be changed to accept a `string` instead.  That way showing the link for the Patient ID column would become `displayProfileLink(result.shortId)` and the `displayProfileLegalName` function would call `displayName` and pass that to `displayProfileLink`.  This would remove the duplicate code for creating the link.

adamloup-enquizit (commented):
Could you add a test to make sure that the Patient ID is a link to the patient profile for the expected patient?

apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+                city: 'OWENSBORO',
+                state: 'KY',
+                zipcode: '30309'
+            }
+        ],
+        phones: ['270-685-4067'],
+        emails: ['emily.reynolds@owensborohealth.org']
+    };
+
+    test('displayPhones returns correct string', () => {
```
adamloup-enquizit, line 60: consistency (non-blocking): All of the unit tests (expect one other) follow the pattern of `it('should ` over `test('`

apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { BrowserRouter } from 'react-router-dom';
+import { PatientSearchResult } from 'generated/graphql/schema';
+import { displayProfileLink, displayPhones, displayEmails } from './patientSearchResult';
+
+jest.mock('name', () => ({
+    displayName: jest.fn(() => jest.fn((name) => `${name.last}, ${name.first}`))
+}));
+jest.mock('address/display', () => ({
+    displayAddress: jest.fn((address) => `${address.street}, ${address.city}`)
+}));
```
adamloup-enquizit, line 11: question:  Why mock these functions to do almost what the actual functions do?  I don't see them being used in the test either

adamloup-enquizit (approved):
👍 Looks good

apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.tsx
```diff
@@ -3,11 +3,14 @@ import { displayName } from 'name';
 import { displayAddress } from 'address/display';
 import { Link } from 'react-router-dom';
 
-const displayProfileLink = (result: PatientSearchResult) => (
-    <Link to={`/patient-profile/${result.shortId}/summary`}>
-        {(result.legalName && displayName('fullLastFirst')(result.legalName)) || 'No Data'}
-    </Link>
-);
+const displayProfileLink = (shortId: number, displayName?: string) => {
```
joeaxcdw, line 6: just curious: Does changing the signature break anything, not caught by tests? Would an overload with `(result: PatientSearchResult)` for backwards compat be appropriate here?

hclarkEnq: Yeah and the shortId is always provided when calling this function, so there's no risk of it being undefined.

joeaxcdw: Okay no problem. I just had something like this in mind that would keep it consistent with the other "display" functions:
```
const displayProfileLink = ({ shortId }: PatientSearchResult, displayName?: string) => {
```

joeaxcdw (approved):

PR #1945, opened 2024-10-18
Fixing dropdown input fields

alaapbharadwaj:
## Description

Bug where dropdown selections were still being retained when not expected. Have fixed that with all the dropdowns. Issue was the default value was being set to 'undefined' instead of either an empty selectable or null. So the change was made to all the non required fields, while the required fields just have a selectable of empty string.

## Tickets

* [CNFT1-3403](https://cdc-nbs.atlassian.net/browse/CNFT1-3403)

### DIFF
--- a/apps/modernization-ui/src/apps/patient/add/extended/inputs/address/AddressRepeatingBlock.tsx
+++ b/apps/modernization-ui/src/apps/patient/add/extended/inputs/address/AddressRepeatingBlock.tsx
@@ -5,18 +5,19 @@ import { AddressView } from './AddressView';
 import { AddressEntryFields } from 'apps/patient/data/address/AddressEntryFields';
 import { RepeatingBlock } from 'design-system/entry/multi-value/RepeatingBlock';
 import { ReactNode } from 'react';
+import { asSelectable } from 'options';
 
 const defaultValue: Partial<AddressEntry> = {
     asOf: today(),
-    type: undefined,
-    use: undefined,
+    type: asSelectable(''),
+    use: asSelectable(''),
     address1: '',
     address2: '',
     city: '',
-    state: undefined,
+    state: null,
     zipcode: '',
-    county: undefined,
-    country: undefined,
+    county: null,
+    country: null,
     censusTract: '',
     comment: ''
 };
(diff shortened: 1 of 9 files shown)


joeaxcdw (approved):

apps/modernization-ui/src/apps/patient/add/extended/inputs/address/AddressRepeatingBlock.tsx
```diff
@@ -5,18 +5,19 @@ import { AddressView } from './AddressView';
...
 import { RepeatingBlock } from 'design-system/entry/multi-value/RepeatingBlock';
 import { ReactNode } from 'react';
+import { asSelectable } from 'options';
 
 const defaultValue: Partial<AddressEntry> = {
     asOf: today(),
-    type: undefined,
-    use: undefined,
+    type: asSelectable(''),
+    use: asSelectable(''),
```
adamloup-enquizit, line 13: issue: This cannot be an empty `Selectable`, it has to be `null` or `undefined`.  An empty `Selectable` is a value which is bypassing the validations.

image

adamloup-enquizit (changes requested):
Can't use an empty `Selectable` as a default value, they are values.  The entry types will have to allow a `null` value for those fields.

apps/modernization-ui/src/apps/patient/data/identification/asIdentification.ts
```diff
@@ -6,7 +6,7 @@ const asIdentification = (entry: IdentificationEntry): Identification => {
     const { type, issuer, ...remaining } = entry;
 
     return {
-        type: asValue(type),
+        type: asValue(type) ?? '',
```
adamloup-enquizit, line 9: suggestion:  I encountered a similar thing with [asEthnicity](https://github.com/CDCgov/NEDSS-Modernization/blob/ca26e4ce518d92a3fca50bf26d7546675f13d75b/apps/modernization-ui/src/apps/patient/data/ethnicity/asEthnicity.ts#L8) and decided to make it return `undefined` if the required values were not present.

adamloup-enquizit:
I am the captain now.  When reviewing the changes there were a instances where `use` and `type` values were made to be optional.  Those values have to exist on the types, changing them to nullable caused some cascading changes.

The individual transformers were changed to conditionally map based on the presence of specific values

A few helper functions (with tests) were also added for these changes.

- `exists` determines the existence of an value while asserting the type.  It's opinionated on what constitutes existence for certain values. 
- `isEmpty` was changed to use `exists` on each property to test for emptiness  
- `maybeMapAll` was changed to ensure that the result of the mapping function is added only if it exists.

adamloup-enquizit (approved):

mpeels (approved):
🍌

stevegsa (approved):

PR #1941, opened 2024-10-18
Cnft1 3277 UI add patient name column

benlam-ignw:
## Description

AC fulfilled: 
User should be able to see Names as a column, in the patient search results table view.
User should see Names as a column in the ‘Column settings’

## Tickets

https://cdc-nbs.atlassian.net/browse/CNFT1-3277

## Checklist before requesting a review
- [x] PR focuses on a single story
- [x] Code has been fully tested to meet acceptance criteria
- [x] PR is reasonably small and reviewable (Generally less than 10 files and 500 changed lines)
- [x] All new functions/classes/components reasonably small
- [x] Functions/classes/components focused on one responsibility
- [x] Code easy to understand and modify (clarity over concise/clever)
- [x] PRs containing TypeScript follow the [Do's and Don'ts](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)
- [x] PR does not contain hardcoded values (Uses constants)
- [x] All code is covered by unit or feature tests

PR #1946, opened 2024-10-18
CNFT1-3349: Header for identifications in search

joeaxcdw:
## Description

Implement headers for identifications in search, similar to names and addresses.

image

**Note: Approve https://github.com/CDCgov/NEDSS-Modernization/pull/1941 first**

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-3349)

### DIFF
--- a/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
+++ b/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
@@ -6,7 +6,8 @@ import {
     displayPhones,
     displayEmails,
     displayAddresses,
-    displayOtherNames
+    displayOtherNames,
+    displayIdentifications
 } from './patientSearchResult';
 
 describe('patientSearchResult functions', () => {
@@ -28,7 +29,16 @@ describe('patientSearchResult functions', () => {
                 type: 'Alias'
             }
         ],
-        identification: [],
+        identification: [
+            {
+                type: 'SSN',
+                value: '123-45-6789'
+            },
+            {
+                type: 'MRN',
+                value: '123456'
+            }
+        ],
         addresses: [
             {
                 use: 'Home',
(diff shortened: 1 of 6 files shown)


apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
```diff
@@ -89,4 +99,12 @@ describe('patientSearchResult functions', () => {
         expect(getByText('Alias')).toBeInTheDocument();
         expect(getByText('Johnny TestnullTest')).toBeInTheDocument();
     });
+
+    it('should render identifications correctly', () => {
```
adamloup-enquizit, line 103: suggestion (non-blocker):  can you specify what "correctly" is?  This is testing that an identification is displayed titled by it's type

apps/modernization-ui/src/design-system/item/ItemGroup.tsx
```diff
@@ -1,15 +1,20 @@
...
+const isAddressType = (type?: ItemType): type is AddressableType =>
+    type != null && ['address', 'phone', 'email'].includes(type);
+
 const ItemGroup = ({ type, label, children }: Props) => {
+    const isAddress = isAddressType(type);
     return (
         <div className={styles.itemgroup} {...(type && { 'data-item-type': type })}>
             {label && <header>{label}</header>}
-            <p>{children}</p>
+            {isAddress ? <address>{children}</address> : <p role="group">{children}</p>}
```
adamloup-enquizit, line 17: question (non-blocker): Is this knowing too much information about the caller?  Should `ItemGroup` render `children` with in a `<p>` if it's a string otherwise just render `children?  This would allow the caller to specify that the child element is an address or not.

I'm not sure if that's the correct use of the group role, I think it's intended to be on the parent element containing multiple elements of some kind

joeaxcdw: After further research <address> is not appropriate here, so I reverted to how it was before.

apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.spec.tsx
joeaxcdw: Fixed

adamloup-enquizit (approved):

hclarkEnq (approved):

PR #1964, opened 2024-10-24
[CNFT1-3423]: correcting content for delete modal in patient profile

hclarkEnq:
## Description

1.  Delete patient modal, message displays a “,“ at the end of message. Replace “, “ with a  “.“
2. When there is no name for a patient, then the modal is displayed as:  Patient Id Number followed by ",," -> Solution: When there is no patient name then the message is displayed as “Patient ID, --,--. “
3. When there is no last name/ first name for a patient, then the modal is displayed as: "FirstName, " or "LastName," -> Solution: When there is no patient name - last name/ first name then the message is displayed as “Patient ID, test,--. “/ “Patient ID, --,test. “
4. When patient name is updated in Patient profile, Its not reflecting when deleting a patient:

Patient name was updated from “test,--” to “Something,--” but the modal is still displayed as: "test,--" -> Solution: When there is a change in the patient name then delete patient modal has to have the updated patient name.

## Tickets

* [CNFT1-3423](https://cdc-nbs.atlassian.net/browse/CNFT1-3423)

[CNFT1-3423]: https://cdc-nbs.atlassian.net/browse/CNFT1-3423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/apps/patient/profile/delete/DeletePatient.tsx
+++ b/apps/modernization-ui/src/apps/patient/profile/delete/DeletePatient.tsx
@@ -1,4 +1,4 @@
-import { DeletePatientMutation, PatientSummary, useDeletePatientMutation } from 'generated/graphql/schema';
+import { DeletePatientMutation, useDeletePatientMutation } from 'generated/graphql/schema';
 import { Icon } from 'design-system/icon';
 import { Confirmation, Warning } from 'design-system/modal';
 import { useAlert } from 'alert';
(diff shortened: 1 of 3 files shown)


joeaxcdw (dismissed):

apps/modernization-ui/src/apps/patient/profile/delete/DeletePatient.tsx
```diff
@@ -11,23 +11,23 @@ import { DeletabilityResult, resolveDeletability } from './resolveDeletability';
...
     patient: Patient;
-    summary: PatientSummary;
 };
 
-const DeletePatient = ({ patient, summary }: Props) => {
+const DeletePatient = ({ patient }: Props) => {
     const { showSuccess, showError } = useAlert();
-
     const { go } = useSearchNavigation();
+    const { summary } = useProfileContext();
```
adamloup-enquizit, line 23: question:  Should the context be used here just to get the summary?  Why not pass the`legalName` into the component?  That will reduce the mocks needed in the unit test....speaking of which

adamloup-enquizit (changes requested):
Can you add unit tests to validate these changes?

apps/modernization-ui/src/apps/patient/profile/delete/DeletePatient.tsx
hclarkEnq: Yes Ill add the unit tests, but I was thinking on that context, since `useProfileContext();` is used in the provider of the DeletePatient component in the patient profile component, so if I pass the legalName using the summary from the patient profile to the deletePatient then it will store the previous value when updating the name in demographics not the actual value.

adamloup-enquizit (approved):
👍 Thanks for the extra testing

alaapbharadwaj (approved):

PR #1965, opened 2024-10-24
CNFT1-331: Create operator dropdown component

joeaxcdw:
## Description

Implement the base operator dropdown to use for first name, last name, and applicable address fields in patient search.

image

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-3317)


apps/modernization-ui/src/options/operator/operators.ts
```diff
@@ -0,0 +1,26 @@
+import { Selectable, asSelectable, findByValue } from 'options';
+
+const STARTS_WITH_OPERATOR = asSelectable('S', 'Starts with');
```
joeaxcdw, line 3: @stevegsa please let me know what the actual values will be for these operators so I can fix it here.

adamloup-enquizit (commented):
This component is more than a dropdown of operators, it's encapsulating a specific use case searching using text in a specific way.  This is going to be a repeatable use case so the component should produce an object representing the intention of searching with this criteria.

At least something like 

```json
 {
 "value": "some value",
 "operator": { "name": "Starts with", "value" : "SW"}
}
```  

Even better would be something way more expressive 

```json
{
 "startsWith" :  "some-value"
}
```

---

Since this is going to represent a common usage pattern it should be in the `design-system` folder.

joeaxcdw:
Recent updates:

- Added `OperatorInput` control that combines the operator select and input
- Updated operation names and object types to match:
```
"city": {
  "startsWith": "some city"
 }
```

adamloup-enquizit (approved):
👍 Cool, can't wait to use it

stevegsa (approved):

PR #2194, opened 2025-01-21
change work phone extension to 1 to 8 digits

benlam-ignw:
## Description

UI: New patient: Phone & email: Work phone extension: character limit has to be 8.
Add character limit (8) to the Extension field. After character 8, user should not be allowed to enter any data.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-3672)


joeaxcdw (commented):
What about extended form, does that need to change?

PR #2286, opened 2025-02-21
Cnft1 3777 Add ComponentSizingProvider to PatientDataProviderWrapper to get sizing in add patient and add patient extended

benlam-ignw:
## Description

Quick fix to apps/modernization-ui/src/apps/patient/add/PatientAddRouting.tsx in order to be able to access sizing within add patient and add patient extended

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-3777)


stevegsa (approved):

adamloup-enquizit (approved):
Ah, the context has an initial value set so it will never be `undefined` making the provider check in the `useComponentSizing` hook useless and making it difficult to see what the actual problem is.

Also, unless there are merge conflicts you shouldn't have to update the PR branch.

PR #2304, opened 2025-02-26
[CNFT1-3855]: Adding dashes to id when no data

hclarkEnq:
## Description

When a patient does not have any data for ID, then in patient search results, ID column has to have "---".

## Tickets

* [CNFT1-3855](https://cdc-nbs.atlassian.net/browse/CNFT1-3855)

### DIFF
--- a/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.tsx
+++ b/apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.tsx
@@ -4,6 +4,7 @@ import { displayAddress } from 'address/display';
 
 import { ItemGroup } from 'design-system/item';
 import { internalizeDate, displayAgeAsOfToday } from 'date';
+import { NoData } from 'components/NoData';
 
 // Displays Other names, that are not the legal name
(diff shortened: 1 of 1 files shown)


apps/modernization-ui/src/apps/search/patient/result/patientSearchResult.tsx
```diff
@@ -60,15 +62,15 @@ const displayPatientAge = (result: PatientSearchResult, variant: 'singleline' |
 
 // Returns JSX that represents a list of IDs to display
 const displayIdentifications = (result: PatientSearchResult): JSX.Element => (
-    <div>
```
adamloup-enquizit, line 63: The `DataTableRow` component is [handling the display of the dashes](https://github.com/CDCgov/NEDSS-Modernization/blob/abc5073b674b910872518884d920d798ce45ba19/apps/modernization-ui/src/design-system/table/DataTableRow.tsx#L47) if the table cell is `undefined`.  The issue here is that the identifications are wrapped in a `<div>` so the cell contents are never `undefined`.  The table should be responsible for displaying the dashes so that it remains consistent for each cell.  

Is that `<div>` necessary?

alaapbharadwaj (approved):

stevegsa (approved):

benlam-ignw (approved):

PR #2357, opened 2025-03-13
Add support for secondary and secondary destructive

mpeels:
## Description

1. Adds `secondary` prop to design system's `Button` component. This prop will be replacing `outline` in the future.
2. Adds styling for when both `secondary` and `destructive` is specified.

Standard
image

Outline
image

Secondary
image

Destructive
image

Outline Destructive
image

Secondary Destructive
image


adamloup-enquizit (approved):
👍 <sup>💥<sup>💥</sup> </sup>

snehaasjournal (approved):
Looks good to me 🏁

PR #2364, opened 2025-03-13
Add SubText to Card Component

snehaasjournal:
## Description

This PR introduces an optional subtext prop to the Card component. When provided, the subtext will be rendered directly below the card's title. If no subtext is passed, the component will function as it did previously, without rendering the subtext section.

Additionally, there are some changes to the `border-radius` etc. but it is propagated only when subtext is included.

Use case for this change:
Mactching Criteria Pages like~
use case 1

use case 2

Added unit tests to verify that:
    Subtext only renders when the subtext prop is provided.
    The component's existing behavior is not affected when the subtext prop is omitted.

## Tickets

* [Jira Ticket](url)

### DIFF
--- a/apps/modernization-ui/src/design-system/card/Card.tsx
+++ b/apps/modernization-ui/src/design-system/card/Card.tsx
@@ -14,10 +14,8 @@ export const Card = ({ id, title, info, subtext, children, level = 2 }: Props) =
     return (
         <section id={id} className={styles.card}>
             <header>
-                <div>
-                    <Heading level={level}>{title}</Heading>
-                    {subtext && <div className={styles.subtext}>{subtext}</div>}
-                </div>
+                <Heading level={level}>{title}</Heading>
+                {subtext && <div className={styles.subtext}>{subtext}</div>}
                 {info}
             </header>
             {children}
--- a/apps/modernization-ui/src/design-system/card/Card.tsx
+++ b/apps/modernization-ui/src/design-system/card/Card.tsx
@@ -11,14 +11,14 @@ type Props = {
     children: ReactNode;
 };
 export const Card = ({ id, title, info, subtext, children, level = 2 }: Props) => {
-    const cardClassName = subtext ? `${styles.card} ${styles.withSubtext}` : styles.card;
-
     return (
-        <section id={id} className={cardClassName}>
+        <section id={id} className={styles.card}>
             <header>
-                <Heading level={level}>{title}</Heading>
+                <div>
+                    <Heading level={level}>{title}</Heading>
+                    {subtext && <div className={styles.subtext}>{subtext}</div>}
+                </div>
                 {info}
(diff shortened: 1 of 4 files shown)


mpeels (approved):

apps/modernization-ui/src/design-system/card/card.module.scss
```diff
@@ -4,6 +4,7 @@
 .card {
     @extend %thin;
     background-color: colors.$base-white;
+    @include borders.rounded();
```
adamloup-enquizit, line 7: praise: Good catch!

apps/modernization-ui/src/design-system/card/card.module.scss
```diff
@@ -13,3 +14,10 @@
         @extend %thin-bottom;
     }
 }
+
+.subtext {
+    font-size: 14px;
```
adamloup-enquizit, line 19: suggestion:  we typically try to use `rem` for sizing.  The designers work in px because that's what Figma uses.  The `font-size` should be `0.875rem` and the `line-height` should be `1.3125rem`.

apps/modernization-ui/src/design-system/card/Card.tsx
```diff
@@ -6,14 +6,18 @@ type Props = {
...
     level?: HeadingLevel;
     children: ReactNode;
 };
-export const Card = ({ id, title, info, children, level = 2 }: Props) => {
+export const Card = ({ id, title, info, subtext, children, level = 2 }: Props) => {
     return (
         <section id={id} className={styles.card}>
             <header>
-                <Heading level={level}>{title}</Heading>
+                <div>
```
adamloup-enquizit, line 17: question:  is this `<div>` necessary?  The `<header>` is already containing the two elements making the `<div>` seem redundant.  The addition of this `<div>` is also what cased the `AddPatientExtendedForm` tests to fail.

apps/modernization-ui/src/design-system/card/card.module.scss
```diff
@@ -13,3 +14,10 @@
...
+    line-height: 21px;
+    color: colors.$base-darker;
+    margin-top: 0.25rem;
```
adamloup-enquizit, line 22: suggestion:  We try not to use margin and padding to define spacing between components.  Since `<header>` is already a `flex` element the `gap` property can be used to define this spacing.  On a related note, I don't see that `0.25rem` spacing in the Designs for the Card component.

apps/modernization-ui/src/design-system/card/Card.tsx
adamloup-enquizit: Nevermind, I see why it's needed.

adamloup-enquizit:
PR #2366 is also making some changes to the Card component so there may be some merge conflicts.

adamloup-enquizit (approved):
Looks good! 👍 📇

mpeels (approved):

PR #2380, opened 2025-03-19
[CNFT1-3892]: Changing input type to text to allow maxLength to work

hclarkEnq:
## Description
 **Number of adults in residence** and **Number of children in residence**
adding conditional check that prevents users from entering more digits than allowed by the max prop. If the input's text length would exceed the specified maximum which is 99999

also added max validation to handle when a value greater than 99999 is entered. We should only be preventing typing when a value has a format. as per https://github.com/CDCgov/NEDSS-Modernization/pull/2380#pullrequestreview-2699943338

## Tickets

* [CNFT1-3892](https://cdc-nbs.atlassian.net/browse/CNFT1-3892)

[CNFT1-3892]: https://cdc-nbs.atlassian.net/browse/CNFT1-3892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/apps/patient/data/general/GeneralInformationEntryFields.tsx
+++ b/apps/modernization-ui/src/apps/patient/data/general/GeneralInformationEntryFields.tsx
@@ -80,7 +80,10 @@ export const GeneralInformationEntryFields = ({ orientation = 'horizontal', sizi
             <Controller
                 control={control}
                 name="general.adultsInResidence"
-                rules={{ min: { value: 0, message: 'Must be greater than 0' } }}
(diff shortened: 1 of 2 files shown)


adamloup-enquizit (commented):
question: Should the `max` property be used so that screen readers have more information on this field?  It may also need a `max` validation to handle when a value greater than 99999 is entered.  We should only be preventing typing when a value has a format.

PR #2462, opened 2025-04-17
CNFT1-4014: Collapsible card component

joeaxcdw:
## Description

Create the `CollapsibleCard` component.

image

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-4014)


adamloup-enquizit (commented):
question:  Was there a design review done for this yet?  I ask because that separator is only supposed to show up if the card is collapsible and there are actions.  It's part of the story that adds the actions.

apps/modernization-ui/src/design-system/card/Card.tsx
```diff
@@ -14,15 +14,19 @@ type Props = {
 };
 export const Card = ({ id, title, info, subtext, children, className, level = 2 }: Props) => {
```
mpeels, line 15: Suggestion: Can we just add a `collapsible?: boolean` that defaults to false to the `Card` component with a conditional render of a collapse icon? This prevents having to create additional components and is backwards compatible with all other uses of the `Card` component.

mpeels (changes requested):

apps/modernization-ui/src/design-system/card/Card.tsx
joeaxcdw: Short answer: no. This Card component has a bunch of existing properties (title, info, subtext, level) that won't work with the new component. This is why I created "BaseCard" to centralize the existing styling. Also the existing Card component is not supposed to be collapsible.

joeaxcdw:
> question: Was there a design review done for this yet? I ask because that separator is only supposed to show up if the card is collapsible and there are actions. It's part of the story that adds the actions.

The story says no design review required (which I find strange), but I will do anyway just in case. However I laid the ground work for hiding the separator [here](https://github.com/CDCgov/NEDSS-Modernization/pull/2462/files#diff-486c90c544a8d86948769f0ae7859897322a631de5b5980e087402b584d62e0fR43), but was planning to implement in the actions story:
```
    &.no-border {
        border-left: none;
    }
```

apps/modernization-ui/src/design-system/card/Card.tsx
mpeels: I have too many questions for the design team to fully understand the plans so i'll defer to @adamloup-enquizit.

My initial thoughts were:
1. `info` is used in the add patient forms to display a `Required` label. If this is changing, can probably be removed and those Cards updated to new design
1. `level` is only used in the Card storybook and if only `h2` is supported in design, should be removed.
1. `subtext` is still a valid property in Card design system. Though it should be converted from `string` to `ReactNode | ReactNode[]`

joeaxcdw: I can make the CollapsibleCard the base component. @adamloup-enquizit thoughts?

joeaxcdw: `CollapsibleCard` is now the base component.

apps/modernization-ui/src/design-system/card/collapsible/CollapsibleCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+    className,
+    collapsible = true,
+    showCollapseSeparator
+}: CollapsibleCardProps) => {
+    const [height, setHeight] = useState('auto');
+    const [collapsed, setCollapsed] = useState<boolean>(false);
+    const contentRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+        const curHeight = contentRef.current?.scrollHeight;
```
hclarkEnq, line 30: Suggestion: Is it cool if we re-name this maybe to currentHeight? or something like that I got confused the first time I read through it even though it was obvious after.

hclarkEnq (dismissed):

apps/modernization-ui/src/design-system/card/collapsible/CollapsibleCard.tsx
joeaxcdw: Renamed to `currentHeight` for clarity

joeaxcdw:
I have resolved everyone's concerns:

- CollapsibleCard is now the base component
- Separator is off by default
- Confusing variable names renamed

hclarkEnq (approved):

adamloup-enquizit (approved):

joeaxcdw:
> If `CollapsibleCard` is the base component and has a property to make it not collapsible, isn't it just a `Card`?

I agree but `Card` is being used in 20+ places. We need a follow-on PR to reorg this with all the pages that are using it.

mpeels (approved):

PR #2474, opened 2025-04-21
CNFT1-3992-aria-role-button

hclarkEnq:
adding aria role of button and fixing tests to match amount of  getAllByRole('button') expected (buttons) count for in the repeating block spec file

**ARIA label has to be updated for the Repeating block table -icons:**

> View
> Edit
> Delete
### AXE Scan before Fix
Screenshot 2025-04-21 at 1 29 17 PM

### AXE Scan after Fix
Screenshot 2025-04-21 at 1 27 16 PM

## Tickets

* [CNFT1-3992](https://cdc-nbs.atlassian.net/browse/CNFT1-3992)

[CNFT1-3992]: https://cdc-nbs.atlassian.net/browse/CNFT1-3992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/design-system/entry/multi-value/RepeatingBlock.spec.tsx
+++ b/apps/modernization-ui/src/design-system/entry/multi-value/RepeatingBlock.spec.tsx
@@ -357,11 +357,11 @@ describe('RepeatingBlock', () => {
         expect(getByText('Render view first value: first-value')).toBeInTheDocument();
         expect(getByText('Render view second value: second-value')).toBeInTheDocument();
 
-        const buttons = getAllByRole('button');
-        expect(buttons).toHaveLength(4);
-        expect(buttons[3]).toHaveTextContent('Add test title');
-        expect(buttons[3].innerHTML).toContain('svg');
-        expect(buttons[3]).toHaveAttribute('aria-description');
+        const buttons = getAllByRole('button', { name: /Add/ });
+        expect(buttons).toHaveLength(1);
+        expect(buttons[0]).toHaveTextContent('Add test title');
+        expect(buttons[0].innerHTML).toContain('svg');
+        expect(buttons[0]).toHaveAttribute('aria-description');
     });
 
     it('should render edit when edit icon clicked', async () => {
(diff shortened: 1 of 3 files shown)


apps/modernization-ui/src/design-system/entry/multi-value/RepeatingBlock.tsx
```diff
@@ -104,21 +104,21 @@ const RepeatingBlock = <V extends FieldValues>({
         className: styles.iconColumn,
         render: (value: V) => (
             <div className={classNames(styles.actions, sizing && styles[sizing])}>
-                <div data-tooltip-position="top" aria-label="View" onClick={() => view(value)}>
+                <div data-tooltip-position="top" aria-label="View" role="button" onClick={() => view(value)}>
```
adamloup-enquizit, line 107: suggestion (non-blocking): I'm not sure the `<div>` is necessary here anymore, we might be able to get away with moving all of these properties to the `Icon`

hclarkEnq: Okay let me check it out

hclarkEnq: I think we need the <div> because if I remove it and put the properties on the Icon the tooltip won't work in that case. So basically I kept the div to ensure it works with the tooltips.

apps/modernization-ui/src/design-system/entry/multi-value/RepeatingBlock.spec.tsx
```diff
@@ -386,12 +386,12 @@ describe('RepeatingBlock', () => {
         expect(getByLabelText('Second Input')).toBeInTheDocument();
 
         const buttons = getAllByRole('button');
-        expect(buttons).toHaveLength(2);
-        expect(buttons[0]).toHaveTextContent('Update test title');
-        expect(buttons[0]).toHaveAttribute('aria-description');
-        expect(buttons[0].innerHTML).not.toContain('svg');
-        expect(buttons[1]).toHaveTextContent('Cancel');
-        expect(buttons[1]).toHaveAttribute('aria-description');
+        expect(buttons).toHaveLength(5);
```
joeaxcdw, line 389: We should do a regex pattern here or selector so that it only selects the 2 buttons in the footer since this is what the test is focused on.
Something like this:
`const buttons = getAllByRole('button', { name: /(Update*|Cancel)/ });`

joeaxcdw (approved):

adamloup-enquizit (dismissed):
👍 Looks good, added a comment regarding testing the buttons

joeaxcdw (approved):

stevegsa (approved):

PR #2482, opened 2025-04-23
CNFT1-4015: TableCard component + actions

joeaxcdw:
## Description

Implement the base `TableCard` component. Implement the actions for managing the table.

image

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-4015)


hclarkEnq:
can we use the collapsible card directly because we are just using the data table, the number count (tags), and table card header but Im thinking maybe if just leverage our existing card with the collapsible card would this work? 

not a blocking question just curious?

hclarkEnq:
> can we use the collapsible card directly because we are just using the data table, the number count (tags), and table card header but Im thinking maybe if just leverage our existing card with the collapsible card would this work?
> 
> not a blocking question just curious?

I think in our Card component we can say in there whether or not it has a collapsible or not I think your doing that too and it will be having headers, actions, numbers, and children. So why can't we use one card here instead of having different components used at the time of display since I heard from Henry T last meeting that it might be better to have a common component used instead of having different versions but just a thought because I also remember you mentioning that this is following the same structure as the repeating blocks.

joeaxcdw:
> > can we use the collapsible card directly because we are just using the data table, the number count (tags), and table card header but Im thinking maybe if just leverage our existing card with the collapsible card would this work?

1 - We will have one universal Card eventually. Right now existing Card is being used extensively

2 - TableCard is like RepeatingBlock in that it is a specialized component with will handle the settings function for sorting/showing columns, showing the row count, and possible a few other functions in the future.

hclarkEnq (approved):

adamloup-enquizit (approved):

PR #2536, opened 2025-05-12
keyboard navigation accessibilty improvement

hclarkEnq:
## Description
This PR improves keyboard navigation accessibility automatically focusing the "Skip to main content" link when a user first presses the Tab key after page navigation.

Added logic to detect the first Tab key press on each page because currently if we navigate to add patient or extended the skip link wont work on the first attempt of using it by clicking tab to activate it.

## Tickets

* [CNFT1-4006t](https://cdc-nbs.atlassian.net/browse/CNFT1-4006)


apps/modernization-ui/src/layout/ScrollToTop.tsx
```diff
@@ -1,11 +1,41 @@
-import { ReactNode, useEffect } from 'react';
+import { ReactNode, useEffect, useState } from 'react';
 import { useLocation } from 'react-router';
 
 const ScrollToTop = ({ children }: { children: ReactNode }) => {
     const location = useLocation();
+    const [tabPressed, setTabPressed] = useState(false);
```
benlam-ignw, line 6: Maybe we could name this something like `initialTabPress` or `firstTabInput`?

apps/modernization-ui/src/layout/ScrollToTop.tsx
```diff
@@ -1,11 +1,41 @@
...
+
+            if (event.code === 'Tab' && !tabPressed && isBodyOrNonFocusable) {
+                setTabPressed(true);
+                const skipLinkElement: HTMLElement | null = document.querySelector('.usa-skipnav');
+                if (skipLinkElement) {
+                    requestAnimationFrame(() => {
+                        skipLinkElement.focus();
+                    });
+                }
+                window.removeEventListener('keydown', handleFirstTab);
```
benlam-ignw, line 29: the `return () => { window.removeEventListener('keydown', handleFirstTab); };` should always remove the event listener anyways but if the useEffect is interrupted by something then it might not work. I think its a good idea to keep this `removeEventListener` but have it be removed when the `event.code === 'Tab'` is fired even if the other conditions are not true

benlam-ignw (approved):
Thanks! 😄

PR #2549, opened 2025-05-14
CNFT1-4075: TableCard enable client-side sorting in DataTable

joeaxcdw:
## Description

Allow client-side sorting in the `DataTable` for `TableCard`.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-4075)


hclarkEnq (approved):

apps/modernization-ui/src/design-system/table/SortableDataTable.tsx
```diff
@@ -0,0 +1,32 @@
+import { ComponentType } from 'react';
+import { maybeSortData, useSorting, SortingProvider } from 'sorting';
+import { DataTable, DataTableProps } from './DataTable';
+
+export const withSortable = <T,>(Component: ComponentType<DataTableProps<T>>) => {
+    const SortableComponent = (props: DataTableProps<T>) => {
+        const { property, direction } = useSorting();
+        const column = props.columns.find((column) => column.id === property);
+        const comparator = column?.comparator;
+        const sortedData = maybeSortData(props.data, comparator, property as keyof T, direction);
```
adamloup-enquizit, line 10: issue:  This assumes that the `property` of a `Column<V>` is a key of `V` which may not always be the case

apps/modernization-ui/src/design-system/table/SortableDataTable.tsx
```diff
@@ -0,0 +1,32 @@
...
+        return <Component {...props} data={sortedData} />;
+    };
+
+    SortableComponent.displayName = 'SortableComponent';
+    return SortableComponent;
```
adamloup-enquizit, line 15: question (non-blocking):  What's the benefit of enhancing the component with a function like this over just creating a new component that renders the `DataTable` directly?

apps/modernization-ui/src/design-system/table/SortableDataTable.tsx
joeaxcdw: I'm not sure how you want me to fix this. In what scenario would it not be `keyof T`? You could enforce type safety in the `useSorting()` hook but I am just working with what's already there.

adamloup-enquizit: In any situation where the content of a single column is made up of various data from the object being shown.  We would need a way to get the raw value needed to sort by.  This is actually giving me an idea that could simplify how `DataTable` does rendering

joeaxcdw: Yeah, no I agree, but the functions in `Sort.ts` assume you are reading a property of an object. We could a `Column.getValue()?` that could be undefined and just fallback to `column.id` to get the property.

joeaxcdw: Tried to make it clearer in the code that if property can't be resolved it will be undefined; nothing will get sorted.

adamloup-enquizit (approved):
👍 ⬇️ ⬆️

PR #2557, opened 2025-05-16
Cnft1 4073 UI summary open investigations

benlam-ignw:
## Description

This PR adds the Patient File Open Investigations card to the summary tab. 

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/CNFT1-4073)


apps/modernization-ui/src/apps/patient/file/patientData/useFindOpenInvestigations.ts
alaapbharadwaj, line 1: Suggestion: We can use a useReducer here instead, it will be easier to keep track of the state given so much is happening on this page already. Can take a look at 'usePatientFileSummary.tsx' for an example

apps/modernization-ui/src/apps/patient/file/patientData/useFindOpenInvestigations.ts
```diff
@@ -0,0 +1,46 @@
...
+    const [error, setError] = useState<string | undefined>(undefined);
+
+    useEffect(() => {
+        fetchOpenInvestigations();
+    }, []);
+
+    const fetchOpenInvestigations = async () => {
+        setLoading(true);
+
+        fetch(`${Config.modernizationUrl}/nbs/api/patient/${patientId}/investigations/open`, {
```
alaapbharadwaj, line 17: Issue(blocking): Can use the service that is generated automatically instead of using 'fetch' to call the endpoint. Here you should be able to use: PatientOpenInvestigationsService.patientOpenInvestigation()

apps/modernization-ui/src/apps/patient/file/summary/PatientFileSummary.tsx
```diff
@@ -1,14 +1,17 @@
...
 export const PatientFileSummary = () => {
     const { id } = usePatient();
 
     const { summary } = usePatientFileSummary(id);
     return (
-        <body>
+        <div>
             <PatientSummaryContent summary={summary} />
-        </body>
+            <br />
```
alaapbharadwaj, line 13: Issue(blocking): Instead of a break line here, we can give the <body>/<div> tag a classname and give it a gap of 1rem as per the designs

alaapbharadwaj (changes requested):

apps/modernization-ui/src/apps/patient/file/patientData/useFindOpenInvestigations.ts
benlam-ignw: Thanks for the pointer, I've updated the hook to utilize a useReducer similar to usePatientFileSummary in usePatientFileOpenInvestigations.tsx

apps/modernization-ui/src/apps/patient/file/patientData/useFindOpenInvestigations.ts
benlam-ignw: Yes good call, I've updated the hook to use 
`import { PatientInvestigation, PatientOpenInvestigationsService } from 'generated';`

apps/modernization-ui/src/apps/patient/file/summary/PatientFileSummary.tsx
benlam-ignw: Thanks, forgot to change this over to a proper style. Resolved now with summary-layout style in patient-file-summary.module.scss

alaapbharadwaj (approved):

PR #2651, opened 2025-06-13
Nis 8 display error messages for date of birth validations

benlam-ignw:
## Description

Instead of using a display change, we can replace the `validateAll` function here with a validator array that collects all of the errors that we need to check against and puts them into a single array that we show to the user.

This will show the errors in the way that matches the updated figma design:

image

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/NIS-8)


adamloup-enquizit (approved):
👍 Looks good!

PR #2837, opened 2025-08-03
Fix incorrect join in Merge History API causing mergedByUser to be null

sizatenquizit:
## Description

Modified the SQL query to update the join condition between the person_merge table and the auth_user table.

## Tickets
https://cdc-nbs.atlassian.net/browse/CND-413


mpeels (approved):

mpeels (changes requested):
Need to discuss this a bit further. After i reviewed i had a realization.

PR #2914, opened 2025-08-19
Null check for ethnicity

mpeels:
## Description

Adds null check to prevent NPE when an `as_of_date_ethnicity` is set for a patient but no `ethnic_group_ind` is set.


adamloup-enquizit (approved):
👍 Looks good

adamloup-enquizit (approved):
😍 Looks great!

PR #2941, opened 2025-08-28
[FIX] Extraneous Patient History

adamloup-enquizit:
## Description

Ensures that history is not recorded for entities that are not affected when a patient is edited.  The audit values were being updated for multi-value demographics which was causing history records to be inserted when the actual demographic values didn't change.  Now multi-value demographics will only be updated if there is a change to the demographic data.


mpeels (approved):
📚

PR #2942, opened 2025-08-28
[FIX] TextArea null handling

adamloup-enquizit:
## Description

Resolves an issue where pressing backspace on the last character of a `TextArea` would revert to the previous value.  The state value was removed to work more like `TextInput`.


mpeels (changes requested):
brokeded

mpeels (approved):

PR #2953, opened 2025-09-04
[FIX] Handling deletes from classic modal

adamloup-enquizit:
## Description

Adds support for delete actions within a `ClassicModal`.  NBS relies on a form action being performed in order to execute the delete for events displayed within a `ClassicModal`.  The `action` is now invoked if specified by the page being displayed.


adamloup-enquizit:
> Nice fix.

Nice find!

PR #2995, opened 2026-01-28
test: temp fix - give our test patient a birthday!

mcmcgrath13:
## Description

We should mock out the current date, but I'm not sure if there are knock on effects for that, so for the quick moment to get tests passing again, change the expected age from 12 to 13.


emyl3 (approved):
🍰

PR #3052, opened 2026-03-06
ci: submit java depedencies

mcmcgrath13:
## Description

Submit the dependency graph so dependabot can hopefully do things.

Base on [these gradle dependabot docs](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#gradle)


.github/workflows/java-dependency-submission.yaml
```diff
@@ -0,0 +1,22 @@
+name: Dependency Submission
```
mcmcgrath13: we might not actually need this - I saw that dependabot was working for java in the DI repo, so changed the dependabot yaml to look more places

thoughts?

JordanGuinn (approved):

PR #3068, opened 2026-03-10
ci: be smarter about when we run some workflows

mcmcgrath13:
## Description

Only run java workflows on java PRs


JordanGuinn (approved):
Good call ✅ Maybe worth adding `Java` to the titles of these workflows/files in some capacity too?

PR #3069, opened 2026-03-10
style: run and enforce formatting in `modernization-ui`

mcmcgrath13:
## Description

Update the calling/formatting rules (barely) and add enforcement in e49134aa9893db89815f21d54621db9d8aa6b7cc. In the following commit run the formatter. 

In a follow on PR, I'll gitblame ignore the merged commit

Notes:
* the `eslint-plugin-prettier` is not recommended anymore by prettier and they suggest just using `prettier`
* the `eslint-config-prettier` isn't in use (and doesn't look like we need)
* `eslint-config-google` wasn't used


brick-green (approved):
lgtm

PR #3074, opened 2026-03-10
feat: spike - mounting custom libraries

mcmcgrath13:
## Description

This PR shows that we are able to mount a custom library into the expected location and run that library when the spec indicates that `is_builtin: False`

Notes:
* Some of the docker/docker compose changes need to get spun out and merged to `main` as this PR isn't intended for merge
* This is mounting one file, but a directory is also fine

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-225


krista-skylight (approved):
Left a non-blocking comment but something I think we should revisit down the line!

krista-skylight (approved):

emyl3 (approved):

PR #3075, opened 2026-03-10
build: fix up report execution container for test running

mcmcgrath13:
## Description

In completing #3074 , I found some of the container set up wasn't actually working fully. This PR fixes that:
* Add the mssql driver dependencies to the docker container
* Use host networking for the integration test container run so both the tests directly and the repot-execution container can talk to the DB using the same `localhost` connection string
* Add the `DATABASE_CONN_STRING` env var to the base compose and sample env
* Set up the docker compose in integration tests to use the sample env files
* Have the integration docker compose use the `--build` flag when setting up containers
* Add an integration test to prove it works


emyl3 (approved):
Thank you for fixing!!!

PR #3079, opened 2026-03-10
[APP-353] feat: convert nbs sr 5

mcmcgrath13:
## Description

Add SR 5 to python libraries.

Side quests:
* Add a test harness to replace a database table with fake data so the logic could be tested
* Only stop containers that the integration tests started (so that your database doesn't keep getting shut down on you)
* Rename `cursor` to `_cursor` in `Transaction` to make clear its "private"
* Change current `execute` to `query` and add `execute` that does not fetch results

image

- [x] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/report/execution/03_ODSE_Data_Report_Library_Init.sql`
  - [x] If the report execution service is not yet in use by any STLT (even beta), you should update this file
  - [x] Update the `name`, `runner`, `last_chg_time`, `last_chg_user_id`, and potentially the description (based on above)
- [ ] Check that the report has parity with the NBS 6 version using (process TBD)
  - [x] Improve the report spec or result expectations


apps/report-execution/src/libraries/nbs_sr_05.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Conversion notes:
+    * Export included the year as a column
+    * Export has columns in different order
+    * Use pipe separator instead of new line for subheader
+    """
+    today = datetime.date.today()
+
+    # date-based variables for use in the queries
+    curr_month = today.strftime('%b%Y').upper()
```
emyl3: ❓ Do we want to update in the queries below where we are calling `MONTH(CURRENT_TIMESTAMP)` or using the current_timestamp (for example) to use a variable instead? 🤔

apps/report-execution/src/libraries/nbs_sr_05.py
```diff
@@ -0,0 +1,175 @@
...
+    state_list = [
+        row[0] if row[0] is not None else 'N/A'
+        for row in states.data
+        if row[0] != filler_state
+    ]
+
+    trx.execute('DROP TABLE #base_data')
+
+    header = 'SR5: Cases of Reportable Diseases by State'
+    subheader = f'{", ".join(state_list)} | {today.strftime("%m/%d/%Y")}'
```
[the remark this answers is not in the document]
mcmcgrath13: nice catch! That's part of the description block, so adding there

emyl3 (approved):
Thanks for addressing all my questions! With the assumption that we will do functional testing later on and that we are OK with my limited SAS knowledge, I approve! 😅

mcmcgrath13:
@emyl3 re-requesting review as the guidance to match the export and not the run required non-trivial changes

apps/report-execution/src/libraries/nbs_sr_05.py
```diff
@@ -0,0 +1,188 @@
...
+        ')\n'
+        # Result select
+        'SELECT \n'
+        'FORMAT('
+        '  IIF('
+        '    median_ytd = 0, '
+        '    0, '
+        '    (curr_ytd - median_ytd) / median_ytd'
+        '  ), '
+        "  'P0', 'en-us'"
```
brick-green: (q, nb): Is zero places after the decimal the right precision?

brick-green (commented):
I wasn't able to run this via the front end. I also tried a few other reports and got this same error screen. I get the same screen if I try to Run or Export. Know what's up with this?

Image

emyl3 (approved):
🚢

brick-green (approved):

PR #3082, opened 2026-03-11
feat: break out description into header, subheader, description

mcmcgrath13:
## Description

Per discussion in standup, change report return format to have a `header`, `subheader` and `description` instead of just `description`


emyl3 (approved):
Looks good!

brick-green (approved):

PR #3092, opened 2026-03-19
APP-410 Convert NBSSR00009 Bar Graph of Selected Disease by Month

krista-skylight:
- [x] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/report/execution/03_ODSE_Data_Report_Library_Init.sql`
- [ ] Check that the report has parity with the NBS 6 version using (process TBD)


apps/report-execution/tests/integration/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,395 @@
...
+            assert row[col_index['cases']] >= 0
+
+            # Month code should be 6-digit YYYYMM format
+            month_code = row[col_index['month_code']]
+            assert len(month_code) == 6
+            assert month_code.isdigit()
+
+            # Month name should be 3 letters
+            month_name = row[col_index['month_name']]
+            assert len(month_name) in [3, 4]  # Some months like 'Sept' might be 4 chars
```
mcmcgrath13: (q, b): does SAS also have this "sometimtes 4" short name?

apps/report-execution/src/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,144 @@
...
+    """Standard Report 09: Monthly Cases by Disease and State.
+
+    'SR9: Bar Graph of Selected Disease by Month.  Report demonstrates, using a
+    vertical bar graph, the total number of monthly Investigation(s) [both
+    Individual and Summary] for a given disease, by State, irrespective of Case
+    Status.'
+
+    NOTE: In python, we are only returning the underlying data for the graph in a
+    tabular format, whereas the SAS version created the bar graph.
+
```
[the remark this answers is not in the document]
krista-skylight: TO DO: follow mary's format on changes

apps/report-execution/tests/integration/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,395 @@
...
+                'subset_query': 'SELECT * FROM [NBS_ODSE].[dbo].[PHCDemographic]',
+                'time_range': {'start': '2024-01-01', 'end': '2024-06-30'},
+            }
+        )
+
+        result = execute_report(report_spec)
+        assert result.content_type == 'table'
+
+        data = result.content.data
+        assert len(data) > 0
```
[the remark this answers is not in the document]
krista-skylight: the specific 6 month time range I used in this test gave me 52 rows so I put the assertion at >50 right now. will that always remain the same as long as `phc_demographic.yaml` has the same random seed? If so, can I assume we're always going to keep it the same?

apps/report-execution/tests/integration/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,395 @@
...
+    """Integration tests for the nbs_sr_09 library.
+
+    This library generates monthly case counts for bar graph data,
+    filtering by time range, state, and disease.
+    """
+
+    @pytest.fixture(autouse=True)
+    def set_time(self, time_machine):
+        time_machine.move_to(datetime.datetime(2024, 6, 24))
+
```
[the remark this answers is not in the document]
krista-skylight: yep, got rid of it

apps/report-execution/src/utils.py
```diff
@@ -32,3 +32,23 @@ def get_int_env_or_default(env_var: str, default: int):
...
+    col_index = {col: idx for idx, col in enumerate(content.columns)}
+
+    state_set = set()
+
+    for row in content.data:
+        state = row[col_index['State']]
+        if state is not None:
+            state_set.add(state)
+
+    state_list = sorted(state_set)
```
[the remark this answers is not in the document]
krista-skylight: Ok I pushed something that I think will work. I also added some functions to the `Table` object to make it easier to get columns by a column name which I call to help pass state columns from content to `gen_subheader`. Let me know what you think!

apps/report-execution/src/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,82 @@
...
+        FROM subset
+        WHERE event_date IS NOT NULL
+        GROUP BY 
+            state_cd,
+            state,
+            county,
+            phc_code_short_desc,
+            FORMAT(event_date, 'MMM'),
+            FORMAT(event_date, 'yyyyMM')
+        HAVING SUM(group_case_cnt) > 0
```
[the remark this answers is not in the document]
krista-skylight: Good catch on the corresponding test. I'll modify so that it checks for positives instead. I double checked and I think SAS automatically excludes values in a group by where sums are missing. SQL, however, will include it and put a null value in the groupby, but the `HAVING` clause takes care of that.

apps/modernization-api/src/main/resources/db/report/execution/libraries/nbs_sr_09.sql
```diff
@@ -0,0 +1,43 @@
...
+
+IF EXISTS (SELECT * FROM [dbo].[Report_Library] WHERE UPPER(library_name) = @sasLib)
+BEGIN
+    UPDATE [dbo].[Report_Library]
+    SET
+        library_name = @pyLib,
+        runner = 'python',
+        desc_txt = @desc,
+        last_chg_time = CURRENT_TIMESTAMP,
+        last_chg_user_id = 99999999
```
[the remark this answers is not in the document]
JordanGuinn: Posting for posterity's sake - @mcmcgrath13 & I chatted about this offline, and landed on leaving as-is, and not declaring a `SYSTEM_USER_ID` variable or anything, given the generally magic nature of the `99999999` paradigm.  Feel free to disregard my initial comment!

apps/modernization-api/src/main/resources/db/report/execution/03_ODSE_Data_Report_Library_Init.sql
```diff
@@ -32,7 +32,7 @@ BEGIN
         ('NBSSR00006.SAS', 'Cases of Selected Diseases by County'),
         ('NBSSR00007.SAS', 'SR7: Cases of Selected Diseases vs. 5-Year Median for Selected Time Period. Report demonstrates, in table form, Investigation(s) [both Individual and Summary] by year-to-date, and 5-year median irrespective of Case Status'),
         ('NBSSR00008.SAS', 'SR8: State Map Report of Disease Cases Over Selected Time Period. Report demonstrates, using a choropleth map, the total number of Investigation(s) [both Individual and Summary] by County irrespective of Case Status.'),
-        ('NBSSR00009.SAS', 'SR9: Bar Graph of Selected Disease by Month.  Report demonstrates, using a vertical bar graph, the total number of monthly Investigation(s) [both Individual and Summary] for a given disease, by State, irrespective of Case Status.'),
+        ('NBSSR00009.SAS', 'SR9: Selected Disease by Month and County.  Report demonstrates the total number of monthly Investigation(s) [both Individual and Summary] for a given disease and State, by County, irrespective of Case Status.''),
```
mcmcgrath13:
```suggestion
        ('NBSSR00009.SAS', 'SR9: Selected Disease by Month and County.  Report demonstrates the total number of monthly Investigation(s) [both Individual and Summary] for a given disease and State, by County, irrespective of Case Status.'),
```

apps/report-execution/src/models.py
```diff
@@ -31,6 +31,19 @@ class Table(BaseModel):
...
+        if col_name not in self.columns:
+            raise ValueError(
+                f"Column '{col_name}' not found. Available columns: {self.columns}"
+            )
+        idx = self.columns.index(col_name)
+        return [row[idx] for row in self.data]
+
+    def get_unique_column(self, col_name: str) -> list[Any]:
+        """Extract unique values from a column."""
+        return list(set(self.get_column(col_name)))
```
mcmcgrath13: 🚀  can we sort this as well 🙏

krista-skylight: yes, just pushed a change. it's a little weird though since python can't sort strings with Nones.

.github/workflows/Regression-tests.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
           CYPRESS_BASE_URL: 'http://localhost:8000/'
           CYPRESS_LOGIN_USERNAME: 'superuser'
           CYPRESS_LOGIN_PASSWORD: ''
-
+
```
JordanGuinn:
```suggestion

```

apps/report-execution/src/utils.py
```diff
@@ -32,3 +35,66 @@ def get_int_env_or_default(env_var: str, default: int):
...
+    """Parse a date string, trying ISO format first, then US format."""
+    try:
+        # Try ISO format (YYYY-MM-DD)
+        return datetime.fromisoformat(date_str)
+    except ValueError:
+        # Try US format (MM/DD/YYYY)
+        return datetime.strptime(date_str, '%m/%d/%Y')
+
+
+def gen_subheader(
```
JordanGuinn, line 51: 🔥 👏

JordanGuinn (approved):
:shipit: 🚀

.github/workflows/Regression-tests.yaml
```diff
@@ -43,7 +43,6 @@ jobs:
           CYPRESS_BASE_URL: 'http://localhost:8000/'
           CYPRESS_LOGIN_USERNAME: 'superuser'
           CYPRESS_LOGIN_PASSWORD: ''
-
       # Create cypress html report
```
mcmcgrath13, line 47: (nit, b-from-merge-but-notapprove): remove this change from the diff to keep things tidy

apps/report-execution/src/libraries/nbs_sr_05.py
```diff
@@ -141,16 +142,12 @@ def execute(
 
     # Get the state(s) in the data set for subheader display
     states = trx.query('SELECT DISTINCT state FROM #base_data ORDER BY state')
-    state_list = [
-        row[0] if row[0] is not None else 'N/A'
-        for row in states.data
-        if row[0] != filler_state
-    ]
+    state_list = states.get_unique_column('state')
```
mcmcgrath13, line 145: ❤️

apps/report-execution/src/utils.py
```diff
@@ -32,3 +35,66 @@ def get_int_env_or_default(env_var: str, default: int):
...
+
+    # Add states if provided
+    if states:
+        has_none = any(s is None for s in states)
+        clean_states = {s for s in states if s and s != '<FILLER>'}
+        if clean_states:
+            if has_none:
+                sorted_states = ['N/A'] + sorted(clean_states)
+            else:
+                sorted_states = sorted(clean_states)
```
mcmcgrath13: (sugg, b): given the general pattern is to call `get_unique_column` to get the values passed here, can we skip all this cleaning (same with disease) and just do the joining and appending?

apps/report-execution/src/utils.py
```diff
@@ -32,3 +35,66 @@ def get_int_env_or_default(env_var: str, default: int):
...
+            parts.append(', '.join(sorted(clean_diseases)))
+
+    # Add date range if time_range provided
+    if time_range:
+        # Parse dates (handles both ISO and US formats)
+        start_dt = parse_date(time_range.start)
+        end_dt = parse_date(time_range.end)
+        # Format as MM/DD/YYYY
+        parts.append(
+            f'{start_dt.strftime("%m/%d/%Y")} to {end_dt.strftime("%m/%d/%Y")}'
```
mcmcgrath13: (sugg, nb): create a lil date formatting helper and/or pull out the format to a variable so we don't repeat this string 3x

apps/report-execution/tests/integration/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,357 @@
...
+                'subset_query': 'SELECT * FROM [NBS_ODSE].[dbo].[PHCDemographic]',
+                'time_range': {'start': '2024-01-01', 'end': '2024-06-30'},
+            }
+        )
+
+        result = execute_report(report_spec)
+
+        # Verify County column exists and has values
+        col_index = {col: idx for idx, col in enumerate(result.content.columns)}
+        assert 'County' in col_index
```
mcmcgrath13: (sugg, nb): since this is the same query as the first test, can these checks be folded in there?

apps/report-execution/src/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,83 @@
...
+        ORDER BY 
+            phc_code_short_desc,
```
mcmcgrath13, line 49: (comment, b): The ordering here looks off from what I see in the export, they should match. It seems like the sas `by group` also does ordering so this should be the same as the `GROUP BY`

apps/report-execution/src/models.py
```diff
@@ -31,6 +31,33 @@ class Table(BaseModel):
...
+
+        Returns:
+            Sorted list of unique values with None placed first
+        """
+        if col_name not in self.columns:
+            raise ValueError(
+                f"Column '{col_name}' not found. Available columns: {self.columns}"
+            )
+        idx = self.columns.index(col_name)
+        values = {row[idx] for row in self.data}
```
```suggestion
        values = set(self.get_column(col_name))
```

apps/report-execution/tests/integration/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,357 @@
...
+                'data_source_name': '[NBS_ODSE].[dbo].[PHCDemographic]',
+                'subset_query': (
+                    'SELECT * FROM [NBS_ODSE].[dbo].[PHCDemographic] '
+                    "WHERE State = 'Tennessee' "
+                    "AND phc_code_short_desc = 'Measles' "
+                    "AND event_date >= '2024-04-01' AND event_date <= '2024-04-30'"
+                ),
+                'time_range': {'start': '2024-04-01', 'end': '2024-04-30'},
+            }
+        )
```
mcmcgrath13, line 120: (sugg/q, nb): these `_with_filter` tests seem like they're testing the subset sql and not login in the library itself. Is there something these cover that the first test doesn't? Can we remove them?

apps/report-execution/tests/utils.py
```diff
@@ -54,3 +57,23 @@ def test_get_int_env_or_default_not_set(self, monkeypatch, caplog):
             res = utils.get_int_env_or_default('REPORT_MAX_ROW_LIMIT_RUN', 100)
             assert res == 100
             assert len(caplog.records) == 0
+
+    def test_gen_subheader_with_single_date(self):
```
mcmcgrath13, line 61: (sugg, b): could you add a test when no arguments are passed (should return an empty string I'd guess?)

mcmcgrath13: or for where states/diseases are empty lists

package-lock.json
mcmcgrath13, line 1: remove this file

apps/report-execution/src/utils.py
krista-skylight: We could get rid of sorting since I put that in `get_unique_column` based on your earlier feedback, but the task of replacing `None` with `<N/A>` and getting rid of `<FILLER>` is only addressed here

apps/report-execution/src/utils.py
krista-skylight: Good idea. I'm also thinking I need to add in some checks for year only dates after having reviewed sr11.

mcmcgrath13: Yeah, looks like @JordanGuinn adapted the function to have the date format be an optional kw arg - might be good to pre-adopt here?

apps/report-execution/src/utils.py
mcmcgrath13: the `<FILLER>` is unique to sr_05, so I'd deal with filtering those out in that library and not have it be part of this method, but the N/A'ing likely makes sense to stay here - could be simplified to `['N/A/ if s is None else s for s in states]` or so

apps/report-execution/src/libraries/nbs_sr_09.py
```diff
@@ -0,0 +1,84 @@
...
+            FORMAT(event_date, 'MMM'),
+            FORMAT(event_date, 'yyyyMM')
+        HAVING SUM(group_case_cnt) > 0
+        ORDER BY 
+            state_cd,
+            state,
+            county,
+            phc_code_short_desc,
+            ord;
```
mcmcgrath13:
```suggestion
            monyr,
            ord;
```

apps/report-execution/tests/integration/libraries/nbs_sr_09.py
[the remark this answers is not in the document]
mcmcgrath13: The plan is that it should always use the same seed, so the data should be steady. Package udgrades and adding new columns can mess with this, but it can generally be assumed to be a relatively constant dataset

apps/report-execution/tests/utils.py
```diff
@@ -54,3 +57,27 @@ def test_get_int_env_or_default_not_set(self, monkeypatch, caplog):
...
+        assert result == '06/24/2024'
+
+    def test_gen_subheader_with_us_date_range(self):
+        time_range = TimeRange(start='01/01/2024', end='12/31/2024')
+        result = utils.gen_subheader(
+            states=['Georgia', 'Tennessee'], time_range=time_range
+        )
+        assert result == 'Georgia, Tennessee | 01/01/2024 to 12/31/2024'
+
+    def test_gen_subheader_with_states_diseases_and_date_range(self):
```
mcmcgrath13, line 72: (sugg, nb): add a test for the year scenario?

mcmcgrath13 (approved):
a few small comments left to address pre-merge, but looks good overall!

PR #3100, opened 2026-03-24
[APP-413] feat: conditional routing to report run ui

mcmcgrath13:
## Description

Add gateway logic to look for `POST` to `/nbs/nfc` for report run (in the sense of opening the run page, not execution)

Notes:
* Added a bit more logging by default to the service

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-413


krista-skylight (approved):
I'm not sure I have expertise to provide a ton of useful feedback! If I was really nitpicking then maybe declare `"sas"` and `"python"` as constants somewhere instead of hardcoding them, but I don't think it's a huge deal. Otherwise lgtm!

emyl3 (approved):
Works as intended for me!! nice work, Mary!! 🔍

apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportExecutionRouteLocatorConfiguration.java
```diff
@@ -0,0 +1,179 @@
...
+              .build()
+              .expand(getParamsFromBody(exchange))
+              .toUriString();
+
+      return WebClient.create()
+          .get()
+          .uri(path)
+          .accept(MediaType.APPLICATION_JSON)
+          .cookies(
+              // Copy the cookies from the original request onto the new request to make sure auth
```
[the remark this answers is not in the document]
JordanGuinn: Maybe worth following up on at some point later on?  I imagine the auth-related cookie is the most sensitive one anyway, so I don't imagine it makes a huge difference from a security perspective, but generally would be nice to not toss 'em all on there.  Def don't think it's an immediate blocker here though!

apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportExecutionRouteLocatorConfiguration.java
```diff
@@ -0,0 +1,187 @@
...
+                    .order(Ordered.HIGHEST_PRECEDENCE)
+                    .method(HttpMethod.POST)
+                    .and()
+                    .path("/nbs/nfc")
+                    .and()
+                    .readBody(LinkedMultiValueMap.class, bodyPredicate())
+                    .and()
+                    .asyncPredicate(isModPredicate(modService))
+                    .filters(filter -> filter.filter(this::redirectToMod))
+                    .uri("no://op"))
```
JordanGuinn: (q, nb): Very new to the `RouteLocatorBuilder` pattern - does the addition of `.uri("no://op"))` here effectively prevent any further manipulation to the route?

JordanGuinn (approved):
📈 👏

apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportExecutionRouteLocatorConfiguration.java
mcmcgrath13: I saw this in other locators that redirect so did the same. I think the `.uri` is maybe required to finish off the route, but we don't actually use it for redirects, hence the place holder

mcmcgrath13: confirmed `.uri` is needed to appease the types

PR #3103, opened 2026-03-24
feat: don't redirect if the api call asked for json

mcmcgrath13:
## Description

It has been a source of dev/application confusion that api requests for JSON data return a 302 that if you follow turns into a 200 (with an empty html page). It will be simpler and more expected for these cases to instead return a 403 that will allow standard error handling and messaging when issues do arise with these calls


emyl3 (approved):
Yay! This works as expected for me!

brick-green (approved):
Nice!

PR #3105, opened 2026-03-25
[APP-345] feat: sas, scriptifying test harness for use outside of tests

mcmcgrath13:
## Description

* Make test harness for loading fake data also available via script
* Add SAS container to build options
* Simplify API for using the table faker harness

Notes:
* Purposefully keeping the details on the SAS container in NEDSS Dev as it is private and you need access there to use it
* Needed for [these instructions on comparison testing](https://cdc-nbs.atlassian.net/wiki/x/CwBmjQ) to work

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-384


apps/report-execution/scripts/load_fake_data.py
JordanGuinn, line 1: (thought, b): Are there any build parameters or env vars we can take advantage of here to ensure this and the `restore_original.data.py` can only ever be run in dev/pre-prod environments?

apps/report-execution/tests/conftest.py
```diff
@@ -181,13 +199,30 @@ def fake_db_table(request):
...
     conn_string = utils.get_env_or_error('DATABASE_CONN_STRING')
 
+    # swap out original data for fake data
+    insert_fake_data(conn_string, faker_sql, db_tables, fk_tables)
+
+    # avoid connection inside connection
+    yield
+
+    # restore the original data
+    restore_original_data(conn_string, db_tables, fk_tables)
```
JordanGuinn, line 215: TIL that the the`restore_original_data` call _after_ the `yield` indicates to pytest that it's teardown code!  For a sec I was confused as to why we were inserting fake data just to immediately truncate it 😅 🤦

JordanGuinn (approved):
One q about a potential env gate for the data scripts, but otherwise looks good to me! 🚀

apps/report-execution/scripts/load_fake_data.py
mcmcgrath13: These files (scripts or tests) aren't included in our built container, which I think is our best proxy for prod. Given a human would need to set up an env var for the connection to a prod database, then also pass an argument, it feels like it would need to be pretty intentional. And the good news is they could maybe quickly restore if they did this by accident?

I don't think there's any way to know from a DB itself if it's production or not 🤔

emyl3 (approved):
worked as expected for me!

apps/report-execution/scripts/load_fake_data.py
JordanGuinn: Alrighty.  Something to keep an eye on moving forward I'd think, should we ever expand usage of that env var, or need to incorporate any Python scripts into the build for whatever reason

PR #3118, opened 2026-03-30
build: configure asset dir to `static` to match prior behavior

mcmcgrath13:
## Description

Found in bug bash, there were strange mime type errors, but at least some of these seem to stem from things trying to authenticate when they did not need it. `vite` has a different default asset directory than `CRA`. This PR changes the default to instead use `static` as CRA did


emyl3 (approved):
Thank you!!

krista-skylight (approved):

PR #3131, opened 2026-04-06
APP-356 Convert POTNTL_DUP_INV_SUM (Potential Duplicate Investigations)

krista-skylight:
## Description

Please include a summary of the changes and any key information a reviewer may need.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-356?atlOrigin=eyJpIjoiN2VmMGFmNjNhNjE3NGFmYWE3ZGE4ZTZkOWEzNDVjYjQiLCJwIjoiaiJ9)

- [x] Check that the report has parity with the NBS 6 version using (process TBD)


apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,120 @@
+from src.db_transaction import Transaction
+from src.models import ReportResult
+
+
+def execute(
+    trx: Transaction,
+    subset_query: str,
+    data_source_name: str,
+    **kwargs,
```
mcmcgrath13, line 10:
```suggestion
    days_value: None | int,
    **kwargs,
```

apps/report-execution/src/execute_report.py
```diff
@@ -19,11 +19,14 @@ def execute_report(report_spec: models.ReportSpec):
...
     with db_transaction(conn_string) as trx:
-        result = library.execute(
-            trx,
-            subset_query=report_spec.subset_query,
-            data_source_name=report_spec.data_source_name,
-        )
+        execute_kwargs: dict[str, typing.Any] = {
+            'subset_query': report_spec.subset_query,
+            'data_source_name': report_spec.data_source_name,
+        }
```
mcmcgrath13: (sugg, maybe-b): can we instead add a regular `days_value=report_spec.days_value` to the original call? since the none value is a fine default, it seems like we end up in the same spot, but more directly

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,254 @@
...
+        assert int(days_str) > 0
+
+        data = result.content
+        assert len(data.data) >= 0
+
+        # Verify column structure
+        for col in self.expected_columns:
+            data.get_column(col)
+
+    def test_execute_report_with_small_days_value(self):
```
mcmcgrath13: what happens if a negative number of days is passed?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,254 @@
...
+                'days_value': 30,
+            }
+        )
+
+        result = execute_report(report_spec)
+        assert result.content_type == 'table'
+        assert result.subheader == 'Duplicate Investigations Time Frame: 30 Days'
+
+        # Just verify it runs and returns data
+        assert len(result.content.data) >= 0
```
mcmcgrath13: is this test functionally different than the first one?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,254 @@
...
+                'library_name': 'potntl_dup_inv_sum',
+                'data_source_name': '[RDB].[dbo].[INV_SUMM_DATAMART]',
+                'subset_query': 'SELECT * FROM [RDB].[dbo].[INV_SUMM_DATAMART]',
+                'days_value': 30,
+            }
+        )
+        result_30 = execute_report(spec_30)
+
+        # 3650 days should return more or equal rows than 30 days
+        assert len(result.content.data) >= len(result_30.content.data)
```
mcmcgrath13: I'd expect more always for that difference - do we ever see equal?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,254 @@
...
+                ),
+                'days_value': 365,
+            }
+        )
+
+        result = execute_report(report_spec)
+        assert result.content_type == 'table'
+
+        # Verify only selected diseases appear
+        disease_cds = result.content.get_column('Disease Code')
```
mcmcgrath13: this test seems to be testing the subset sql not really the library - remove?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: Then I believe no data is returned since, in my query, I filter for anything less than the `days_value`. Should I have the library throw an error instead?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: No, it's just testing a different `days_value`. delete?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: I think it could be equal if all potential duplicate investigations in the data are already less than 30 days.

apps/report-execution/src/execute_report.py
krista-skylight: Ok, I modified so all the lambda functions in the original call have `**kwargs` and removed this.

apps/report-execution/src/models.py
```diff
@@ -13,6 +13,7 @@ class ReportSpec(BaseModel):
     library_name: str = Field(min_length=1)
     data_source_name: str = Field(min_length=1)
     subset_query: str = Field(min_length=1)
+    days_value: int | None = None  # Specific to potntl_dup_inv_sum
```
JordanGuinn, line 16: (thought, nb): Probably not worth addressing at this moment, but if there end up being _multiple_ reports that require unique properties like this, maybe we end up sketching out a `custom_props` Object field or similar to capture them all?  Just in the spirit of minimizing the amount of library-specific bits on the `ReportSpec` model.

cdc-sandbox/docker-compose.nbs6.yml
```diff
@@ -32,6 +32,7 @@ services:
   sas:
     image: ghcr.io/cdcent/nbs7-sas-linux:v1.0.4
     platform: linux/amd64
+    container_name: "${COMPOSE_PROJECT_NAME}-sas"
```
JordanGuinn, line 35: (q, nb): Any particular reason for this addition?

apps/report-execution/tests/conftest.py
```diff
@@ -167,7 +168,7 @@ def get_tables_from_faker(schema_name: str) -> tuple[list[str], list[str]]:
         schema = yaml.safe_load(f.read())
 
     db_tables = [t['table_name'] for t in schema['tables']]
-    fk_tables = schema['config']['nbs']['fk_tables']
+    fk_tables = schema['config'].get('nbs', {}).get('fk_tables', [])
```
JordanGuinn, line 179: (q, nb): What's this change for?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,244 @@
...
+        'Case Status',
+        'Event Date',
+        'Event Date Type',
+        'MMWR Year',
+        'Notification Record Status',
+        'Disease Code',
+    ]
+
+    def test_execute_report_with_days_value(self, snapshot):
+        """Test with a specific days value (e.g., 365 days)."""
```
JordanGuinn: (q, nb): Can/should we do any actual date comparisons of the underlying results returned by the report lib, to ensure they actually match the days value provided?

JordanGuinn (commented):
This is looking good!! 🔥

apps/report-execution/tests/conftest.py
mcmcgrath13: so if there are no fk tables specified (not always relevant), then we default to empty list

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
mcmcgrath13: given the data is largely a fixed data set, is that true of our data? I'd expect not given the randomness at play

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
mcmcgrath13: yeah, let's 🪓

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
mcmcgrath13: the front end doesn't let you enter negatives, so I don't think it's high risk in practice. We probably just want to make sure it doesn't explode in an opaque way and if it does make it error more clearly - add a test for this case?

apps/report-execution/src/models.py
krista-skylight: Yea def a good idea to revisit as we work through more translations!

cdc-sandbox/docker-compose.nbs6.yml
krista-skylight: I do not know enough to understand why but when I was troubleshooting issues with @mcmcgrath13 doing sas translations, she suggested making this change to get around an error message that I can no longer recall lol. 

Mary, do you think I should push this for everyone or leave it as something only I modify?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: That's not true of our data. I can modify to `>`, but add a little comment about this.

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: No, unfortunately, the report doesn't retain data about the two event dates for the potential duplicate or how close in days they are to each other. That's why I did a comparison between 3650 and 30 days to check that 3650 has more potential dupilcates than 30.

JordanGuinn (approved):
🚀

cdc-sandbox/docker-compose.nbs6.yml
mcmcgrath13: I think it's fine to push - makes working with the sas container easier via compose

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,261 @@
...
+      
+      - column_name: DIAGNOSIS_DATE
+        type: string
+        data: None
+        null_percentage: 1.0
+      
+      - column_name: ILLNESS_ONSET_DATE
+        type: string
+        data: None
+        null_percentage: 1.0
```
mcmcgrath13: (q, nb): why include the column, but have everything be null?

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,261 @@
...
+        type: string
+        data: random.choice(["COMPLETED", "APPROVED", "REJECTED", None])
+        null_percentage: 0.15
+      
+      - column_name: EVENT_DATE
+        type: string
+        data: f"{fake.date_between(start_date='-1y', end_date='today').strftime('%Y-%m-%d')} 12:00:00.000"
+        null_percentage: 0.05
+      
+      - column_name: EVENT_DATE_TYPE
```
mcmcgrath13, line 167: should this be linked off of the non-null ness of `EVENT_DATE`? (similar to how in phdc the state is dependent on state cd)

apps/report-execution/tests/conftest.py
```diff
@@ -153,6 +153,7 @@ def get_faker_sql(schema_name: str) -> str:
 
     # KLUDGE: NULL writing is not always correct
     result = result.replace(' nan,', ' NULL,')
+    result = result.replace('nan', ' NULL')
```
mcmcgrath13: (q, nb): why did we need to add this one? there's a risk that a valid part of a string with `nan` in it ill now be turned into ` NULL` is it the opening paren case of `(nan,`?

apps/report-execution/tests/conftest.py
```diff
@@ -234,6 +235,7 @@ def insert_fake_data(
     with db_transaction(conn_string) as trx:
         # Tables with foreign keys pointing to the table we want to replace need to
         # be backed up and cleared out to avoid FK constraint violations
+
```
mcmcgrath13:
```suggestion
```

mcmcgrath13:
I only see the SAS output and not the python in the report catalog - can you update with the python?

Also, could you update the spreadsheet tracker and add the e2e ticket? 

image

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: I was just trying to mimic everything I saw in the original table as much as possible since I wasn't sure what downstream effects empty columns have on this library or other ones using this table

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: oooh great point! I just modified accordingly

apps/report-execution/tests/conftest.py
krista-skylight: I remember adding some of these to get `get_faker_sql` to not break for my data but now that I'm trying it again without that line, it still does work, so I have removed it!

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: @mcmcgrath13 for some reason with this change and a new snapshot is causing tests to fail in CI but pass locally :cry:. Let me know if you have additional feedback. I'm going to work on fixing the CI issue before merging.

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
mcmcgrath13: for other fake tables, we've been including the columns that are definitely used and/or required, but otherwise leaving things off. I think it's fine to include more, but if a column is included it should have reasonable values, whereas this one is in neither state

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,270 @@
...
+          (
+            lambda ed: None if ed is None else random.choice([
+              "Investigation Start Date", 
+              "Date of Report", 
+              "Specimen Collection Date of Earliest Associated Lab", 
+              "Illness Onset Date", 
+              "Date of Diagnosis"
+            ])
+          )(EVENT_DATE)
+        null_percentage: 0
```
mcmcgrath13: table faker nullifies after generating the data, so for this to do what you want, you need to not use `null_percentage` in the `EVENT_DATE` spec and instead do something like `blah if random.random() < .05 else blah`

mcmcgrath13:
Looking at the comparison test files, the date formatting looks very different - is this expected?

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: also get rid of `today` in `EVENT_DATE`

krista-skylight: need to mock time instead

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,255 @@
...
+        null_percentage: 1.0
+      
+      - column_name: NOTIFICATION_STATUS
+        type: string
+        data: random.choice(["COMPLETED", "APPROVED", "REJECTED", None])
+        null_percentage: 0.15
+      
+      - column_name: EVENT_DATE
+        type: string
+        data: f"{fake.date_between(start_date='-1y', end_date='today').strftime('%Y-%m-%d')} 12:00:00.000" if random.random() > 0.05 else None
```
mcmcgrath13: (nit, b): need to change these dates to use a set time frame, otherwise this is going to fail CI tomorrow (then likely a time needs to be frozen in the tests, similar to nbs_sr_05)

krista-skylight: thanks for catching that, just modified!

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
mcmcgrath13: in your integration test file, you likely want to freeze time at a specific date as well (like is done in sr_05)

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,255 @@
...
+        null_percentage: 0.05
+      
+      - column_name: PATIENT_LAST_NAME
+        type: string
+        data: fake.last_name()
+        null_percentage: 0.05
+      
+      - column_name: PATIENT_DOB
+        type: string
+        data: f"{fake.date_of_birth(minimum_age=0, maximum_age=100).strftime('%Y-%m-%d')} 00:00:00.000"
```
mcmcgrath13: this likely should be changed to a frozen range of specific dates as well - this is implicitly "today"-based

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
mcmcgrath13: looking at the sql again, it doesn't seem like today's date matters at all, so freezing test time shouldn't be necessary (unless I'm missing some use of current time). Could be the birth date thing I pointed out above

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: good catch, just pushed a change!

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,254 @@
...
+      
+      - column_name: PATIENT_CURRENT_SEX
+        type: string
+        data: random.choice(["Male", "Female", "Unknown", None])
+        null_percentage: 0.10
+      
+      - column_name: AGE_REPORTED
+        type: int
+        data: fake.random_int(0, 100)
+        null_percentage: 0.15
```
mcmcgrath13, line 56:
```suggestion
        data: fake.random_int(0, 100) if random.random() < .85 else None
```

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,254 @@
...
+
+      - column_name: PATIENT_DOB
+        type: string
+        data: None if AGE_REPORTED is None else (datetime(2024, 6, 24) - timedelta(days=AGE_REPORTED * 365)).strftime('%Y-%m-%d')
```
mcmcgrath13: (nit, nb): for this conditional null to work as it looks like intended, the `AGE_REPORTED` column needs to change to the suggestion above

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -0,0 +1,254 @@
...
+        null_percentage: 0.15
+      
+      - column_name: EVENT_DATE
+        type: string
+        data: f"{fake.date_between(start_date=datetime(2010, 1, 1), end_date=datetime(2024, 6, 24)).strftime('%Y-%m-%d')}" if random.random() > 0.05 else None
+      
+      - column_name: EVENT_DATE_TYPE
+        type: string
+        data: None if EVENT_DATE is None else random.choice(["Investigation Start Date", "Date of Report", "Specimen Collection Date of Earliest Associated Lab", "Illness Onset Date", "Date of Diagnosis"])
+        null_percentage: 0
```
mcmcgrath13:
```suggestion
```

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -0,0 +1,230 @@
...
+class TestIntegrationNbsSrDupInvLibrary:
+    """Integration tests for the Potential Duplicate Investigations library.
+
+    This report identifies potential duplicate investigations for the same patient
+    with the same disease within a user-specified number of days.
+    """
+
+    @pytest.fixture(autouse=True)
+    def set_time(self, time_machine):
+        time_machine.move_to(datetime(2024, 6, 24))
```
mcmcgrath13: (q, nb): why was this needed in the end?

mcmcgrath13 (approved):
a couple minor suggestions/questions, but overall looks good!

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: You know what, I don't think it's needed since I didn't end up using any references to `today` in my tablefaker yaml and ended up hardcoding dates to 6-24-2024. that said, maybe it'd make more sense to not hardcode that and rely on the `time_machine` fixture?

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: @mcmcgrath13 I reverted back to the `null_percentage` because tablefaker uses pandas datatypes somewhere in its stack trace and pandas integers are not nullable so this will fail. I think that's why it applies the null percentage at the end. In this case, I can just have `AGE_REPORTED` and `PATIENT_DOB` be nullified separately since that's probably realistic anyways. Are you ok with that solution for now? If we come across a situation where one field has to follow the nulls of another field that is an integer, I'll put in the effort to find a workaround.

mcmcgrath13: yup! it doesn't seem like this is critical to correctness. The `PATIENT_DOB` should probably also be changed to use null percentage with a comment on why since `AGE_REPORTED` will never be None in its logic (I think 🤔 )

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
mcmcgrath13: The table faker dates have to be hard coded for stability of tests. I think you can remove the time machine usage in this test suite

(table faker runs before time machine, so it's seeing real dates/times, not the frozen ones)

PR #3139, opened 2026-04-07
build: bump graphql codegen deps

mcmcgrath13:
## Description

Upgrade graphql-codegen deps, npm was pointing to them as the root of vulnerabilities


emyl3 (approved):
🚢 🔒

PR #3149, opened 2026-04-10
build: rm kafka from mod repo as it isn't used

mcmcgrath13:
## Description

Remove kafka dev set up as it isn't used in this repo at all


JordanGuinn (approved):
🪓

brick-green (commented):
did a global search for "kafka" and see other references. Does everything need to be yanked out?

mcmcgrath13:
> did a global search for "kafka" and see other references. Does everything need to be yanked out?

What are you seeing? I see the below and the one non-git reference I see left is talking about how they decided not to use kafka (which seems like useful history to retain)

image

brick-green:
@mcmcgrath13 took another look and my global search was picking up files in the python venv. I excluded those and that cleared up the list. Except for one more reference in moderinization-api/README.md:

```
### With Kafka Enabled

The Modernization API can connect to the Kafka instance defined in `cdc-sandbox/kafka` by starting the service with
`kafka.enabled` set to `true`.
```

brick-green (approved):
that one comment about the README is non-blocking ✅

mcmcgrath13:
> @mcmcgrath13 took another look and my global search was picking up files in the python venv. I excluded those and that cleared up the list. Except for one more reference in moderinization-api/README.md:
> 
> ```
> ### With Kafka Enabled
> 
> The Modernization API can connect to the Kafka instance defined in `cdc-sandbox/kafka` by starting the service with
> `kafka.enabled` set to `true`.
> ```

Looks like I'd deleted this, but not actually saved it, so search didn't show it, but it also wasn't committed 🤦 - fixed

PR #3160, opened 2026-04-13
[APP-367] feat: Update subset_sql to use basic filters from request

emyl3:
## Description

The PR takes basic filters from a ReportConfiguration then creates a WHERE clause and adds that to the subset_sql. The funcationality has been pulled from `buildWhereClause()` in `ReportSQLBuilder`. This only implements the basic filters.

## Tickets

* [APP-367](https://cdc-nbs.atlassian.net/jira/software/c/projects/APP/boards/626?assignee=712020%3A7c8ce8a3-9019-43e6-a33b-4e6d4e7b8237&selectedIssue=APP-367)

[APP-367]: https://cdc-nbs.atlassian.net/browse/APP-367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -0,0 +1,36 @@
...
+import java.time.YearMonth;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import org.springframework.stereotype.Component;
+
+@Component
+public class FieldFormatter {
+  public String formatField(String type, String value) {
+    return switch (type.toUpperCase()) {
+      case "STRING" -> "'" + value.replace("'", "''") + "'";
```
mcmcgrath13, line 15: do we need to do any other sql escaping?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.StringJoiner;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+/** Service to generate the SQL WHERE clause from a report configuration. */
+@Service
+@RequiredArgsConstructor
+public class WhereClauseService {
```
mcmcgrath13, line 21: (q, nb): why make this a service vs just a utility class that's imported?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,117 @@
...
+   */
+  public String buildBasicWhereClause(ReportConfiguration reportConfig) {
+    if (reportConfig.filters() == null || reportConfig.filters().isEmpty()) return "";
+
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(" AND ", "WHERE ", "");
+
+    for (FilterConfiguration filter : reportConfig.filters()) {
+      findColumn(reportConfig, filter)
+          .ifPresent(
```
mcmcgrath13: what about building filters that don't have columns? do all basic filters target a column?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,117 @@
...
+              column -> {
+                String segment = buildFilterFragment(filter, column);
+                if (!segment.isEmpty()) finalWhere.add(segment);
+              });
```
mcmcgrath13: (q, b): should there be an error if there's no column?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  /**
+   * Builds the SQL fragment (e.g., "(COL_NAME IN ('A', 'B') OR COL_NAME IS NULL)")
+   *
+   * @param filterConfig provides the filter values to the column with and the operators
+   * @param column provides the column name and type
+   * @return an SQL fragment or empty string
+   */
+  private String buildFilterFragment(FilterConfiguration filterConfig, ReportColumn column) {
```
mcmcgrath13: (issue, b): the filter fragments should come from the requests, not the configuration itself as we want to be responsive to the form. The default values are used to populate the starting point for the form

mcmcgrath13: we probably need to add an optional `isNullable` to the `BasicFilterRequest` record type

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,117 @@
...
+
+    // Safety: If no values and no nulls were processed, return empty to avoid "()"
+    if (!hasValues && !allowNulls) return "";
+
+    return segment.append(")").toString();
+  }
+
+  private boolean isSingleValue(FilterConfiguration filter) {
+    return Integer.valueOf(1).equals(filter.maxValueCnt())
+        && Integer.valueOf(1).equals(filter.minValueCnt());
```
mcmcgrath13: (thing to check, b): I think this logic may only work for the dropdown fields, the `BAS_TXT` has `2` as the max value count even though its a single value

mcmcgrath13 (changes requested):
Overall structure looks good, but need to fix up the data source for the filter building

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: Started as a util then I broke out FieldFormatter and wanted to inject that so I moved over to a Service. I also think it makes testing a little easier when both are injected.

Side note: I broke out FieldFormatter since my instinct told me there would be more to be added there. IDK if it's really necessary though.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: The current NBS6 code calls`getDataSourceColumnDT()` as one of the first steps. That method throws an exception if a column is not found. I was just working from that. Is that a safe route?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: yup, I'll change to throw an exception

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
mcmcgrath13: I think so. The advanced filter doesn't have a column, but I'm pretty sure every basic filter does and since this is for basic an error makes sense

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: Oh right, this was a weird part of the old code I wanted to point out. In NBS6 there are if-else check if minValue = maxValue. Then if true it only pulls the first filter value and applies it. I couldn't think what usecase that was meant to address. Plus I only saw minValue of 1 = maxValue of 1. So I thought it was better to be explicit about the values. 

Any ideas of why this functionality is useful?

mcmcgrath13: it's for knowing if a drop down is single select or multi-select so we know whether to do `in (list, of, things)` vs `= thing`, but I think it needs the context of the type of filter to interpret perhaps?

brick-green: note: dropped this method in favor of always doing `IN (value1)`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: Added this in :)

apps/modernization-ui/src/generated/models/FilterType.ts
```diff
@@ -7,8 +7,8 @@ export type FilterType = {
     codeTable?: string;
     descTxt?: string;
     code?: string;
-    filterCodeSetName?: string;
-    filterType?: string;
```
mcmcgrath13, line 11: this `filterType` field is definitely used in the code - can you update references? (for another PR - we should make sure our build fails if typescript doesn't typecheck probably?)

mcmcgrath13: https://github.com/CDCgov/NEDSS-Modernization/pull/3195

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdvancedFilterConfiguration.java
```diff
@@ -6,4 +6,6 @@ public record AdvancedFilterConfiguration(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long reportFilterUid,
     Long reportColumnUid,
     Expr defaultValue,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) FilterType filterType) {}
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) FilterType filterType,
+    Integer maxValueCnt,
+    Integer minValueCnt) {}
```
mcmcgrath13: (sugg, b): remove this. not relevant to advanced filter config

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterConfiguration.java
```diff
@@ -6,8 +6,9 @@
...
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long reportFilterUid,
     Long reportColumnUid,
-    List<String> defaultValue,
+    List<String> defaultValues,
     Integer minValueCount,
     Integer maxValueCount,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean isRequired,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) FilterType filterType) {}
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) FilterType filterType,
+    Boolean defaultIncludeNulls) {}
```
mcmcgrath13: (sugg, nb): can we make this required since it defaults to false in the mapper?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -0,0 +1,36 @@
...
+      case "DATE" -> convertToSQLDate(value);
+      case "INTEGER", "NUMBER" -> value;
+      default -> throw new IllegalArgumentException("Unexpected Column Type: " + type);
+    };
+  }
+
+  private static String convertToSQLDate(String date) {
+    if (date == null || date.isBlank()) return null;
```
mcmcgrath13: (q, nb): what's the scenario where date is blank or null? should this error instead of return null?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -0,0 +1,36 @@
...
+
+    LocalDate localDate;
+    try {
+      if (date.length() > 7) {
+        localDate = LocalDate.parse(date, DateTimeFormatter.ofPattern("MM/dd/yyyy"));
+      } else {
+        localDate = YearMonth.parse(date, DateTimeFormatter.ofPattern("MM/yyyy")).atDay(1);
```
mcmcgrath13: (comment, b): This will always make the mm/yyyy the first day of the month, but if it's the end of the range, it should be the last day of the month

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -0,0 +1,36 @@
...
+      }
+    } catch (DateTimeParseException e) {
+      throw new IllegalArgumentException("Can't Convert Date!");
```
mcmcgrath13: (q, b): do `yyyy` format dates also need to be handled in this blocK?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -77,7 +80,8 @@ public ReportSpec build() {
 
     String selectClause = buildSelectClause(columns);
     String fromClause = String.format("FROM %s", dataSourceName);
-    String whereClause = "";
+    String whereClause =
+        whereClauseService.buildBasicWhereClause(reportConfig, reportExecRequest.basicFilters());
```
mcmcgrath13:
```suggestion
        whereClauseService.buildWhereClause(reportConfig, reportExecRequest);
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+   *     columns and active filters.
+   * @return A formatted SQL string starting with WHERE. An empty string is returned if required
+   *     parts are empty
+   */
+  public String buildBasicWhereClause(
+      ReportConfiguration reportConfig, List<BasicFilterRequest> basicFilterRequests) {
+    if (basicFilterRequests == null || basicFilterRequests.isEmpty()) return "";
+
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(" AND ", "WHERE ", "");
```
mcmcgrath13: (sugg, nb): The `WHERE` should probably be handled in the overall where clause builder and not the basic one. should this joiner get lifted up to the parent method and passed in here?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+      ReportColumn column =
+          findColumn(reportConfig, config.reportColumnUid())
+              .orElseThrow(
+                  () ->
+                      new IllegalArgumentException(
+                          "No report column found for columnUid: " + config.reportColumnUid()));
+
+      // Build the SQL segment
+      String segment = buildBasicFilterFragment(filterRequest, column);
+      if (!segment.isEmpty()) {
```
mcmcgrath13: (q, nb): what's the case where a segment is validly empty? should this be an error state instead?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+        finalWhere.add(segment);
+      }
+    }
+
+    // Only return the WHERE clause if it contains anything beyond the initial "WHERE " prefix
+    return finalWhere.length() > 6 ? finalWhere.toString() : "";
```
mcmcgrath13: (sugg, nb): also lift to parent method

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+   */
+  private String buildBasicFilterFragment(
+      BasicFilterRequest basicFilterRequest, ReportColumn column) {
+
+    List<String> values = basicFilterRequest.values();
+    boolean includeNulls = basicFilterRequest.includeNulls();
+
+    // Guard clause: if there are no values, and we aren't including nulls, there is no filter to
+    // apply
+    if ((values == null || values.isEmpty()) && !includeNulls) return "";
```
mcmcgrath13: (q, nb): should we do a check here if this is a required filter and error if it isn't? (or are we assuming that validation is done elsewhere? (cc @JordanGuinn who is working on validation)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+
+    // Standardize the values list to avoid NullPointer exception on stream
+    List<String> rawValues = (values == null) ? List.of() : values;
```
mcmcgrath13: (q, nb): given we've already done a null check on line 109, do we still need this?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+
+    // Transform raw client values into type-formatted strings
+    List<String> formattedValues =
+        rawValues.stream()
+            .map(v -> fieldFormatter.formatField(column.columnSourceTypeCode(), v))
+            .toList();
+
+    StringBuilder segment = new StringBuilder("(");
+    String colName = "[" + column.columnName() + "]"; // Brackets protect against SQL reserved words
+    boolean hasValues = !formattedValues.isEmpty();
```
mcmcgrath13, line 149: (q, nb): when could this be empty given the input values is known to not be empty?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,143 @@
...
+
+    // Append the IN clause if actual data values were provided
+    if (hasValues) {
+      segment
+          .append(colName)
+          .append(" IN (")
+          .append(String.join(", ", formattedValues))
+          .append(")");
+    }
```
mcmcgrath13, line 158: (q, b): what about dates where this is a range?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
brick-green: added functionality for date ranges

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: lifted up to the buildWhereClause method

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,210 @@
...
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(SQL_AND, "WHERE ", "");
+
+    String basicWherefragment =
+        buildBasicWhereFragment(reportConfig, executionRequest.basicFilters());
+
+    finalWhere.add(basicWherefragment);
+
+    // Only return the WHERE clause if it contains anything beyond the initial "WHERE " prefix
+    return finalWhere.length() > 6 ? finalWhere.toString() : "";
```
JordanGuinn: (nit, nb): We could also check to see if `basicWherefragment` returns a non-empty value?  Not a huge deal by any means, just seems a bit precarious to check against the base `WHERE` clause length.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
JordanGuinn, line 1: (nit, nb): Thoughts on naming this `WhereClauseBuilder` or similar, in the spirit of following the pattern of `ReportSpecBuilder`?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -0,0 +1,71 @@
...
+
+  private static String convertToSQLDate(String date) {
+
+    LocalDate localDate;
+    try {
+      localDate = LocalDate.parse(date, DateTimeFormatter.ofPattern("MM/dd/yyyy"));
+    } catch (DateTimeParseException e) {
+      throw new IllegalArgumentException("Can't Convert Date: " + date);
+    }
+    // Produces '2023-12-31'
```
JordanGuinn: (q, nb): Is this true?  It's not the output I would expect with us parsing to `"MM/dd/yyyy"`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
JordanGuinn: Maybe worth throwing if there are somehow no formatted values?  Or if the number of formatted values somehow doesn't match the number of values on the request (which should never happen, but yannow)

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
```diff
@@ -0,0 +1,394 @@
...
+    // Setup Config
+    BasicFilterConfiguration config =
+        createBasicFilterConfiguration(List.of(), filterUid, columnUid, false, "BAS_CON_LIST");
+
+    ReportColumn reportColumn = mockReportColumn(columnUid, "STRING", "ColumnName");
+    ReportConfiguration reportConfig = createReportConfig(List.of(config), List.of(reportColumn));
+
+    // Malicious Input: Attempt to close the IN clause and drop the Reports table
+    // The -- at the end comments out the rest of your generated query (the closing parentheses)
+    String maliciousInput = "'); DROP TABLE Reports; --";
```
JordanGuinn, line 275: 👏 📈

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapper.java
```diff
@@ -30,8 +31,15 @@ public static BasicFilterConfiguration fromReportFilter(ReportFilter filter) {
     // For the future: A list of strings may end up being too simple for all use cases,
     // may need to evolve to be a small object with a key and value
     List<String> defaultValue = null;
```
JordanGuinn:
```suggestion
    List<String> defaultValues = null;
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterConfiguration.java
```diff
@@ -6,8 +6,9 @@
...
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long reportFilterUid,
     Long reportColumnUid,
-    List<String> defaultValue,
+    List<String> defaultValues,
     Integer minValueCount,
     Integer maxValueCount,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean isRequired,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) FilterType filterType) {}
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) FilterType filterType,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean defaultIncludeNulls) {}
```
mcmcgrath13: (nit, nb): the minorest of baby nits, but I'd probably put this field just after defaultValues for colocation of sort of thing reasons

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,210 @@
...
+
+    // Standardize the values list to avoid NullPointer exception on stream
+    List<String> rawValues = (values == null) ? List.of() : values;
+
+    // Transform raw client values into type-formatted strings
+    List<String> formattedValues = fieldFormatter.convertToSQLFromDateRange(rawValues);
+
+    StringBuilder criteria = new StringBuilder("(");
+
+    String colName = "[" + column.columnName() + "]";
```
mcmcgrath13: (sugg, nb): probably overkill, but do we want a utility somewhere for escaping/formatting column Names? We do it here and elsewhere in the query building

apps/modernization-api/src/test/java/gov/cdc/nbs/report/utils/FieldFormatterTest.java
```diff
@@ -0,0 +1,82 @@
...
+    assertThat(formatter.formatField("STRING", "O'Brian")).isEqualTo("'O''Brian'");
+
+    // Multiple quotes
+    assertThat(formatter.formatField("STRING", "''")).isEqualTo("''''''");
+  }
+
+  @Test
+  void should_return_numbers_as_is() {
+    assertThat(formatter.formatField("INTEGER", "123")).isEqualTo("123");
+    assertThat(formatter.formatField("NUMBER", "45.67")).isEqualTo("45.67");
```
mcmcgrath13, line 28: (sugg, b): for security, can we ensure that the numbers are number shaped before allowing them through?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/utils/FieldFormatterTest.java
```diff
@@ -0,0 +1,82 @@
...
+  void should_handle_date_ranges_with_month_and_year_defaults() {
+    // Test MM/yyyy Range
+    List<String> monthRange = List.of("02/2024", "02/2024");
+    List<String> monthResult = formatter.convertToSQLFromDateRange(monthRange);
+
+    assertThat(monthResult.get(0)).isEqualTo("'2024-02-01'");
+    assertThat(monthResult.get(1)).isEqualTo("'2024-02-29'"); // Leap year check
+
+    // Test yyyy Range
+    List<String> yearRange = List.of("2023", "2023");
```
mcmcgrath13:
```suggestion
    List<String> yearRange = List.of("2023", "2024");
```
to stress test leap years

apps/modernization-api/src/test/java/gov/cdc/nbs/report/utils/FieldFormatterTest.java
```diff
@@ -0,0 +1,82 @@
...
+  void should_be_case_insensitive_for_column_type() {
+    // Tests the .toUpperCase() logic in the switch expression
+    assertThat(formatter.formatField("string", "test")).isEqualTo("'test'");
+    assertThat(formatter.formatField("Date", "12/31/2023")).isEqualTo("'2023-12-31'");
+  }
+
+  @Test
+  void should_throw_exception_for_invalid_range_size() {
+    List<String> tooShort = List.of("2023");
+    assertThatThrownBy(() -> formatter.convertToSQLFromDateRange(tooShort))
```
mcmcgrath13, line 78: can we also test too long?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
```diff
@@ -0,0 +1,394 @@
...
+    ReportConfiguration reportConfig =
+        createReportConfig(basicFilterConfigs, List.of(reportColumn));
+
+    List<BasicFilterRequest> basicFilterRequest =
+        List.of(new BasicFilterRequest(filterUid, List.of(filterDefaultValue), false));
+
+    String whereFragment =
+        mockWhereClauseService.buildBasicWhereFragment(reportConfig, basicFilterRequest);
+
+    assertThat(whereFragment).isEqualTo("([ColumnName] IN (condition1))");
```
mcmcgrath13:
```suggestion
    assertThat(whereFragment).isEqualTo("([ColumnName] IN ('condition1'))");
```

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
```diff
@@ -0,0 +1,394 @@
...
+
+  private WhereClauseService mockWhereClauseService;
+
+  @BeforeEach
+  void setUp() {
+    mockWhereClauseService = new WhereClauseService(fieldFormatter);
+    // Default behavior for formatter: return value as is
+    Mockito.lenient()
+        .when(fieldFormatter.formatField(anyString(), anyString()))
+        .thenAnswer(ff -> ff.getArgument(1));
```
mcmcgrath13: (q, nb): why do we need to mock the where clause service? it's stateless so couldn't we just use it? My gut-code-smell still doesn't really grok why the where clause service is a service and not just a class that's normally imported and used where it needs using (and then we wouldn't mock it, but just use it) - can you eli5?

mcmcgrath13: same with the field formatter - seems like those both could be static classes

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -0,0 +1,71 @@
+package gov.cdc.nbs.report.utils;
+
+import java.time.LocalDate;
+import java.time.YearMonth;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
```
mcmcgrath13, line 11: (q, nb): why have this be a component instead of just a regular utility class?

apps/modernization-ui/src/generated/models/FilterType.ts
mcmcgrath13: it looks like the linting isn't catching things maybe because they are already optionally chained - can you do a manual check looking for the places to update - definitely at least in `BasicFilter.tsx`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: I think we left that require field checks will be a future validation ticket

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: throwing an exception here now for both the scenarios

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: my thought was a buildAdvancedWhereFragment() would be added next so that check on the `WHERE` length would cover for both. What do you think?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
brick-green: That's the [default SQL Server date format](https://learn.microsoft.com/en-us/sql/t-sql/data-types/date-transact-sql?view=sql-server-ver17). So it seemed like the best format to converge on for building the SQL criteria

apps/modernization-api/src/test/java/gov/cdc/nbs/report/utils/FieldFormatterTest.java
brick-green: good call! Added

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
JordanGuinn: Ahhh, sure that makes sense.  I suppose so long as we have tests covering both those cases, and the resulting where clause output, that should be fine!

apps/modernization-ui/src/generated/models/FilterType.ts
```diff
@@ -7,8 +7,8 @@ export type FilterType = {
...
-    filterName?: string;
+    codeSetName?: string;
+    type?: string;
```
mcmcgrath13, line 11: (comm, b): The rest of the UI needs to be updated to use these new names, same with `defaultValues`

mcmcgrath13: I think https://github.com/CDCgov/NEDSS-Modernization/pull/3217 should address why the tests are still passing even though there're a lot of off references

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterRequest.java
```diff
@@ -5,4 +5,5 @@
 
 public record BasicFilterRequest(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long reportFilterUid,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) List<String> values) {}
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) List<String> values,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean includeNulls) {}
```
mcmcgrath13: (thought, nb): as we get around to the front end side of this work, we might want to reconsider the required-ness here since not all filters allow nulls, so it may be simpler to allow missingness as false

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,225 @@
...
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(SQL_AND, "WHERE ", "");
+
+    String basicWhereFragment =
+        buildBasicWhereFragment(reportConfig, executionRequest.basicFilters());
+
+    finalWhere.add(basicWhereFragment);
+
+    // Only return the WHERE clause if it contains anything beyond the initial "WHERE " prefix
+    return finalWhere.length() > 6 ? finalWhere.toString() : "";
```
mcmcgrath13: (nit, nb): might be minor-ly more readable to pull out `WHERE ` to a constant and then change this to `finalWhere.equals(SQL_WHERE)` and flip the ternary branches so then it's a more explicit check vs implicit with the comment. Though maybe that doesn't work easily with string joiners?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * Builds an SQL date/time range criteria using a BETWEEN clause. Result format: (([COLUMN_NAME]
+   * BETWEEN 'Start' AND 'End') OR ([COLUMN_NAME] IS NULL))
+   *
+   * @param basicFilterRequest The user-provided date range values.
+   * @param column The metadata for the column being filtered.
+   * @return A parenthesized SQL fragment.
+   */
+  private String buildBasicTimeRangeCriteria(
```
mcmcgrath13, line 179: (sugg, nb): it seems like the logic here is almost the same as the standard one - would it make sense to combine and instead branch at the point of formatting the primary clause?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapperTest.java
```diff
@@ -76,7 +86,32 @@ void fromReportFilter_should_map_all_fields() {
 
     assertThat(mapped.reportFilterUid()).isEqualTo(reportFilter.getId());
     assertThat(mapped.reportColumnUid()).isEqualTo(column.getId());
-    assertThat(mapped.defaultValue()).isEqualTo(List.of("value"));
+    assertThat(mapped.defaultValues()).isEqualTo(List.of("value"));
+    assertThat(mapped.minValueCount()).isEqualTo(1);
```
mcmcgrath13, line 91: (sugg, b): can we add the `defaultIncludeNulls` to the assertion block?

mcmcgrath13: (and have a test where it will be false)

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
```diff
@@ -26,19 +30,46 @@
 @ExtendWith(MockitoExtension.class)
 class ReportSpecBuilderTest {
 
+  private WhereClauseService whereClauseService;
+
+  @BeforeEach
+  void setUp() {
+    // Instantiate real dependencies
+    FieldFormatter fieldFormatter = new FieldFormatter();
```
mcmcgrath13, line 38: ❤️

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
```diff
@@ -26,19 +30,46 @@
...
-    Mockito.lenient().when(basicFilterConfig.reportColumnUid()).thenReturn(columnId);
-
-    return basicFilterConfig;
+  private BasicFilterConfiguration mockBasicFilterConfiguration(
+      List<String> filterDefaultValues,
+      Long reportFilterUid,
+      Long reportColumnUid,
+      Boolean defaultIncludeNulls,
+      String type) {
+    BasicFilterConfiguration basicFilterConfiguration =
```
mcmcgrath13: (q, nb): why mock here vs use the real filter configuration? it's a relatively straightforward object

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
```diff
@@ -0,0 +1,386 @@
...
+    ReportConfiguration reportConfig =
+        createReportConfig(basicFilterConfigs, List.of(reportColumn));
+
+    List<BasicFilterRequest> basicFilterRequest =
+        List.of(new BasicFilterRequest(filterUid, List.of(filterDefaultValue), true));
+
+    String whereFragment =
+        mockWhereClauseService.buildBasicWhereFragment(reportConfig, basicFilterRequest);
+
+    assertThat(whereFragment).isEqualTo("([ColumnName] IN ('condition1') OR [ColumnName] IS NULL)");
```
mcmcgrath13: (sugg, b): could we add a test that combines this with the AND-ing to ensure overall precedence/parens end up correct?

brick-green: good call - added

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
brick-green: got a little mock happy on the work. Switched to instantiated config

apps/modernization-api/src/test/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapperTest.java
brick-green: added for true and false

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: I like this. I already used a const for the AND so another for WHERE feels right. Added

apps/modernization-ui/src/generated/models/FilterType.ts
brick-green: think I got all these fixed

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterConfiguration.java
```diff
@@ -6,7 +6,8 @@
 public record BasicFilterConfiguration(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long reportFilterUid,
     Long reportColumnUid,
-    List<String> defaultValue,
+    List<String> defaultValues,
+    Boolean defaultIncludeNulls,
```
mcmcgrath13:
```suggestion
    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean defaultIncludeNulls,
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -0,0 +1,226 @@
...
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+
+    String basicWhereFragment =
+        buildBasicWhereFragment(reportConfig, executionRequest.basicFilters());
+
+    finalWhere.add(basicWhereFragment);
+
+    // Only return the WHERE clause if it contains anything beyond the initial "WHERE " prefix
+    return finalWhere.toString().equals(SQL_WHERE) ? "" : finalWhere.toString();
```
mcmcgrath13: (nit/q, nb): is there any cost to stringifying the finalWhere twice? if so, pull that out into a var

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
```diff
@@ -84,11 +101,15 @@ private ReportExecutionRequest mockReportExecutionRequest(List<Long> columnUids)
 
   @Test
   void build_should_set_all_fields_correctly() {
+    List<String> filterDefaultValue = List.of("condition1");
```
mcmcgrath13:
```suggestion
    List<String> filterDefaultValues = List.of("condition1");
```

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
```diff
@@ -84,11 +101,15 @@ private ReportExecutionRequest mockReportExecutionRequest(List<Long> columnUids)
...
+
     Long columnUid1 = 1L;
     Long columnUid2 = 2L;
 
-    BasicFilterConfiguration filterConfig1 = mockBasicFilterConfiguration(columnUid1);
-    BasicFilterConfiguration filterConfig2 = mockBasicFilterConfiguration(columnUid2);
+    BasicFilterConfiguration filterConfig1 =
+        mockBasicFilterConfiguration(filterDefaultValue, 1L, columnUid1);
+    BasicFilterConfiguration filterConfig2 =
+        mockBasicFilterConfiguration(filterDefaultValue, 1L, columnUid1);
```
mcmcgrath13:
```suggestion
        mockBasicFilterConfiguration(filterDefaultValues, 1L, columnUid1);
    BasicFilterConfiguration filterConfig2 =
        mockBasicFilterConfiguration(filterDefaultValues, 1L, columnUid1);
```

apps/modernization-ui/src/generated/models/FilterType.ts
mcmcgrath13: I'm seeing one squiggle on `ReportRunPage.tsx`

Image

Though I think the easiest way to deal with this is not make `includeNulls` required on the `BasicFilterRequest`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterRequest.java
```diff
@@ -7,4 +7,5 @@
 
 public record BasicFilterRequest(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @Positive Long reportFilterUid,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull List<String> values) {}
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull List<String> values,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean includeNulls) {}
```
mcmcgrath13:
```suggestion
    Boolean includeNulls) {}
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: The cost is pretty light but I think I found a decent solution

mcmcgrath13 (approved):
🚀

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
JordanGuinn: @brick-green Did this ever get answered/resolved?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
JordanGuinn: @brick-green Following up on Mary's q - any other spots that need to be covered?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
JordanGuinn: @brick-green ping

JordanGuinn (approved):
Looks good overall, just waiting on resolution from a few minor q's ✅

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
brick-green: I don't see a right or wrong on this one. When I originally wrote it I went with component so that mocking in the tests would be simpler. Didn't end up doing that though so the spring management isn't providing much value here. But I don't see it hurting either and I liked leaning into spring so I left it. 🤷

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
brick-green: I can't think of anything else needed here now wince we really only using `IN` statements. Might need to think about `%` if `LIKE` is introduced in the future

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: Oh right, I decided to leave this since it didn't feel like this fit into the builder pattern. But now I wish I didn't use the term "build" in the method names. That does kind of muddy the waters a bit

PR #3152, opened 2026-04-10
APP-379, Take Three - Address outstanding Sonar code smells

JordanGuinn:
## Description

I thought I had addressed the outstanding Sonar issues [within this PR](https://github.com/CDCgov/NEDSS-Modernization/pull/3143), given [Sonar said that it passed](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&id=CDCgov_NEDSS-Modernization&pullRequest=3143), but upon merge apparently [there are still outstanding issues against `main`](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&id=CDCgov_NEDSS-Modernization). 🤔 

After some digging, my best guess is that by moving the `@Nullable` annotation _above_ the `@Override` annotation across the board resulted in the lines of code generating the code smells not actually being touched by that PR, resulting in not being analyzed in Sonar.  It also seems that [there's no way to trigger a full code-base scan outside of running a Sonar against a designated "long-lived" branch](https://community.sonarsource.com/t/how-do-i-trigger-a-full-codebase-scan/120758).

So in order to HOPEFULLY actually run a full code-base scan against this PR, I went ahead and followed [the guidance outlined here ](https://community.sonarsource.com/t/how-do-i-trigger-a-full-codebase-scan/120758/8) and created a temporary "long-lived" branch by following our designated long-lived branch pattern of `(branch|release)-.*`. Third time's the charm hopefully 🤞 

**Update**: It's still not running a full scan 😭 I _do_ think I've addressed all the code smells this time around, but at this point I'm unclear on how to be 100% sure of that without either merging this into `main` and checking again, or actually modifying the lines of code in question.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-379)

## Checklist before requesting a review
- [ ] PR focuses on a single story
- [ ] Code has been fully tested to meet acceptance criteria
- [ ] PR is reasonably small and reviewable (Generally less than 10 files and 500 changed lines)
- [ ] All new functions/classes/components reasonably small
- [ ] Functions/classes/components focused on one responsibility
- [ ] Code easy to understand and modify (clarity over concise/clever)
- [ ] PRs containing TypeScript follow the [Do's and Don'ts](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)
- [ ] PR does not contain hardcoded values (Uses constants)
- [ ] All code is covered by unit or feature tests

PR #3161, opened 2026-04-13
APP-379, Take Four - Suppress remaining Sonar code smells

JordanGuinn:
## Description

[My previous PR](https://github.com/CDCgov/NEDSS-Modernization/pull/3152) attempting to remediate the outstanding Sonar issues didn't take, so this PR simply suppresses the remaining issues across the board.  The two types of issues being suppressed here are as follows:

### 1. Parameter 1 to this call is marked "@Nonnull" but null could be passed.
There are 3 instances of this issue still being thrown.  I decided to suppress these because it's hard for me to see these as anything other than false positives at this point - all three instances of this error are contained within an explicit `!= null` check for the parameter being described as possibly `null`.  That, combined with the presence of [an outstanding Sonar issue](https://sonarsource.atlassian.net/browse/JAVASE-72) regarding false positives for this very rule, has me pretty confident this is a Sonar bug.
Screenshot 2026-04-13 at 3 21 48 PM

### 2. Fix the incompatibility of the annotation @Nullable to honor @NonNullApi at package level of the overridden method.
This issue also exists in 3 different spots within the repo.  In all 3 cases, there is a class implementing a `@Nullable`-annotated method of a third-party interface with an equivalent `@Nullable`-annotated method.  I'm wondering if there is a `@NonNullApi` annotation specified on the parent third-party packages that Sonar is getting tripped up on, but I decided to suppress them because I'm not sure what more can be done to demonstrate compatibility.
Screenshot 2026-04-13 at 3 51 17 PM

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-379)

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/data/pagination/DirectionCoercing.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/data/pagination/DirectionCoercing.java
@@ -32,6 +32,13 @@ class DirectionCoercing implements Coercing<Sort.Direction, String> {
     return (input instanceof String value) ? FlexibleDirectionConverter.from(value) : null;
   }
 
+  //  Suppressing java:2637 ("@NonNull" values should not be set to null) because there is a known
+  //  issue regarding that Sonar rule throwing false positives in certain situations, which seems
+  //  to be the case here.  The `value.getValue() != null` check prior to invoking `parseValue`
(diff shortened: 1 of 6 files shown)


krista-skylight (approved):
Sounds like you did your due diligence and left thorough in-line comments so LGTM!

brick-green (approved):
Agree with Krista; excellent comments

PR #3162, opened 2026-04-14
fix: data source mapping configuration usage

mcmcgrath13:
## Description

Found an issue with the data mapping not loading via docker build when trying out functional testing flow. This addresses that. It's currently into the env compare PR, but I won't merge until that's in and this can go to main


emyl3 (approved):
Thank you for fixing 🙏

JordanGuinn (approved):
✅

PR #3166, opened 2026-04-15
[APP-417] feat: first basic filter(s) (text+date range)

mcmcgrath13:
## Description

Implement our first basic filter and set up front end and apis for hopefully long lasting patterns/utility:
* Give up on `Filter` as one concept and use the basic and advanced all the places - we kept splitting them and mushing back together and it felt not worth it given they pretty much always have different handling requirements
* Start the logical flow of umbrella run page handling the params/meta state with a configuration page and result page for those sub-states
* use react form hook (as in other places)
* set up a `BasicFilter` wrapper component that deals with common logic and setting up the form state
* First `TextFilter` to show pattern of the component (future ones may need to do some handling around more complex value set fetching/state dealing) and the default value getter (assuming that sometimes we'll need some light serialization/deserializaiton between the form state and config state)
* Also did `DatePickerRange` to make sure the patterns established would work with a more complex multi-field type filter

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-417


apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportConfiguration.java
```diff
@@ -8,7 +8,9 @@ public record ReportConfiguration(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportDataSource dataSource,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Library reportLibrary,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String reportTitle,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) List<FilterConfiguration> filters,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+        List<BasicFilterConfiguration> basicFilters,
+    AdvancedFilter advancedFilter,
```
mcmcgrath13:
```suggestion
    AdvancedFilterConfiguration advancedFilter,
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/FilterDefaultValue.java
```diff
@@ -1,11 +0,0 @@
-package gov.cdc.nbs.report.models;
```
JordanGuinn, line 1: I see this was pulled from what used to be `FilterConfiguration` - did this data end up not really being necessary at all?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterRequest.java
JordanGuinn, line 1: If I'm parsing this correctly, `BasicFilter` and `AdvancedFilter` are entities driven by the UI for incoming report execution requests, and `BasicFilterConfiguration`/`AdvancedFilterConfiguration` are equivalent entities extended for API usage when consuming `ReportConfiguration`.  If so, thoughts on renaming them to `BasicFilterRequest`/`AdvancedFilterRequest` or similar, or bundling them as types within `ReportConfiguration`?

I only mention I was initially expecting `BasicFilter` and `AdvancedFilter` to be props within `BasicFilterConfiguration`/`AdvancedFilterConfiguration` respectively - at least that's how those types read to me.

JordanGuinn (commented):
Couple q's around naming and types in general, but overall this Filter split makes sense to me!  I imagine it'll make API usage and general testing simpler and more explicit, so I'm all for it 🚀

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/FilterDefaultValue.java
mcmcgrath13: for the basic configuration we need very minimal data from it in practice it seems. For advanced we'll need more, but the target format is also very different, so it seemed easier to just inline the data where needed

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/BasicFilterRequest.java
mcmcgrath13: gone with `Request` to match the `ReportExecutionRequest`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
brick-green, line 1: (suggestion, nb) Could use some commenting here. It took me a bit to parse out what this was for. The tests were helpful though.

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
brick-green, line 1: (q, b) Should the AdvancedFilterConfiguration be mocked in here too?

brick-green (commented):
Small questions

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
mcmcgrath13: this code is about to all be ripped out and changed in an upcoming PR where we fix up these types, so I don't want to do anything right now - is that okay?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
mcmcgrath13: Because nothing is done with advanced filters yet and that spec has changed (and the code is lagging) it didn't feel worth it at this time. When the advanced config is ipmlemented and used it should be updated in these tests at that point

brick-green (approved):
🆒

JordanGuinn (approved):

PR #3169, opened 2026-04-16
APP-335: Validate ReportSpec as part of report execution

JordanGuinn:
## Description

~This PR validates the `ReportSpec` model that comes in to the report execution service during any report exec request.  Given [FastAPI and Pydantic together validate the underlying shape of the request automatically](https://fastapi.tiangolo.com/tutorial/body/#results) (and we're already verifying whether or not the corresponding library exists), this PR focuses on parameters that are correctly typed but otherwise invalid.~

**UPDATE:** As we've [since decided to remove time range as a prop](https://skylight-hq.slack.com/archives/C09HXK6MKPE/p1776367894497399), this PR simply adds empty string checks to the relevant properties on the `ReportSpec` Pydantic model, along with a number of tests to ensure that request body validation is operating as expected.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-335)


apps/report-execution/src/models.py
```diff
@@ -1,25 +1,25 @@
...
     is_export: bool
     is_builtin: bool
-    report_title: str
-    library_name: str
-    data_source_name: str
-    subset_query: str
+    report_title: str = Field(min_length=1)
+    library_name: str = Field(min_length=1)
+    data_source_name: str = Field(min_length=1)
+    subset_query: str = Field(min_length=1)
```
mcmcgrath13, line 22: in practice, these should all be longer, but I'm not sure where would be the 100% all the time always correct cut off - thoughts?

emyl3 (approved):
LGTM!

apps/report-execution/src/models.py
JordanGuinn: Agreed!  I'm not sure what a good cut-off would be either; my goal with this was just to treat empty strings the same as `None` values, which Pydantic (understandably) does not do by default

krista-skylight (approved):
(q, nb) Was there anything you came across that made you want to write these checks? Jw since I noticed the UI throws and error if your start date is before your end date and doesn't even allow the user to pick dates in the future. But regardless, it doesn't hurt to do extra validation!

JordanGuinn:
@krista-skylight Nothing specific, just seemed important to handle various input error cases!  And agreed, certainly nice that the UI has some similar gates as well, but imo validation of any one layer of the stack shouldn't make any assumptions about similar validation happening across the surrounding layers of the stack

mcmcgrath13 (approved):
🏁

PR #3179, opened 2026-04-21
APP-336 - Validate report result returned from report library invocation

JordanGuinn:
## Description

This PR adds validation to the report execution service, specifically around the resulting response from a given report library's invocation, and ensuring it matches the general shape we'd expect.  Let me know if there are any other aspects of invocation worth testing here.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-336)


apps/report-execution/src/errors.py
```diff
@@ -49,6 +49,16 @@ def __init__(self, message):
...
 
 
+class InvalidResultError(BaseReportExecutionError):
+    """The report result from library execution is invalid."""
+
+    def __init__(self, library_name: str, message=None):
+        error_message = f'Invalid report result from library `{library_name}`'
+        if message:
+            error_message += f': {message}'
+        super().__init__(error_message, 422)
```
mcmcgrath13: if the report result is invalid, this seems more likely to be a system internal variant - thoughts? 

If the request is valid, the result being bad seems like the fault of the library and not the user

mcmcgrath13 (approved):
overall looks good - just the question/thought on the error code/messaging outward

apps/report-execution/src/errors.py
JordanGuinn: Ahhhh, no that makes sense - while it's validation, it's not predicated necessarily on anything the user has provided.  Will update to 500!

PR #3141, opened 2026-04-07
APP-354 - Convert NBSSR00011 (Cases of Selected Diseases By Year Over Time)

JordanGuinn:
## Description

This PR converts the NBSSR00011 SAS report to a Python equivalent.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-354)

## Checklist for adding a library:
- [x] Copy this checklist into your (draft) PR description
- [x] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelog/report/execution/03_ODSE_Data_Report_Library_Init.sql`
  - [x] If the report execution service is not yet in use by any STLT (even beta), you should update this file
  - [x] Update the description in the library-specific migration
- [x] Add a migration to `apps/modernization-api/src/main/resources/db/report/execution/libraries` named `<your library name>.sql`
  - [x] Copy another migration and update the variables at the top of the file
- [x] Add the migration to `apps/modernization-api/src/main/resources/db/report-execution-changelog.yml`. It should be added to the latest `changeSet` since the last release - this could require a new `changeSet` if there isn't one since last release
- [ ] Check the migration works by building/running the containerized dev setup
- [x] Add a python library file to `apps/report-execution/src/libraries` named in lowercase, but generally following the naming convention of SAS (needs to be human recognizable as the same library)
- [x] Follow the pattern established in existing libraries to create your library. The `execute` function is the required method and its signature will (someday) be checked for validity
  - [x] Make sure pertinent text embedded in the reports is included in the description of the report
  - [x] Note in the doc string any deviation from the SAS
- [x] Keep an eye out for things in the sas library that are/could be centralized to pre- or post-processing or a util would be helpful for (e.g. common formatting needs)
- [x] Add integration tests in a `apps/report-execution/tests/libraries/<your_library_here>py` file following the conventions established for other libraries
  - [x] The `subset_sql` can be assumed to be well tested by the modernization-api, so focus on any logic and additional joins/queries/analysis that is added in the library
  - [x] The database used in integration testing is our standard golden db + seed used in other testing - if additional data is needed to test the report, add it to the seeding process
- [x] Add e2e test coverage for the report
  - [x] If this step is not set up, create a ticket to track this as follow on work needed for this report
- [x] Check that the report has parity with the NBS 6 version using (process TBD)
  - [x] If this step is not set up, create a ticket to track this as follow on work needed for this report
- [x] Keep an eye out for opportunities to:
[...]

PR #3180, opened 2026-04-21
[APP-355][APP-503] feat: add sr12 as sr11, fix sr11 per functional testing

mcmcgrath13:
## Description

SR 12 is the same as SR 11 save the capitalization of the "Year" column header, so have SR 12 reports repoint to SR 11. Also did functional testing of both SR 11 and SR 12 and found some differences in sort order, which have been addressed.

Notes:
* Added `--pull` to the nbs 6 setup so DB's don't go stale
* Changed `Polk County` to `Middlesex County` so our counties are on either side of `N/A`, which better tests sort order
* Have the downloaded file name match the report config and not the result (though maybe we should have the result take the report request's name in general unless it's actually wrong (like for nbs custom where we add in the data source name))
* readme note on networking host mode
* Changed the csv serialization so we keep the smallest precision we need to match how sas generally serializes numbers, as well as `\r\n` new lines and no trailing new line

- [ ] Add a python library file to `apps/report-execution/src/libraries` named in lowercase, but generally following the naming convention of SAS (needs to be human recognizable as the same library)
- [ ] Follow the pattern established in existing libraries to create your library. The `execute` function is the required method and its signature will (someday) be checked for validity
- [ ] Add integration tests in a `apps/report-execution/tests/libraries/<your_library_here>py` file following the conventions established for other libraries

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-503
* https://cdc-nbs.atlassian.net/browse/APP-355


apps/report-execution/README.md
```diff
@@ -70,7 +70,7 @@ Or something similar to the following, if you're looking to run a specific test
 uv run pytest src/main_test.py
 ```
 
-To run integration tests, ensure the `.env` is loaded or passed to `uv` and run
+To run integration tests, ensure the `.env` is loaded or passed to `uv`, **ensure host networking mode in docker is enabled**, and run
```
emyl3, line 73: 🙌

apps/modernization-api/src/main/resources/db/report/execution/libraries/nbs_sr_12.sql
```diff
@@ -0,0 +1,18 @@
+-- Migrate the NBSSR00011.SAS library to the nbs_sr_11 python library
+
+USE [NBS_ODSE]
+
+-- SR12 is the same as SR11 once pivoting is removed. Re-pointing SR12 to SR11 via migration
```
emyl3, line 5: ❓, blocking maybe?? -- Because we are doing this... when we generate the report for a SR12 report the heading will have SR11 is that OK?

mcmcgrath13: I've been wondering if we want to change those returned report headers to generally use the report title unless there's a good reason not to - thoughts?

emyl3: That makes more sense to me!! I can't think of a reason we wouldn't want to do that... and that would mean we wouldn't have to maintain a separate hardcoded header values...

apps/report-execution/tests/integration/assets/tablefaker_schema/phc_demographic.yaml
```diff
@@ -58,13 +58,13 @@ tables:
             return None
       - column_name: county
         type: string
-        data: random.choice(["Washington County", "Polk County"]) if random.random() < .95 else None
+        data: random.choice(["Washington County", "Middlesex County"]) if random.random() < .95 else None
```
emyl3, line 61: I'm still a little confused why we had to change the name even though I read the PR description 😅

mcmcgrath13: `Mi.. < N/A < Wa..` but `N/A < Po.. < Wa` AND `NULL < Po.. < Wa` potentially with sort order - basically make sure we know if we should sort on `NULL` or `"N/A"` more decisively

apps/modernization-api/src/main/resources/db/report/execution/libraries/nbs_sr_12.sql
mcmcgrath13: I'll pull that work into a separate PR

apps/report-execution/tests/integration/assets/tablefaker_schema/phc_demographic.yaml
emyl3: Ah ha!!! Thank you for breaking that down hahah 🧠 💡

emyl3 (approved):
LGTM!

apps/report-execution/tests/main.py
```diff
@@ -50,8 +50,15 @@ def test_execute_report_api_with_valid_spec(self, client, mock_db_transaction):
         assert result
 
         # check we can round trip back to DF
-        str_io = io.StringIO(result['content'])
+        content = result['content']
+        assert '1,a' in content
```
JordanGuinn, line 54: (q, nb): Just for my own clarification - this is confirming the resulting content matches the mock data defined in `conftest.py`, reads it into a dataframe, and confirm that same data can be read with slightly reduced precision?

apps/report-execution/src/libraries/nbs_sr_11.py
```diff
@@ -13,21 +13,20 @@ def execute(
 
     Conversion notes:
     * Matched export format without pivot
+    * Made "Year" header uppercase y to match SR12 output (and other columns)
```
JordanGuinn, line 16: (q, nb): I remember keeping `year` as lowercase for 11 was [an explicit feedback request](https://github.com/CDCgov/NEDSS-Modernization/pull/3141#discussion_r3051884287) - is that no longer a requirement?

apps/modernization-api/src/main/resources/db/report/execution/libraries/nbs_sr_12.sql
mcmcgrath13: https://github.com/CDCgov/NEDSS-Modernization/pull/3184

apps/report-execution/tests/main.py
mcmcgrath13: correct and that it gets written at the precision we expect

apps/report-execution/src/libraries/nbs_sr_11.py
mcmcgrath13: generally, it is. We're using this library to now power two prior sas libraries and the only thing they differ on is that capitalization. Given all other columns are capitalized, we're now matching the sr12 format insteda

JordanGuinn (approved):

PR #3184, opened 2026-04-22
fix: generally defer to report title for header

mcmcgrath13:
## Description

The SAS report headers are hardcoded, but it generally makes more sense and gives more flexibility and a consistent experience to use the report's title (nbs custom is an exception)


emyl3 (approved):
Thanks for adding this so quickly!

brick-green (approved):
🥇

PR #3193, opened 2026-04-28
[APP-488] feat: Convert NBSSR00008 and functional testing updates of nbs_sr_02 and nbs_sr_08

emyl3:
- [X] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelog/report/execution/03_ODSE_Data_Report_Library_Init.sql`
  - [X] If the report execution service is not yet in use by any STLT (even beta), you should update this file
  - [] Update the description in the library-specific migration
  - [X] Make sure pertinent text embedded in the reports is included in the description of the report
  - [X] Note in the doc string any deviation from the SAS
- [X] Keep an eye out for things in the sas library that are/could be centralized to pre- or post-processing or a util would be helpful for (e.g. common formatting needs)
- [X] Add integration tests in a `apps/report-execution/tests/libraries/<your_library_here>py` file following the conventions established for other libraries
  - [X] The `subset_sql` can be assumed to be well tested by the modernization-api, so focus on any logic and additional joins/queries/analysis that is added in the library
  - [X] If this step is not set up, create a ticket to track this as follow on work needed for this report ([ticket](https://cdc-nbs.atlassian.net/browse/APP-547))
- [X] Check that the report has parity with the NBS 6 version using (https://cdc-nbs.atlassian.net/wiki/spaces/NE/pages/2372272139/Comparison+Testing+Libraries)
  - [x] Consolidate reports by using parameters (parameter design tbd)


apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
             parts.append(', '.join(clean_diseases))
 
     return ' | '.join(parts)
+
+
+def build_case_count_query(column_mapping: dict[str, str], subset_query: str):
```
emyl3: ⚠️ Hopefully this is the best place for this? If not, I can move!

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
...
+        else:
+            select_fields.append(f'{db_col} AS "{alias}"')
+    select_field_query = ', '.join(select_fields)
+    group_cols = ', '.join(column_mapping.keys())
+    return (
+        f'WITH subset as ({subset_query})\n'
+        f'SELECT {select_field_query}, '
+        f'sum(group_case_cnt) as Cases\n'
+        f'FROM subset\n'
+        f'WHERE event_date IS NOT NULL\n'
```
emyl3: ⚠️ Added `f'WHERE event_date IS NOT NULL\n'` after functional testing of nbs_sr_02 and nbs_sr_08

**NBSSR00002 - SAS**
[04012010-04282026_SR2-SAS_GA-All_conditions.csv](https://github.com/user-attachments/files/27182255/04012010-04282026_SR2-SAS_GA-All_conditions.csv)

**NBS_SR_02 - Python**
[SR2_ Counts of Reportable Diseases by County for Selected Time Frame.csv](https://github.com/user-attachments/files/27182256/SR2_.Counts.of.Reportable.Diseases.by.County.for.Selected.Time.Frame.csv)

**NBSSR00008 - SAS**
[04012010-04282026_SAS-SR8-Measles-GA.csv](https://github.com/user-attachments/files/27182257/04012010-04282026_SAS-SR8-Measles-GA.csv)

**NBS_SR_08 - Python**
(Couldn't filter this to match exactly due to limitations on front end + backend support but the values are all there 😅)
[SR8_ State Map Report of Disease Cases Over Selected Time Period (1).csv](https://github.com/user-attachments/files/27182254/SR8_.State.Map.Report.of.Disease.Cases.Over.Selected.Time.Period.1.csv)

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
...
+        if db_col == 'event_date':
+            select_fields.append(
+                f'''UPPER(FORMAT({db_col}, 'ddMMMyyyy:HH:mm:ss.fff')) AS "Event Date"'''
+            )
+        else:
+            select_fields.append(f'{db_col} AS "{alias}"')
+    select_field_query = ', '.join(select_fields)
+    group_cols = ', '.join(column_mapping.keys())
+    return (
+        f'WITH subset as ({subset_query})\n'
```
brick-green: suggestion, b: can this be formatted as a triple quoted f string to match the queries in the libraries?

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
...
+        f'SELECT {select_field_query}, '
+        f'sum(group_case_cnt) as Cases\n'
+        f'FROM subset\n'
+        f'WHERE event_date IS NOT NULL\n'
+        f'GROUP BY {group_cols}\n'
+        f'ORDER BY {group_cols}'
```
brick-green: nit, nb: probably a good practice to terminate our queries?
```suggestion
        f'ORDER BY {group_cols};'
```

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
...
+        'state',
+        'county',
+        'phc_code_short_desc',
+        'event_date',
+        'cnty_cd',
+    }
+    if not subset_query:
+        raise ValueError('Subset query cannot be empty')
+
+    if not all(key in valid_columns for key in column_mapping):
```
brick-green: question, nb: Do you need to account for case sensitivity here?

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
...
+
+    select_fields = []
+    for db_col, alias in column_mapping.items():
+        if db_col == 'event_date':
+            select_fields.append(
+                f'''UPPER(FORMAT({db_col}, 'ddMMMyyyy:HH:mm:ss.fff')) AS "Event Date"'''
+            )
+        else:
+            select_fields.append(f'{db_col} AS "{alias}"')
+    select_field_query = ', '.join(select_fields)
```
brick-green: question, b: What happens here if only a single column is sent through?

brick-green (commented):
got a few questions

apps/report-execution/src/utils.py
emyl3: Since the libraries we are defining would be calling this I didn't want to account for case sensitivity here and just make sure the columns are exactly the ones being requested. What do you think, Brick? 👀

apps/report-execution/src/utils.py
emyl3: Done -- let me know if the update looks good!

apps/report-execution/src/utils.py
emyl3: Added it when updating to triple quoted f string

apps/report-execution/src/utils.py
emyl3: I believe the join method won't add the extra comma if there's only one column but I wrote a test to cover that scenario!

apps/report-execution/src/utils.py
brick-green: yeah I don't mind leaving it as is and solving for it later if a problem comes up

brick-green (approved):
🏆 🥇

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,40 @@ def gen_subheader(
...
+        'county',
+        'phc_code_short_desc',
+        'event_date',
+        'cnty_cd',
+    }
+    if not subset_query:
+        raise ValueError('Subset query cannot be empty')
+
+    if not all(key in valid_columns for key in column_mapping):
+        raise ValueError('One or more columns are invalid')
```
krista-skylight: nit, sugg: might be more helpful as an error message to store the invalid columns and name them in the error message

krista-skylight (changes requested):
Only left one small suggestion about the `ValueError` message but otherwise this looks great!

apps/report-execution/src/utils.py
emyl3: Thank you, Krista! This has been updated to list the invalid columns!

krista-skylight (approved):
thank you!

apps/report-execution/src/utils.py
```diff
@@ -76,3 +76,44 @@ def gen_subheader(
             parts.append(', '.join(clean_diseases))
 
     return ' | '.join(parts)
+
+
+def build_case_count_query(column_mapping: dict[str, str], subset_query: str):
```
JordanGuinn: (q, nb): This is somewhat me playing devil's advocate here, but I'm wondering if we actually _need_ this utility?  I'm all for DRY, but it seems like there's enough of a difference between the two queries that sharing a util between them adds more effort and complexity than just writing each query out in their respective lib.

That's just my thinking though, very happy to leave as-is as well 🙂 Curious to hear what other folks think!

JordanGuinn (approved):
One general question about the utility, but overall this looks great! 🚀

apps/report-execution/src/utils.py
mcmcgrath13: I generally agree with @JordanGuinn , unless we think there's more times this will be used, I think it would be easier to maintain and read if these queries existed inline in hardcoded form in their respective libraries. It would have been nice if the outputs were pure subsets/remappings, but that just doesn't seem to be the case

emyl3: Thank you both for the feedback! I don't think it will be used more times... so I will fix this up!!

emyl3: Ok!! All fixed up! Let me know what you both think!!

mcmcgrath13: looks good to me!

PR #3195, opened 2026-05-04
build: add ts linting rules to check

mcmcgrath13:
## Description

Enable the typescript recommended lint rules on our lint and make sure that our regression tests check this (we currently have a crack where updated generated types cause undefined property access, but it isn't caught in CI)

### DIFF
--- a/apps/modernization-ui/eslint.config.mjs
+++ b/apps/modernization-ui/eslint.config.mjs
@@ -33,8 +33,8 @@ export default defineConfig([
             storybook,
             jsdoc,
         },
-        /* extends: ['jsdoc/recommended'],*/
         rules: {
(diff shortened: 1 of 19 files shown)


apps/modernization-ui/eslint.config.mjs
```diff
@@ -49,6 +49,7 @@ export default defineConfig([
                 'error',
                 { caughtErrors: 'none', destructuredArrayIgnorePattern: '^_' },
             ],
+            '@typescript-eslint/no-explicit-any': 'warn',
```
mcmcgrath13, line 52: there are a lot of these...

JordanGuinn (approved):
🧹

PR #3197, opened 2026-05-04
[APP-524] feat: advanced filter setup

mcmcgrath13:
## Description

Wire up the query builder to use for the advanced filter builder

Notes:
* Updated the advanced filter configuration / request on the API side as per the design (plus ID fields from learnings of implementation - doc updated)
* Made `ReportColumn` a bit tighter and better matching of the reality of what's required for the app to function (which doesn't match db required-ness 😭 )
* Fixed the overflow styling on the collapsible cards as it was causing some very unfortunate clipping
* The Accessibility of the base query filter components is not up to par, but when we swap in our own components, it should become accessible (and those checks turned on)
* The date format of the base date picker is ISO and not `mm/dd/yyyy` as we have in the DB - when we swap to our datepicker, this should be fixed
* Went with the dnd-kit drag and drop adapter since it looks like it's the closest to what we need for keyboard interactions to work (in follow on ticket)
* The react form hook was interpretting the uids as array indices, so added the `id_` prefix so they'd be treated as object keys instead (the ids can be pretty big and don't need to create that large of an array)

Questions:
* For the `field` on the config/request, it's currently the columnUid, but in practice we have to translate this to the column's name for the query builder to work. Given we need to translate, should we a) have the field just be the column name? (makes validation kinda weaker perhaps? though maybe not as we still know what column names are valid in the context of the report/filter)

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-524


apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportColumn.java
```diff
@@ -5,10 +5,10 @@
...
 public record ReportColumn(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long id,
-    Integer columnMaxLength,
-    String columnName,
-    String columnTitle,
-    String columnSourceTypeCode,
+    Integer maxLength,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String name,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String title,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String sourceTypeCode,
```
mcmcgrath13, line 13: these are all required in the UI and are always populated in practice, without them, things don't work

apps/modernization-ui/src/design-system/card/collapsible.module.scss
```diff
@@ -12,15 +12,15 @@
 
 .collapsible {
     @include expanded('height');
-    overflow: hidden;
+    overflow: visible;
```
mcmcgrath13, line 15: this fixes the weird card clipping pop up date pickers and multi-select drop downs

apps/modernization-ui/src/generated/models/ReportColumn.ts
```diff
@@ -4,10 +4,10 @@
 /* eslint-disable */
 export type ReportColumn = {
     id: number;
-    columnMaxLength?: number;
-    columnName?: string;
-    columnTitle?: string;
-    columnSourceTypeCode?: string;
+    maxLength?: number;
+    name: string;
+    title: string;
```
mcmcgrath13, line 9: simplify names for the API since we already know we're in the context of a column

apps/modernization-ui/src/apps/report/run/filters/AdvancedFilter.tsx
krista-skylight, line 1: Super minor sugg, nb: maybe move the constants to another file for organization since this module is a bit long? Not super opinionated on this though.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportColumn.java
krista-skylight: sugg, nb: Might be worth an inline comment so future folks know why it's there

krista-skylight (approved):
Just left a couple minor suggestions! To answer your question, I don't think doing translations seems that cumbersome so if I had to pick I think staying with `columnUid` is fine.

apps/modernization-ui/src/apps/report/run/filters/AdvancedFilter.tsx
mcmcgrath13: I kept flip flopping on whether to split or not. It all felt a bit arbitrary, so I left with comment sectioning, but I feel 🥴 🤷 about it

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
```diff
@@ -3,25 +3,29 @@
...
-      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Expr left,
-      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Expr right)
+              allowableValues = {"or", "and"})
+          String combinator,
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) List<Expr> rules)
+      implements Expr {}
+
+  record Rule(
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String id, // uuid
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String field, // column UID stringified
```
JordanGuinn: (q, nb): What's the reason for not just naming this `columnId` or similar?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
```diff
@@ -3,25 +3,29 @@
...
+import java.util.List;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
 @JsonSubTypes({
-  @JsonSubTypes.Type(value = Expr.Clause.class, name = "Clause"),
-  @JsonSubTypes.Type(value = Expr.Connector.class, name = "Connector")
+  @JsonSubTypes.Type(value = Expr.RuleGroup.class, name = "RuleGroup"),
+  @JsonSubTypes.Type(value = Expr.Rule.class, name = "Rule")
 })
 public sealed interface Expr {
```
JordanGuinn: (q, nb): Should this parent model be renamed to `RuleGroup` or similar at this point?  Seems like we're now interacting directly with RuleGroups when working with advanced filters, so I'm no longer sure what an `Expr` really even is following this change

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -0,0 +1,25 @@
...
+import gov.cdc.nbs.report.models.Expr;
+import gov.cdc.nbs.report.models.FilterType;
+
+public class AdvancedFilterConfigurationMapper {
+  private AdvancedFilterConfigurationMapper() {}
+
+  public static AdvancedFilterConfiguration fromReportFilter(ReportFilter filter) {
+    FilterType filterType = FilterTypeMapper.fromFilterCode(filter.getFilterCode());
+
+    if (!filterType.filterType().equals("ADV_WCB")) {
```
JordanGuinn: (nit, nb): Thoughts on creating a const for this where clause builder filter type, and referencing everywhere we type this out?

JordanGuinn (approved):

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
mcmcgrath13: conflicting needs of things - thoughts on this question from the description?

> for the field on the config/request, it's currently the columnUid, but in practice we have to translate this to the column's name for the query builder to work. Given we need to translate, should we a) have the field just be the column name? (makes validation kinda weaker perhaps? though maybe not as we still know what column names are valid in the context of the report/filter)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
mcmcgrath13: I wasn't really sure what to name that since not all rules are rule groups. We could go with `RuleOrRuleGroup`? clunky, but accurate

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
mcmcgrath13: started a `ReportConstants.java` and seeded it with that value

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
JordanGuinn: Ohhhh, I missed that the `rules` in a given `RuleGroup` can still include another `RuleGroup`.  Either option makes sense in that case

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
JordanGuinn: If I'm understanding correctly, we either pass in the columnUid from the UI and fetch the corresponding name API-side, or just pass the column name directly from the UI?  If that's the case, I do think using the column ID is probably _slightly_ safer, in the off-chance there are somehow similarly named columns across multiple tables both pertinent to a given report, but also that seems _very_ unlikely.

Though in either direction, we'd probably benefit from renaming the `field` field to match whatever we decide to go with.

mcmcgrath13: only one table is allowed per report, so I don't think the cross-table name-clash issue is at play (though I agree technically the uid is safer). `field` is what the react querybuilder version of these types uses. If we go with `field` populated with the column's `name`, then we don't need to translate that property on submit, but still do on setup as we need more metadata. And other properties need to be translated on submit. I think going with straight up columnUid feels most consistent with the api as a whole

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Expr.java
mcmcgrath13: what about `AdvancedQuery`?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -0,0 +1,8 @@
+package gov.cdc.nbs.report;
+
+public final class ReportConstants {
+  public static final String ADV_FILTER_TYPE = "ADV_WCB";
+
+  private ReportConstants() {}
+  ;
```
mcmcgrath13:
```suggestion
```

PR #3209, opened 2026-05-05
APP-489: Convert NBSSR00010 (Multi-Year Line Graph of Disease Cases)

JordanGuinn:
## Description

This PR converts NBSSR00010 to Python (which is effectively just NBSSR00011 in line-graph form).

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-489)

- [ ] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelog/report/execution/03_ODSE_Data_Report_Library_Init.sql`
- [ ] Add a python library file to `apps/report-execution/src/libraries` named in lowercase, but generally following the naming convention of SAS (needs to be human recognizable as the same library)
- [ ] Follow the pattern established in existing libraries to create your library. The `execute` function is the required method and its signature will (someday) be checked for validity
- [ ] Keep an eye out for things in the sas library that are/could be centralized to pre- or post-processing or a util would be helpful for (e.g. common formatting needs)
- [ ] Add integration tests in a `apps/report-execution/tests/libraries/<your_library_here>py` file following the conventions established for other libraries


apps/modernization-api/src/main/resources/db/report/execution/03_ODSE_Data_Report_Library_Init.sql
```diff
@@ -33,7 +33,7 @@ BEGIN
...
         ('NBSSR00008.SAS', 'SR8: Report of Disease Cases Over Selected Time Period. Report demonstrates, in table form, the total number of Investigation(s) [both Individual and Summary] by County irrespective of Case Status.'),
         ('NBSSR00009.SAS', 'SR9: Selected Disease by Month and County.  Report demonstrates the total number of monthly Investigation(s) [both Individual and Summary] for a given disease and State, by County, irrespective of Case Status.'),
-        ('NBSSR00010.SAS', 'SR10: Multi-Year Line Graph of Disease Cases. Report demonstrates, using a multi-year line graph, the total number of yearly Investigation(s) [both Individual and Summary] for a given disease, by State, irrespective of Case Status.'),
+        ('NBSSR00010.SAS', 'SR10: Multi-Year Line Graph of Disease Cases. Report demonstrates, in table form, the total number of yearly Investigation(s) [both Individual and Summary] for a given disease, by State, irrespective of Case Status.'),
```
JordanGuinn, line 36: Not sure what we want to do about the name of the report here - `Multi-Year Line Graph` is _in_ the title 😬 Do we want to simply rename it?

mcmcgrath13: I think I'd leave it since this is one we're not directly replacing and are ultimately deleting this data if all goes well

mcmcgrath13:
@JordanGuinn I don't see the functional testing CSVs in the library catalog - can you upload?

JordanGuinn:
@mcmcgrath13 Will do, just wanted to get enough data in the reports first so that comparisons were relatively meaningful.

apps/report-execution/src/libraries/nbs_sr_11.py
```diff
@@ -25,6 +25,7 @@ def execute(
         YEAR(event_date) as [Year],
         SUM(group_case_cnt) as [Cases]
         FROM subset
+        WHERE event_date IS NOT NULL
```
JordanGuinn, line 28: Added this to avoid discrepancies found during functional testing for both 11 and 10.

mcmcgrath13:
I ran diff, but it looks like there's a bunch of wonky whitespace on the python version - any ideas?
image

mcmcgrath13:
found the padding issue: https://github.com/CDCgov/NEDSS-Modernization/pull/3210

JordanGuinn:
@mcmcgrath13 Python report for testing has been regenerated and re-uploaded!

mcmcgrath13:
I'm still seeing non-printable character/new line discrepancies on the diff, which makes the diff noisy - any ideas why?

mcmcgrath13 (approved):
🥇

JordanGuinn:
Finally realized that the failing regression test is _not_ in fact a flaky test, but the result of the conversion/modernization of the SR10 report.  The test currently resides in the `nbs-classic` test suite, but I'm not sure if we want to keep validating that 6 version of the module, or instead validate 7 (or both!).  So I suppose the options I see here are:

- Update the existing test to work with the new 7 UI
- Update the existing test to instead reference 6 directly (or disable reporting for this test case)
- Skip the test entirely for now

PR #3213, opened 2026-05-07
[APP-527] feat: add keyboard nav to advanced filter drag and drop

mcmcgrath13:
## Description

Add keyboard-nav drag and drop to the advanced filter builder

Notes:
* Originally thought using dnd-kit adapter would be easiest since it has a keyboard sensor, but as time went on I realized that driving the dragging via keyboard was pretty painful/finnicky and we know what we want to happen, so we can just move things directly
* There's a shift button ux build into the query builder, but it adds even more buttons which isn't great, but looked at how it was implemented to create our bespoke drag handle with what we need
* Because of react re-mounting when a rule group changed level, needed to keep track of what drag handle is active at any given time so we can restore focus there when things move
* Changed the adapter to the atlaskit one as I generally found it smoother/more intuitive (and didn't want keyboard handlers to clash)

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-527


emyl3:
(❓, maybe b) It feels like we should override the default of scrolling when we hit the spacebar when we're focused on that element to toggle it. What do you think, Mary?

https://github.com/user-attachments/assets/62f4ea9c-eb8a-4b41-bc55-84115327776e

mcmcgrath13:
> (❓, maybe b) It feels like we should override the default of scrolling when we hit the spacebar when we're focused on that element to toggle it. What do you think, Mary?

Nice find! pushed up a fix

apps/modernization-ui/src/apps/report/run/filters/AdvancedFilter.tsx
```diff
@@ -253,7 +264,82 @@ const validateRule = (rule: RuleGroupTypeAny | RuleType | string, result: Valida
...
+        if (!dir) return;
+
+        // move the rule and update the query
+        const nextQuery = move(getQuery(), props.path, dir);
+        dispatchQuery(nextQuery);
+        event.preventDefault();
+    };
+
+    return (
+        <span
```
emyl3: Thank you for addressing the other issue! 🤩 
(❓) Another thing that came up... If a user tabs away from the drag and drop toggle with the toggle being on and then they return to it, the toggle is still on which could be confusing. 

Question 1: Should we auto turn off the toggle when they navigate away? Or leave it on?

There's nothing that seems to indicate it is on/off visually and with like an aria-label or something like that. I also tried navigating this using Mac's VoiceOver utility and nothing was obvious. 

Question 2: Should we add something here to address this visually and not visually? (Or would that be a separate ticket?)

mcmcgrath13: Excellent question! I'm going to work on making this match the experience for keyboard nav drag and drop on the patient search column preferences

apps/modernization-api/src/main/resources/db/report/execution/libraries/nbs_sr_12.sql
```diff
@@ -1,4 +1,4 @@
--- Migrate the NBSSR00011.SAS library to the nbs_sr_11 python library
+-- Migrate the NBSSR00012.SAS library to the nbs_sr_11 python library
```
emyl3, line 1: did you want this change in here? 👀

emyl3: oops nvm you already removed 😅 I was on a stale page

apps/modernization-ui/src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { ReactNode } from 'react';
+import styles from './VisuallyHidden.module.scss';
+
+// From https://www.joshwcomeau.com/snippets/react-components/visually-hidden/
+// Display text for screen readers only, but in dev mode can hold `Alt` to display the value
```
emyl3, line 5: 💡 that's very neat/handy!!

apps/modernization-ui/src/apps/report/run/filters/advanced/ShiftableDragHandle.tsx
```diff
@@ -0,0 +1,86 @@
...
+const ShiftableDragHandle: ForwardRefExoticComponent<DragHandleProps & RefAttributes<HTMLSpanElement>> = forwardRef<
+    HTMLSpanElement,
+    DragHandleProps
+>((props, dragRef) => {
+    const { activeId, activate, reset, drag, drop } = useKeyboardDnd();
+    const id = props.ruleOrGroup.id!;
+    const [isActive, setIsActive] = useState<boolean>(activeId === id);
+    const { getQuery, dispatchQuery } = props.schema;
+
+    // When a rule group changes level, the component re-mounts and we need to move focus
```
emyl3, line 19: I see why we need to re-focus... From my testing around, it looks like doing that is causing [this message to be read again in VoiceOver](https://github.com/CDCgov/NEDSS-Modernization/pull/3213/changes#diff-8c8034c60165cdbd08ed2aedf3204c61f48804facf89ff73b69e9abae24e4290R75) which means that the message of moving the element can get missed...

mcmcgrath13: We could leave that as hidden description text in the front matter of the section, but not make it describe each drag handle? That relies on the user going through that content on the way to the widget - not sure if that's a better or worse UX as I can argue it either way @kenieh any thoughts here?

emyl3 (approved):
Thank you for all your work on this 🙌

apps/modernization-ui/src/apps/report/run/filters/advanced/useKeyboardDnd.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const announce = (message: string): void => {
+        // slight timeout improves screen reader flakiness
+        setTimeout(() => setAnnouncedMessage(message), 100);
+    };
+
+    return (
+        <KeyboardDndContext.Provider value={{ activeId, activate, reset, drag, drop }}>
+            <VisuallyHidden id="keyboard-dnd-instructions">
+                In the advanced filter builder below, you can drag and drop rules and groups to change the logic of how
+                statements are combined. When focused on a drag handle for a rule or group, press space bar to start and
```
krista-skylight, line 77: (sugg, nb) Maybe more of a comment for @kenieh but I feel like if at least the first 3 sentences here were visually available on the screen it would have helped me! I figured out how the `+Rule` and `+Group` buttons worked pretty quickly after seeing the `Preview Where Statement` but it wasn't immediately obvious.

krista-skylight (commented):
https://github.com/user-attachments/assets/0be9977d-dd61-4a82-9f01-5fe5108f6c49

(q, maybe b?)

This is cool! FYI, I played around with the tabbing and it didn't work for the second `Patient Last Name` filter. When I click the third row with the `Patient Date of Birth` filter, it works again. Hopefully you can see that in the screen recording.

Is this expected behavior or a legitimate bug?

mcmcgrath13:
> Screen.Recording.2026-05-14.at.12.07.10.PM.mov 
> (q, maybe b?)
> 
> This is cool! FYI, I played around with the tabbing and it didn't work for the second `Patient Last Name` filter. When I click the third row with the `Patient Date of Birth` filter, it works again. Hopefully you can see that in the screen recording.
> 
> Is this expected behavior or a legitimate bug?

Nice Catch!! Ahhh - yeah. If the drag handle is "active" it doesn't let you tab away from it, which I'm guessing is what happened.

I looked more into how this is handled with the other dnd in the app and lifted/adapted some code from that library so it matches much more closely. I still need to appease the tests, but could you kick the tires?

krista-skylight:
It seems to be working as expected now!

mcmcgrath13:
> It seems to be working as expected now!

Awesome! I got the tests happy and expanded them, so ready for real review again

krista-skylight (approved):
lgtm!

PR #3220, opened 2026-05-12
[APP-494] feat: convert QA03.sas

emyl3:
## Description

Convert QA03.sas to python

## Tickets
[APP-494](https://cdc-nbs.atlassian.net/browse/APP-494)

- [X] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelog/report/execution/03_ODSE_Data_Report_Library_Init.sql`
  - [X] If this step is not set up, create a ticket to track this as follow on work needed for this report ([See ticket](https://cdc-nbs.atlassian.net/browse/APP-605))

[APP-494]: https://cdc-nbs.atlassian.net/browse/APP-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/report-execution/tests/integration/libraries/qa_03.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # Sanity check the result's shape beyond the snapshot
+        record = None
+        for row in result.content.data:
+            if (
+                row[0] == 'CAS10015820GA01'  # INV_LOCAL_ID
+                and row[2] == 'PSN10089835GA01'  # PATIENT_LOCAL_ID
+            ):
+                record = row
+                break
```
krista-skylight: question, I have been seeing this convention but didn't think to ask until now. Why not just write this as follows?

```
row = result.content.data[-1]
assert row[0] == 'CAS10015820GA01' and row[2] == 'PSN10089835GA01'
```

apps/report-execution/tests/integration/libraries/qa_03.py
```diff
@@ -0,0 +1,97 @@
...
+        base = {
+            'is_export': True,
+            'is_builtin': True,
+            'report_title': 'QA 03',
+            'library_name': 'qa_03',
+            'data_source_name': '[RDB].[dbo].[STD_HIV_DATAMART]',
+            'subset_query': 'SELECT * FROM [RDB].[dbo].[STD_HIV_DATAMART]',
+        }
+        base.update(overrides)
+        return ReportSpec.model_validate(base)
```
krista-skylight, line 27: that's neat!

krista-skylight (commented):
Looks pretty good but have one question!

mcmcgrath13:
I'm seeing a difference in the formatting of titles in the names - could you take a look?
image

apps/report-execution/src/libraries/qa_03.py
```diff
@@ -0,0 +1,52 @@
...
+        shd.PATIENT_NAME,
+        shd.PATIENT_LOCAL_ID,
+        shd.DIAGNOSIS_CD,
+        shd.PATIENT_AGE_REPORTED AS [PATIENT_AGE_REPORTED],
+        FORMAT(CAST(shd.CONFIRMATION_DT AS DATETIME), 'MM/dd/yyyy') 
+            AS [CONFIRMATION_DT],
+        shd.JURISDICTION_NM,
+        o.ORGANIZATION_NAME,
+        LTRIM(
+        RTRIM(
```
mcmcgrath13:
```suggestion
        TRIM(
```

does this get to the same place?

https://learn.microsoft.com/en-us/sql/t-sql/functions/trim-transact-sql?view=sql-server-ver17

apps/report-execution/tests/integration/libraries/qa_01.py
```diff
@@ -33,7 +33,7 @@ def test_execute_report_check_data(self, snapshot):
         assert result.content_type == 'table'
 
         data = result.content.data
-        assert len(data) == 108  # two combinations with no data, zeros not filled
+        assert len(data) == 80  # two combinations with no data, zeros not filled
```
mcmcgrath13: does this comment still make sense?

apps/report-execution/tests/integration/libraries/qa_03.py
```diff
@@ -0,0 +1,97 @@
...
+                break
+
+        assert record is not None
+        assert record[3] == '710'  # DIAGNOSIS_CD
+        assert record[4] == '36'  # PATIENT_AGE_REPORTED
+        assert record[5] == '06/09/2020'  # CONFIRMATION_DT
+        assert record[6] == 'Middlesex County'  # JURISDICTION_NM
+        assert record[7] == 'Torres Medical Center'  # ORGANIZATION_NAME
+        assert record[8] == 'Don'  # PROVIDER
+        assert record[9] == '89835'  # PATIENTID
```
mcmcgrath13: this is all in the snapshot itself and could easily change if the tablefaker evolves - is there a more probabaistic check we could do instead?

apps/report-execution/tests/integration/libraries/qa_03.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_execute_report_no_data(self, snapshot):
+        report_spec = self.create_spec(
+            subset_query='SELECT * FROM [RDB].[dbo].[STD_HIV_DATAMART] '
+            "WHERE patient_name = 'Russell, Lee'"
```
mcmcgrath13:
```suggestion
            "WHERE patient_name = 'Not a name'"
```

just to make it even more obvious (and we never accidentally break this)

apps/report-execution/tests/integration/libraries/qa_03.py
emyl3: That's all @brick-green! Thank you, Brick! 😸

apps/report-execution/src/libraries/qa_03.py
emyl3: Yes! Will update!!

apps/report-execution/tests/integration/libraries/qa_03.py
emyl3: OOOO good idea!! Thank you for the suggestion, Krista!

apps/report-execution/tests/integration/libraries/qa_01.py
emyl3: Nope! I will remove!

apps/report-execution/tests/integration/libraries/qa_03.py
emyl3: Updated! Let me know if that works better!

emyl3:
> I'm seeing a difference in the formatting of titles in the names - could you take a look? image

Oh nice catch!! I fixed this!!

apps/report-execution/tests/integration/assets/tablefaker_schema/std_hiv_datamart.yaml
```diff
@@ -1,25 +1,75 @@
...
+        null_percentage: 0.05
       - column_name: patient_sex
         data: random.choice(["F", "M", ""]) if random.random() < .95 else None
       - column_name: patient_race
         data: random.choice(["B", "A", "AI", "W", ""]) if random.random() < .95 else None
-
+      - column_name: patient_local_id
+        data: |
+          id_num = fake.unique.random_int(10000, 99999)
+          return f"PSN100{id_num}GA01"
```
mcmcgrath13: (q, b): I'm seeing the weird patient id math in the sas, can we make some of the test data start with a 2 instead of 1 to make sure we're getting the same behavior as 6?

apps/report-execution/tests/integration/libraries/qa_03.py
```diff
@@ -0,0 +1,84 @@
...
+        data = result.content.data
+        assert len(data) == 514
+        assert len(data[0]) == 10
+        assert len(data[0]) == len(result.content.columns)
+
+        snapshot.assert_match(yaml.dump(data), 'snapshot.yml')
+
+        # Sanity check the result's shape beyond the snapshot
+        row = result.content.data[-1]
+        assert row[0] == 'CAS10078848GA01' and row[2] == 'PSN10057362GA01'
```
mcmcgrath13: (q, b): are these values ones that are definitely always going to be in the result? this looks like it could be finnicky to maintain as is. The reason for the loop to find the row was to find a row that definitely matched an expected pattern as a sanity check. This seems to be emulating the snapshot, which is a different goal

apps/report-execution/tests/integration/assets/tablefaker_schema/std_hiv_datamart.yaml
emyl3: Yikes! You are right! Updated to do the weird patient id math 😓

apps/report-execution/tests/integration/libraries/qa_03.py
emyl3: Updated the assertions to check the shape -- let me know if that makes more sense!

emyl3: Updated qa01's integration test as well... let me know what you think!

mcmcgrath13: cc @krista-skylight who is also fighting this test in qa04

mcmcgrath13 (approved):
🚀 

heads up @krista-skylight this merge is probably going to be a bit rough with your qa04 PR

apps/report-execution/tests/integration/libraries/qa_03.py
mcmcgrath13: looks good!

krista-skylight (approved):
lgtm!

krista-skylight:
cc @timoballard @adaball this merge will touch your inflight PRs as well!

PR #3230, opened 2026-05-13
fix: sql migrations to allow multi-runs

mcmcgrath13:
## Description

Fix up the migrations so you can run them multiple times and still be happy and in the correct state


emyl3 (approved):
✅

PR #3232, opened 2026-05-13
[APP-604] feat: Add more options api's to support advanced filters

mcmcgrath13:
## Description

Add more endpoints to the options api, so they are available for the advanced filter value sets (https://cdc-nbs.atlassian.net/browse/APP-525).

Notes:
* Made `datasource-utils` project and moved those to a common spot in libs so they could be used for the options api and mod
* Tried to re-use `database-entites` for the finding of the column's name and data source, but this somehow introduced a bunch of auth stuff that I don't really understand why and decided it wasn't worth it to figure it out and instead just wrote a little query
* Did some minor fixing of existing queries to match the current nbs 6 query more closely
* There was a bug in a common step, so I fixed it, which then required fixing a bunch of tests

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-604


apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -56,6 +59,13 @@ public ReportConfiguration getReport(Long reportUid, Long dataSourceUid) {
                       .map(BasicFilterConfigurationMapper::fromReportFilter)
                       .toList();
 
+              AdvancedFilterConfiguration advancedFilter =
+                  report.getReportFilters().stream()
+                      .filter(f -> f.getFilterCode().getFilterType().equals("ADV_WCB"))
+                      .map(AdvancedFilterConfigurationMapper::fromReportFilter)
+                      .findFirst()
+                      .orElse(null);
+
```
mcmcgrath13, line 75:
```suggestion
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -69,7 +79,7 @@ public ReportConfiguration getReport(Long reportUid, Long dataSourceUid) {
                   new Library(report.getReportLibrary()),
                   report.getReportTitle(),
                   basicFilters,
-                  null, // advanced filter
+                  advancedFilter,
```
mcmcgrath13, line 89:
```suggestion
                  null, // advanced filter
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -35,7 +35,7 @@ private String buildSelectClause(List<ReportColumn> columns) {
 
     return "SELECT "
         + columns.stream()
-            .map(column -> "[" + column.columnName() + "] AS [" + column.columnTitle() + "]")
+            .map(column -> "[" + column.name() + "] AS [" + column.title() + "]")
```
mcmcgrath13, line 41:
```suggestion
            .map(column -> "[" + column.columnName() + "] AS [" + column.columnTitle() + "]")
```

apps/modernization-api/build.gradle
```diff
@@ -104,7 +105,7 @@ testing {
                 implementation libs.spring.test
                 implementation libs.spring.security.test
                 implementation.bundle(libs.bundles.testcontainers.elasticsearch)
-                implementation 'net.datafaker:datafaker:2.5.4' 
+                implementation 'net.datafaker:datafaker:2.0.2'
```
mcmcgrath13:
```suggestion
                implementation 'net.datafaker:datafaker:2.5.4'
```

libs/options-api/src/main/java/gov/cdc/nbs/option/condition/ConditionOptionFinder.java
```diff
@@ -15,6 +15,7 @@ class ConditionOptionFinder extends SQLBasedOptionFinder {
           indent_level_nbr    as [order]
       from NBS_SRTE..Condition_code
           where status_cd = 'A'
+            and condition_codeset_nm = 'PHC_TYPE'
```
mcmcgrath13, line 18: this part of the where was missing when translated from nbs 6 query (shouldn't make a difference in practice from what I see, but to be safe ported it over)

libs/options-api/src/main/java/gov/cdc/nbs/option/occupations/OccupationOptionFinder.java
```diff
@@ -15,7 +15,7 @@ class OccupationOptionFinder extends SQLBasedOptionFinder {
           row_number() over (order by [code]) as [order]
       from NBS_SRTE..NAICS_Industry_code
       where status_cd = 'A'
-        and parent_is_cd = 'root'
+        and indent_level_nbr = 1
```
mcmcgrath13, line 18: make match the nbs 6 query (though should be equivalent in practice)

libs/options-api/src/main/java/gov/cdc/nbs/option/program/area/list/ProgramAreaOptionFinder.java
```diff
@@ -0,0 +1,25 @@
...
+@Component
+class ProgramAreaOptionFinder extends SQLBasedOptionFinder {
+
+  private static final String QUERY =
+      """
+      select
+          prog_area_cd        as [value],
+          prog_area_desc_txt  as [name],
+          nbs_uid             as [order]
+      from [NBS_SRTE].[dbo].Program_area_code
```
mcmcgrath13:
```suggestion
      from NBS_SRTE..Program_area_code
```

libs/options-api/src/main/java/gov/cdc/nbs/option/jurisdictions/list/JurisdictionOptionFinder.java
```diff
@@ -0,0 +1,25 @@
...
+@Component
+class JurisdictionOptionFinder extends SQLBasedOptionFinder {
+
+  private static final String QUERY =
+      """
+      select
+          code                as [value],
+          code_short_desc_txt as [name],
+          indent_level_nbr    as [order]
+      from [NBS_SRTE].[dbo].Jurisdiction_code
```
mcmcgrath13:
```suggestion
      from NBS_SRTE..Jurisdiction_code
```

libs/options-api/src/test/java/gov/cdc/nbs/option/OptionsVerificationSteps.java
```diff
@@ -30,7 +30,7 @@ public void the_named_option_is_included(final String concept) throws Exception
   public void the_option_named_is_not_included(final String concept) throws Exception {
     this.response
         .active()
-        .andExpect(jsonPath("$.[*].name", hasItem(not(equalToIgnoringCase(concept)))));
+        .andExpect(jsonPath("$.[*].name", not(hasItem(equalToIgnoringCase(concept)))));
```
JordanGuinn, line 33: 😭

libs/options-api/src/main/java/gov/cdc/nbs/option/report/distinct/DistinctValuesFinder.java
```diff
@@ -0,0 +1,65 @@
...
+      return this.client
+          .sql(DATA_QUERY.formatted(columnName, dataSourceName))
+          .query()
+          .singleColumn()
+          .stream()
+          .map(o -> new Option(o.toString()))
+          .toList();
+
+    } catch (NumberFormatException | EmptyResultDataAccessException e) {
+      // report column uid not a number or  column not found
```
JordanGuinn, line 64: (q, nb): If we have a logger pattern established in options API already, made worth adding errors logs here?

JordanGuinn (approved):
✅ 🙏

libs/options-api/src/main/java/gov/cdc/nbs/option/report/distinct/DistinctValuesFinder.java
mcmcgrath13: I added a generic logger as we sometimes do elsewhere and added an info log. I could be talked into error, but it seems maybe heavy/noisy for what could be more of a stale datasource config issue - thoughts?

libs/options-api/src/main/java/gov/cdc/nbs/option/clinical/concept/ClinicalConceptController.java
```diff
@@ -0,0 +1,27 @@
...
+  ClinicalConceptController(final ClinicalConceptFinder finder) {
+    this.finder = finder;
+  }
+
+  @Operation(
+      operationId = "clinicalConcepts",
+      summary = "Clinical concept options by value set",
+      description = "Provides options from clinical concepts grouped into a value set.",
+      tags = "ClinicalConceptOptions")
+  @GetMapping("nbs/api/options/clinical/concepts/{name}")
```
brick-green: (q, nb): Does a leading slash make a difference? I'm seeing is in more the the endpoints we currently have but not all.
```suggestion
  @GetMapping("/nbs/api/options/clinical/concepts/{name}")
```

libs/options-api/src/main/java/gov/cdc/nbs/option/clinical/concept/ClinicalConceptFinder.java
```diff
@@ -0,0 +1,36 @@
...
+
+import gov.cdc.nbs.option.concept.ConceptOption;
+import gov.cdc.nbs.option.concept.ConceptOptionRowMapper;
+import java.util.Collection;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.simple.JdbcClient;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ClinicalConceptFinder {
```
brick-green, line 11: q, nb: I think the default is package private? Then that's why the controller class can access `find()`
```suggestion 
class ClinicalConceptFinder {
```

libs/options-api/src/main/java/gov/cdc/nbs/option/places/list/PlaceOptionFinder.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Component
+class PlaceOptionFinder extends SQLBasedOptionFinder {
+
+  private static final String QUERY =
+      """
+      select
+          ei.root_extension_txt                                 as [value],
+          p.nm                                                  as [name],
+          row_number() over(order by [ei.root_extension_txt])   as [order]
```
brick-green: q,b: do the brackets need to be around the parts?
```suggestion
          row_number() over(order by [ei].[root_extension_txt])   as [order]
```

brick-green (commented):
a few Qs

libs/options-api/src/main/java/gov/cdc/nbs/option/clinical/concept/ClinicalConceptController.java
mcmcgrath13: I'm honestly not sure. I copied from concepts and adapted as needed

libs/options-api/src/main/java/gov/cdc/nbs/option/clinical/concept/ClinicalConceptFinder.java
mcmcgrath13: What are the pros/cons to public here? I copied this from the county finder and that one is public

brick-green: I don't really see any pros and the con is relatively small tbh.  But how I'm seeing it the default visibility is package private. So the `find()` method is package private already. `ClinicalConceptFinder` being public makes it available outside the package but then `find()` wouldn't be. So the public accessor doesn't provide value.

libs/options-api/src/main/java/gov/cdc/nbs/option/clinical/concept/ClinicalConceptController.java
brick-green: Seems like the leading slash is optional and spring adds one if it's not there:

https://stackoverflow.com/questions/12742695/use-or-not-leading-slash-in-value-for-requestmapping-need-official-docs-or-poi?noredirect=1&lq=1

brick-green (approved):
the sql brackets was my only real concern 🥇

libs/options-api/src/main/java/gov/cdc/nbs/option/report/distinct/DistinctValuesFinder.java
JordanGuinn: Info is good with me ✅ Just wanted to make sure we surfaced em in some capacity

PR #3237, opened 2026-05-14
APP-495 Convert QA04 Cases Missing Lab and/or Treatment

krista-skylight:


apps/modernization-api/src/main/resources/db/report/execution/libraries/qa_04.sql
```diff
@@ -0,0 +1,43 @@
+-- Migrate the QA05.SAS library to the qa_05 python library
```
mcmcgrath13:
```suggestion
-- Migrate the QA04.SAS library to the qa_04 python library
```

apps/report-execution/src/libraries/qa_04.py
```diff
@@ -0,0 +1,59 @@
...
+    subset_query: str,
+    data_source_name: str,
+    **kwargs,
+):
+    """QA04 Cases Missing Lab and/or Treatment.
+
+    Cases Missing Lab or Treatment. This report generates a list, by name,
+    of individuals with cases that are not linked to a positive lab test record
+    (for this reported case) or to a treatment record.
+    """
```
mcmcgrath13, line 19: (nit, b): can you add the docstring section on conversion notes?

apps/report-execution/src/libraries/qa_04.py
```diff
@@ -0,0 +1,59 @@
...
+        AND (
+            b.LAB_TEST_KEY IS NULL
+            OR c.TREATMENT_KEY IS NULL
+        )
+    ORDER BY a.PATIENT_NAME
+    """
+
+    content = trx.query(full_query)
+
+    header = 'QA04 Cases Missing Lab and/or Treatment'
```
mcmcgrath13: (nit/q, b): can we instead defer to the report header here and leave this un-valued?

apps/report-execution/tests/integration/libraries/qa_01.py
```diff
@@ -40,13 +40,8 @@ def test_execute_report_check_data(self, snapshot):
...
         # Sanity check the result's shape beyond the snapshot
-        record = None
-        for row in result.content.data:
-            if row[0] == 45481149 and row[3] == 'Van Dam, Venus' and row[4] == '43':
-                record = row
-                break
-
-        assert record is not None
+        row = result.content.data[-1]
+        assert row[0] == 944 and row[3] == 'Van Dam, Venus' and row[4] == '43'
```
mcmcgrath13: (sugg, b): this isn't really sanity checking beyond the snapshot - ideally this test looks for things that should always be true about the data (even if the underlying data powering the snapshot changes in it's precise details). Things that should with a very high probability always be true about the shape of this result such as presence/absence of data, rough counts, relative counts, etc - can we update this test to something resiliently true?

apps/report-execution/tests/integration/libraries/qa_04.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+
+        result = execute_report(report_spec)
+        data = result.content
+
+        # All returned rows should be from the main subset
+        # (the UNION ALL with Suspect should be excluded)
+        assert len(data.data) == 0
+
+    def test_execute_report_duplicate_handling(self):
```
mcmcgrath13, line 138: these tests are all really excellent examples of the testing things that should always be true I was rambling about earlier ❤️

mcmcgrath13:
the diff is quite large for the two functional test reports - some of that is quoting, but most is real differences (in order at a minimum) - could you look into this?

image

krista-skylight:
> the diff is quite large for the two functional test reports - some of that is quoting, but most is real differences (in order at a minimum) - could you look into this?
> image

I had to make changes to my script and have loaded new comparison files to confluence!

apps/modernization-api/src/main/resources/db/report/execution/libraries/qa_04.sql
```diff
@@ -0,0 +1,43 @@
...
+    WHERE
+        UPPER(library_name) = @sasLib;
+END
+ELSE IF NOT EXISTS (SELECT * FROM [dbo].[Report_Library] WHERE library_name = @pyLib)
+BEGIN
+    -- Create a row for this library
+    INSERT INTO [dbo].[Report_Library] (
+        library_name,
+        desc_txt,
+        runner,
```
mcmcgrath13, line 27:
```suggestion
        runner,
        column_select_ind,
```

apps/modernization-api/src/main/resources/db/report/execution/libraries/qa_04.sql
```diff
@@ -0,0 +1,43 @@
...
+        is_builtin_ind,
+        add_time,
+        add_user_id,
+        last_chg_time,
+        last_chg_user_id
+    ) VALUES (
+        @pyLib,
+        @desc,
+        'python',
```
mcmcgrath13, line 37:
```suggestion
        'python',
        'N',
```

apps/report-execution/src/libraries/qa_04.py
mcmcgrath13: can you add any notes or fill in with it's a perfect match to the export format?

apps/report-execution/src/libraries/qa_04.py
```diff
@@ -0,0 +1,62 @@
...
+
+    """
+    full_query = f"""
+    WITH subset AS ({subset_query})
+    SELECT DISTINCT
+        a.INV_LOCAL_ID,
+        a.PATIENT_NAME,
+        a.PATIENT_LOCAL_ID,
+        a.DIAGNOSIS_CD,
+        CONVERT(varchar(10), a.CONFIRMATION_DT, 101) AS CONFIRMATION_DT,
```
mcmcgrath13:
```suggestion
        CAST(a.CONFIRMATION_DT AS DATE) AS CONFIRMATION_DT,
```

apps/report-execution/tests/integration/assets/tablefaker_schema/std_hiv_datamart.yaml
```diff
@@ -2,7 +2,7 @@ config:
   seed: 4
   nbs:
     fk_tables:
-      - '[RDB].[dbo].[CONFIRMATION_METHOD_GROUP]'
+        - '[RDB].[dbo].[CONFIRMATION_METHOD_GROUP]'
```
mcmcgrath13: (q, nb): why add more indent here?

apps/report-execution/src/libraries/qa_04.py
```diff
@@ -0,0 +1,62 @@
...
+        CASE
+            WHEN b.LAB_TEST_KEY IS NULL AND c.TREATMENT_KEY IS NULL
+                THEN 'No Treatment or Lab'
+            WHEN b.LAB_TEST_KEY IS NULL
+                THEN 'No Lab'
+            WHEN c.TREATMENT_KEY IS NULL
+                THEN 'No Treatment'
+            ELSE 'N/A'
+        END AS ERROR_TXT,
```
mcmcgrath13, line 36: (sugg, b): the test data doesn't seem to be stressing this logic at all since every row is coming back `'No treatment or Lab'` - can you update the faking so we hit each of the 4 cases?

apps/report-execution/tests/integration/libraries/qa_04.py
```diff
@@ -0,0 +1,159 @@
...
+
+        result = execute_report(report_spec)
+        data = result.content
+        error_col_idx = data.columns.index('ERROR_TXT')
+        for row in data.data:
+            assert row[error_col_idx] in (
+                'No Treatment or Lab',
+                'No Lab',
+                'No Treatment',
+            )
```
mcmcgrath13, line 112: can we also test we see all of the options? what about the 'N/A' case?

apps/report-execution/tests/conftest.py
```diff
@@ -241,8 +241,10 @@ def insert_fake_data(
     with db_transaction(conn_string) as trx:
         # Tables with foreign keys pointing to the table we want to replace need to
         # be backed up and cleared out to avoid FK constraint violations
+
```
mcmcgrath13: (nit, nb): revert changes to this file to keep the diff tidy

apps/report-execution/tests/integration/libraries/qa_04.py
krista-skylight: This line `IF ERROR_TXT NE 'N/A';` in the SAS code gets rid of 'N/A', so it shouldn't show up in the output at all. I handle this in my WHERE clause with `b.LAB_TEST_KEY IS NULL OR c.TREATMENT_KEY IS NULL`

apps/report-execution/src/libraries/qa_04.py
krista-skylight: nice catch, done! loaded new e2e comparison files to confluence as well

apps/report-execution/src/libraries/qa_04.py
```diff
@@ -0,0 +1,60 @@
...
+        SUBSTRING(a.PATIENT_LOCAL_ID, 4, 8) - 10000000 AS PATIENTID
```
mcmcgrath13, line 37: The uploaded files have a difference in column name here - can you confirm what this should be and update the files?

Image

apps/report-execution/tests/integration/libraries/qa_10.py
```diff
@@ -28,7 +28,7 @@ def test_execute_report_check_data(self, snapshot):
         assert result.content_type == 'table'
 
         data = result.content.data
-        assert len(data) == 26
+        assert len(data) == 22
```
brick-green, line 31: q, b: Why does this value change?

apps/report-execution/src/libraries/qa_04.py
krista-skylight: It should be `PATIENTID` without the underscore. I've been running the SAS script in the interactive session which unfortunately means I set the column names myself. That was likely a typo on my part.

apps/report-execution/tests/integration/libraries/qa_10.py
krista-skylight: QA10 also uses data from STD_HIV_DATAMART and relies on a join with the INVESTIGATION table. I moved those row ids a bit for the purpose of making my library have some records that join successfully and some that don't in order to create different `ERROR_TXT` categories. So that expectedly reduces the number of records in the snapshot for QA10.

apps/report-execution/src/libraries/qa_04.py
mcmcgrath13: ahhhh - cool cool

mcmcgrath13 (approved):
🥼

timoballard (approved):

PR #3238, opened 2026-05-18
APP-542 & APP-665: Add endpoints for admin report create and edit

JordanGuinn:
## Description

This PR adds new endpoints to the reports module for admin report creation and editing (including filter support as part of that).

## Tickets

* [First Ticket](https://cdc-nbs.atlassian.net/browse/APP-542)
* [Second Ticket](https://cdc-nbs.atlassian.net/browse/APP-665)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -33,18 +25,69 @@
...
+        reportLibraryRepository
+            .findById(request.libraryId())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "No report library found for ID " + request.libraryId()));
+
+    List<ReportFilter> reportFilters = null;
+    if (!request.filterIds().isEmpty()) {
+      reportFilters = reportFilterRepository.findAllById(request.filterIds());
```
mcmcgrath13: the report filters won't exist yet and need to be created, so this should probably be changed to a more full fledged filter configuration request (thinking down the road, the save as will have values to save, so probably want to include that optionally/plan ahead for how that will be handled)

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/Report.java
```diff
@@ -20,7 +20,7 @@
 @Entity
 @Table(name = "Report", catalog = "NBS_ODSE")
 public class Report {
-  @NonNull @EmbeddedId private ReportId id;
+  @EmbeddedId private ReportId id;
```
JordanGuinn: Removed the `@NonNull` here so that I can actually create a report without having an id beforehand (so that inserts can be performed).  My understanding is that that annotation ensures the field is required on the entity, but does _not_ indicate requiredness of the underlying DB columns, so we should be good

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
```diff
@@ -0,0 +1,9 @@
+package gov.cdc.nbs.report.models;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+
+public record CreateFilterRequest(
```
JordanGuinn: Didn't see any additional fields that seemed necessary for report filter creation here, but v open to thoughts/suggestions

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportServiceTest.java
```diff
@@ -71,161 +75,287 @@ private Report mockReport(
...
-        .hasSize(1)
-        .allSatisfy(
-            filterConfig -> {
-              assertThat(filterConfig.reportFilterUid()).isEqualTo(3L);
-
-              assertThat(filterConfig.filterType().code()).isEqualTo("J_S01");
-            });
-    assertThat(config.advancedFilter().reportFilterUid()).isEqualTo(6L);
-  }
+  @Nested
```
JordanGuinn, line 87: Ended up experimenting with `@Nested` test groups here, as a means to hopefully better distinguish between the things being tested, and to be able to use `beforeEach` meaningfully across those groups.  Thoughts?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
mcmcgrath13: single/multi select and requiredness are also at play and should be accounted for

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterValueRequest.java
```diff
@@ -0,0 +1,4 @@
+package gov.cdc.nbs.report.models;
+
+public record CreateFilterValueRequest(
+    Integer sequenceNumber, String valueType, String operator, String valueTxt) {}
```
mcmcgrath13: (though, nb-for-this-pr-but-maybe-leave-a-comment-if-you-don't-want-to-address-now): these filter value requests/values should be pretty much the same as our `BasicFilterRequest` and `AdvancedFilterRequest` as we have values when we've hit run/export and go to the landing page to ask to save/save as. It should be the job of the api to translate those into the db table appropriately

mcmcgrath13: to that end, maybe it's simplest to leave the values off of the filter request directly and instead add a whole (optional) report request to the `CreateReportRequest`?

mcmcgrath13: or maybe that's what the save/save as endpoints take and then turn it into these formats... 🤔 

thoughts?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateReportRequest.java
```diff
@@ -0,0 +1,13 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+
+public record CreateReportRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long dataSourceId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long libraryId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String sectionCode,
+    String description,
+    String ownerId,
```
mcmcgrath13: needs `group` as well - need all the fields from https://www.figma.com/design/9xgVFGnzhK3uVuFZuAFnAg/Report-module---removing-SAS?node-id=16235-149927&m=dev

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
```diff
@@ -35,15 +41,6 @@ public ResponseEntity<String> handleFailedSerialization(HttpMessageNotReadableEx
     return new ResponseEntity<>(ex.getMessage(), HttpStatus.UNPROCESSABLE_ENTITY);
   }
 
-  /**
-   * Ensure the status code set on ResponseStatusExceptions is maintained through to the response,
-   * and is not overridden with 500 through our ExceptionHandler for the base Exception class.
```
mcmcgrath13, line 40: (q, nb): why don't we need this anymore?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -34,21 +25,73 @@
...
+                    new IllegalArgumentException(
+                        "No report library found for ID " + request.libraryId()));
+
+    Report newReport =
+        Report.builder()
+            .dataSource(dataSource)
+            .reportLibrary(reportLibrary)
+            .reportTitle(request.reportTitle())
+            .sectionCd(request.sectionCode())
+            .build();
```
mcmcgrath13: (nit, b): this looks incomplete vs the fields in the request

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -34,21 +25,73 @@
...
+
+    Report savedReport = reportRepository.save(newReport);
+
+    if (request.reportFilters() != null && !request.reportFilters().isEmpty()) {
+      createReportFilters(savedReport, request.reportFilters());
+    }
```
mcmcgrath13: (nit, b): this should be in the same transaction as above

mcmcgrath13 (changes requested):
thoughts:
* can we add an integration test or two using the cucumber stuff?
* how are the record statuses and update times and users and such all tracked? are there any central utilities for adding that info to writes? should there be?
* most of the columns below are indeed populated, but some look like they're imputed from the context - can you research those and make sure they're populated correctly?

Image

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: Is that not handled with the `@Transactional` annotation?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateReportRequest.java
JordanGuinn: ~I do not have access to that document~ Access granted, will review

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
JordanGuinn: I initially added this to handle throwing `ResponseStatusException` exceptions as part of the req param validation portion.  In this PR, I moved that equivalent handling to an `MethodArgumentNotValidException` exception handler ([the underlying exception Spring throws on failed validations](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/MethodArgumentNotValidException.html)) in the ControllerAdvice for this controller, where I'm able to just _return_ an equivalent response for those same kinds of errors.  So this exception handler is superfluous unless/until we start throwing `ResponseStatusException` exceptions elsewhere.

JordanGuinn:
> can we add an integration test or two using the cucumber stuff?

Can do

> how are the record statuses and update times and users and such all tracked? are there any central utilities for adding that info to writes? should there be?

Unclear, will investigate.  Should I just assume that all of those fields are important?

> most of the columns below are indeed populated, but some look like they're imputed from the context - can you research those and make sure they're populated correctly?

I don't know what you mean by `context` - are you referring to `@AuthenticationPrincipal`, like the general user data associated with the JWT the request has been authenticated with?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
JordanGuinn: I noticed those, but it wasn't clear to me how (or even if) they were meaningfully persisted API-side, and certainly not what fields they pertain to.  Will investigate.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterValueRequest.java
JordanGuinn: I think that makes sense - adding a dedicated prop to `CreateReportRequest` that represents the basic and/or advanced filters associated with the Report.  We could then extract that out into it's own dedicated model once we get to `save`/`save as` too.

mcmcgrath13:
> > how are the record statuses and update times and users and such all tracked? are there any central utilities for adding that info to writes? should there be?
> 
> Unclear, will investigate. Should I just assume that all of those fields are important?

I'd assume anything that is populated by 6 is potentially important

> > most of the columns below are indeed populated, but some look like they're imputed from the context - can you research those and make sure they're populated correctly?
> 
> I don't know what you mean by `context` - are you referring to `@AuthenticationPrincipal`, like the general user data associated with the JWT the request has been authenticated with?

`context` meaning general user+place we are in the workflow + derived information from _somewhere_ - in the english sense of the word vs a literal computer context

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -0,0 +1,45 @@
...
+  private ReportMapper() {}
+
+  public static Report fromCreateReportRequest(
+      CreateReportRequest request,
+      NbsUserDetails user,
+      ReportLibrary reportLibrary,
+      DataSource dataSource) {
+    LocalDateTime now = LocalDateTime.now();
+
+    return Report.builder()
```
emyl3: (❓, maybe blocking) Are we no longer setting the [`location` ](https://github.com/cdcent/NEDSSDev/blob/7b3deb35bc4b40db2143a01d4f3b8a0d055aee6a/source/gov/cdc/nedss/webapp/nbs/action/reportadmin/CreateReport.java#L80) and [`report_type_code`](https://github.com/cdcent/NEDSSDev/blob/7b3deb35bc4b40db2143a01d4f3b8a0d055aee6a/source/gov/cdc/nedss/webapp/nbs/action/reportadmin/CreateReport.java#L83) because they are specific to the SAS implementation of creating a new `Report`?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
emyl3: Is the single/multi-select/requiredness [being captured here](https://github.com/cdcent/NEDSSDev/blob/7b3deb35bc4b40db2143a01d4f3b8a0d055aee6a/source/gov/cdc/nedss/webapp/nbs/action/reportadmin/CreateReportFilter.java#L76-L104)? 🧐

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+                        "Filter code not found for type: " + filter.filterCodeUid()));
+
+    if (!filterCode.getFilterType().startsWith("BAS_")) {
+      throw new IllegalArgumentException("Cannot create basic filter from advanced filter");
+    }
+
+    int minValueCount;
+    int maxValueCount;
+
+    switch (filter.type()) {
```
emyl3: oops already posted this code reference in a comment above... but I am seeing other [combinations of min/max here](https://github.com/cdcent/NEDSSDev/blob/7b3deb35bc4b40db2143a01d4f3b8a0d055aee6a/source/gov/cdc/nedss/webapp/nbs/action/reportadmin/CreateReportFilter.java#L76-L104)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
mcmcgrath13: yes, the min/max count's currently capture the single/multiness like @emyl3 linked to

mcmcgrath13: the requiredness is persisted in the `Report_Filter_Validation` table

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
mcmcgrath13: If the report execution services 422's does that bubble back up still?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterTypeMapper.java
```diff
@@ -16,4 +18,19 @@ public static FilterType fromFilterCode(FilterCode filterCode) {
         filterCode.getFilterType(),
         filterCode.getFilterName());
   }
+
+  public static FilterCode toFilterCode(FilterType filterType) {
```
mcmcgrath13: (q, nb): do we need this? the filter code's are a static set of options, so we need the id to be able to insert references, but not sure when we'd need a full on filter code for inserts/updates/etc

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdminEditReportRequest.java
```diff
@@ -0,0 +1,12 @@
+package gov.cdc.nbs.report.models;
+
+import gov.cdc.nbs.report.ReportConstants;
+
+public record AdminEditReportRequest(
+    Long libraryId,
+    String reportTitle,
+    String sectionCode,
+    String description,
+    Long ownerId, //  TODO: Should an admin be able to edit ownerId?
```
mcmcgrath13: yes, they can

Image

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
```diff
@@ -0,0 +1,17 @@
+package gov.cdc.nbs.report.models;
+
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+
+public record CreateFilterRequest(
+    List<Long> columnUids, List<BasicFilter> basicFilters, AdvancedQuery advancedQuery) {
```
mcmcgrath13: what are these arguments doing in this case? a create filter request is agnostic to basic/advanced and to the list of columns (minus what was the list of columns to select from)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
```diff
@@ -0,0 +1,17 @@
...
+  public record BasicFilter(
```
mcmcgrath13: this is for all filters, not just basic

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateReportRequest.java
```diff
@@ -0,0 +1,14 @@
...
+
+public record CreateReportRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long dataSourceId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long libraryId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String sectionCode,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long ownerId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportConstants.ReportGroup group,
+    String description,
+    CreateFilterRequest filterRequest) {}
```
mcmcgrath13: (q,nb): should this be a list of create filter request?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SaveAsReportRequest.java
```diff
@@ -0,0 +1,11 @@
+package gov.cdc.nbs.report.models;
+
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+public record SaveAsReportRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String sectionCode,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportConstants.ReportGroup group,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) CreateFilterRequest filterRequest,
```
mcmcgrath13: the save as can't add/change the filters, just the values, so this would take a `ReportExecutionRequest` not a `CreateFilterRequest`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
```diff
@@ -0,0 +1,17 @@
...
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+          Long filterCodeUid, //  Pertains to FilterCode.ID
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long columnUid,
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+          Boolean includeNulls, //  Pertains to ReportFilterValidation
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportConstants.FilterType type,
```
mcmcgrath13: the filter type can be null as single/multi-select doesn't apply to all filters

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
```diff
@@ -0,0 +1,17 @@
...
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+
+public record CreateFilterRequest(
+    List<Long> columnUids, List<BasicFilter> basicFilters, AdvancedQuery advancedQuery) {
+  public record BasicFilter(
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+          Long filterCodeUid, //  Pertains to FilterCode.ID
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long columnUid,
```
mcmcgrath13: column can also be null as it doesn't apply to all filters (e.g. advanced where clause builder filter)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
```diff
@@ -0,0 +1,17 @@
...
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+          Boolean includeNulls, //  Pertains to ReportFilterValidation
```
mcmcgrath13: this is derived from the filter code, and won't be on the request itself

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -27,10 +22,53 @@ public class ReportController {
...
+  @PreAuthorize("hasAuthority('ADDREPORT-REPORTING')")
+  public ResponseEntity<ReportId> createReport(
+      @AuthenticationPrincipal NbsUserDetails user,
+      @Valid @RequestBody CreateReportRequest request) {
+    Report report = reportService.createReport(request, user);
+    return new ResponseEntity<>(report.getId(), HttpStatus.OK);
+  }
+
+  @PutMapping("/configuration/{reportUid}/{dataSourceUid}")
+  @PreAuthorize("hasAuthority('EDITREPORT-REPORTING')")
```
mcmcgrath13: (comment, b): these authorities don't look right

Image

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,82 @@
+package gov.cdc.nbs.report;
+
```
mcmcgrath13, line 2: @mcmcgrath13 continue here

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
JordanGuinn: Oooooo, good catch!  I completely missed that somehow - will update!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
JordanGuinn: Partially!  `location` seemed fully redundant since `Report` has a `ReportLibrary`, which has a `library_name` itself.  But `report_type_code` seemed fully specific to the SAS implementation, yeah - we could theoretically add a new Python type code, or reuse one of the existing SAS codes, but it's unclear to me what purpose that would serve rn.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
mcmcgrath13: for this put/post, since it's the admin one, probably just `REPORTADMIN-REPORTING`? the save/saveas will probably need something else and the get mapping probably also needs more involved logic...

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterTypeMapper.java
JordanGuinn: We do not, no - will move this back to `Draft`, this isn't fully ready

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+  private final FilterCodeRepository filterCodeRepository;
+
+  public ReportFilterBuilder(
+      DataSourceColumnRepository dataSourceColumnRepository,
+      FilterCodeRepository filterCodeRepository) {
+    this.dataSourceColumnRepository = dataSourceColumnRepository;
+    this.filterCodeRepository = filterCodeRepository;
+  }
+
+  public ReportFilter buildBasicReportFilter(
```
mcmcgrath13: (nit, nb): continued from above, I don't think we want to distinguish "basic" here, since at this level, a report filter is a report filter

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+        filterCodeRepository
+            .findById(filter.filterCodeUid())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "Filter code not found for type: " + filter.filterCodeUid()));
+
+    if (!filterCode.getFilterType().startsWith("BAS_")) {
+      throw new IllegalArgumentException("Cannot create basic filter from advanced filter");
+    }
```
mcmcgrath13: (nit, b): this shouldn't be an issue/validation here

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    switch (filter.type()) {
+      case ReportConstants.FilterType.MS -> {
+        minValueCount = 1;
+        maxValueCount = -1;
+      }
+      case ReportConstants.FilterType.SS -> {
+        minValueCount = 1;
+        maxValueCount = 1;
+      }
+      default -> throw new IllegalArgumentException("Unsupported filter type: " + filter.type());
```
mcmcgrath13: null is valid, not all filters need a selection (e.g. dates, where clause builder, bas_days, etc)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+      CreateFilterRequest.BasicFilter filter, Report report) {
+    Character statusCd = 'A';
+    LocalDateTime now = LocalDateTime.now();
+
+    DataSourceColumn dataSourceColumn =
+        dataSourceColumnRepository
+            .findById(filter.columnUid())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
```
mcmcgrath13: not all filters have a column, this is expected (e.g. where clause builder, bas days)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
JordanGuinn: Ohhhhhh, I misunderstood the expectation here.  I was trying to allow for the actual creation of the advanced filter _criteria_ (with all underlying `Filter_value` rows to be created) as part of this ticket, rather than just a `Report_Filter` with an `ADV_WCB` `Filter_code`.

Okay cool, that simplifies things a bunch - will update.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -34,19 +23,76 @@
...
+  public Report createReport(CreateReportRequest request, NbsUserDetails user) {
+    DataSource dataSource =
+        dataSourceRepository
+            .findById(request.dataSourceId())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "No data source found for ID " + request.dataSourceId()));
+
+    ReportLibrary reportLibrary =
```
mcmcgrath13: (q, nb): do we really need the whole object here? or would checking the id exists be sufficient?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -34,19 +23,76 @@
...
+
+      List<ReportFilter> basicFiltersToCreate =
+          request.filterRequest().basicFilters().stream()
+              .map(f -> reportFilterBuilder.buildBasicReportFilter(f, savedReport))
+              .toList();
+      reportFilters.addAll(basicFiltersToCreate);
+    }
+
+    if (request.filterRequest().advancedQuery() != null) {
+      //  TODO: Translate AdvancedQuery to ReportFilter; APP-505
```
mcmcgrath13: this should be this ticket - doesn't need to be different handling for admin side

apps/modernization-ui/src/libs/permission/permissions.ts
```diff
@@ -43,6 +43,7 @@ export const permissions = {
         },
         run: 'RUNREPORT-REPORTING',
         export: 'RUNREPORT-REPORTING',
+        add: 'ADDREPORT-REPORTING',
```
mcmcgrath13: (nit, b): fix this up with the permissions in general

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/CreateFilterRequest.java
JordanGuinn: Ah, right - this is the `_N` suffix business?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: The `dataSource` object itself is an expected parameter when creating a `Report`, both with a constructor and using the builder pattern.  Figured since we probably wanna verify the ID is a valid dataSource anyway, wouldn't be worth the trouble of trying to figure out how to get Jakarta to accept just `dataSourceId` as a parameter

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
JordanGuinn: `IllegalArgumentException` and `MethodArgumentNotValidException` are still mapped to the `422` status code in here, yes - this only impacted instances where we were explicitly throwing `ResponseStatusException` rather than returning it, which aren't doing anywhere anymore

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -0,0 +1,53 @@
...
+    Report.ReportBuilder builder = Report.builder();
+
+    if (existingReportId != null) {
+      builder = builder.id(existingReportId);
+    }
+
+    return builder
+        .dataSource(dataSource)
+        .addTime(now)
+        .addUserUid(user.getId())
```
mcmcgrath13: should these only change if it's a create and not an edit?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SaveAsReportRequest.java
```diff
@@ -0,0 +1,11 @@
+package gov.cdc.nbs.report.models;
+
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+public record SaveAsReportRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String sectionCode,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportConstants.ReportGroup group,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportExecutionRequest filterRequest,
```
mcmcgrath13:
```suggestion
    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportExecutionRequest executionRequest,
```
not specific to filters as it will also have columns and sort, so broaden the name?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+            .findById(filter.filterCodeUid())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "Filter code not found for UID: " + filter.filterCodeUid()));
+
+    Integer minValueCount = null;
+    Integer maxValueCount = null;
+
+    switch (filter.filterCodeUid().intValue()) {
```
mcmcgrath13: maybe add a comment about how this is directly mapped from NBS 6 logic (it's quite kludgey and not how we'd do it from scratch, so the historical note could be helpful for the future)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+            maxValueCount = 1;
+          }
+        }
+      }
+    }
+
+    ReportFilter.ReportFilterBuilder filterBuilder =
+        ReportFilter.builder()
+            .report(report)
+            .dataSourceColumn(dataSourceColumn)
```
mcmcgrath13: there won't always be a column, does this need a conditional like the min/ax values? (and/or why doesn't it?)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/UpsertFilterRequest.java
```diff
@@ -0,0 +1,11 @@
+package gov.cdc.nbs.report.models;
+
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+public record CreateFilterRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+        Long filterCodeUid, //  Pertains to FilterCode.ID
+    Long columnUid,
```
mcmcgrath13: (comment, b): in the case of an edit, the ID will be available and we'll probably need that so it's treated as an update instead of an insert (and also use that for the requiredness)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+
+    if (minValueCount != null) {
+      filterBuilder.minValueCnt(minValueCount);
+    }
+
+    if (maxValueCount != null) {
+      filterBuilder.maxValueCnt(maxValueCount);
+    }
+
+    if (filter.isRequired()) {
```
mcmcgrath13, line 62: (q, b): if a filter edit flips from required to not required, will this handle deleting/updating (whatever 6 did) the prior validation row?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+  private final FilterCodeRepository filterCodeRepository;
+
+  public ReportFilterBuilder(
+      DataSourceColumnRepository dataSourceColumnRepository,
+      FilterCodeRepository filterCodeRepository) {
+    this.dataSourceColumnRepository = dataSourceColumnRepository;
+    this.filterCodeRepository = filterCodeRepository;
+  }
+
+  public ReportFilter build(CreateFilterRequest filter, Report report) {
```
mcmcgrath13: (sugg, b): needs to handle updates

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -36,19 +29,80 @@
...
+        reportRepository.save(
+            ReportMapper.fromAdminReportRequest(
+                request, user, reportLibrary, dataSource, existingReportId));
+
+    if (!request.filterRequests().isEmpty()) {
+      List<ReportFilter> reportFilters =
+          request.filterRequests().stream()
+              .map(f -> reportFilterBuilder.build(f, savedReport))
+              .toList();
+
```
mcmcgrath13, line 97: (q, b): if a filter is deleted, will this clean out/deactivate/whatevr 6 did the no longer active filters associated with this report?

mcmcgrath13: and does that also need to cascade down to any filter values associated with those deleted filters?

mcmcgrath13: another update corner: what if a filter flips from multi-select to single select - what happens to the default values associated with that filter?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
JordanGuinn: Good call, will update

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: I was largely wondering about this - I was trying to keep it as-is so that we could reuse `AdminReportRequest` for both creates and edits per our conversation, but that's proving to be a bit of a pain.  Particularly if the edit requests aren't going to include the _entire_ set of filters associated with a given report, and/or will only include the fields to be edited?

JordanGuinn: Per Slack convo - they _will_ include the entire set of filters, as well as the full set of filters, so will update accordingly

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
JordanGuinn: Updated to handle that

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: > another update corner: what if a filter flips from multi-select to single select - what happens to the default values associated with that filter?

I don't think I understand the question here.  Though also, we'll first need to support adding filter values, which will take place during the work to add the `save` endpoint, correct?

mcmcgrath13: Yeah, this is thinking slightly ahead.

Say a report filter is currently multi-select and has saved multiple default values. The admin goes in and changes that filter's type from multi- to single-select. What happens to the default values?

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/Report.java
JordanGuinn: Jk, apparently there's no way to actually auto-generate this field, since it's not in and of itself a primary key, and thus not marked as `IDENTITY` nor has a sequence associated with it.

So gonna revert this to being required, and manually fetch the next available `report_uid` on creation as we do in 6 https://github.com/cdcent/NEDSSDev/blob/7b3deb35bc4b40db2143a01d4f3b8a0d055aee6a/source/gov/cdc/nedss/reportadmin/dao/ReportDAO.java#L55

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: Looks like in 6, the answer is nothing.  All of those filter values are left in place in the DB as they were, and then the UI just seems to pick the first one when you select the report to be run.  Certainly not great

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
mcmcgrath13: sorry, more-so when we call the python report execution service (via its API), does it's result status pass through appropriately still? (422 or 500 or whatever)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -0,0 +1,67 @@
...
+      DataSource dataSource,
+      ReportId existingReportId) {
+    LocalDateTime now = LocalDateTime.now();
+
+    Report.ReportBuilder builder = Report.builder();
+
+    if (existingReportId != null) {
+      builder.id(existingReportId);
+    } else {
+      builder.id(new ReportId(reportRepository.getNextReportId(), dataSource.getId()));
```
mcmcgrath13: (q, maybe-blocking): do we need to worry about race conditions across different transactions here?

-- after research --

looks like 6 does this helper thing, which gets/reserves the ID

Looks like the 7 equivalent is `IdGeneratorService`

Image

libs/database-entities/src/main/java/gov/cdc/nbs/repository/ReportRepository.java
```diff
@@ -3,5 +3,15 @@
...
-public interface ReportRepository extends JpaRepository<Report, ReportId> {}
+public interface ReportRepository extends JpaRepository<Report, ReportId> {
+  /**
+   * Generates the next available `report_uid` in the DB; to be used during report creation.<br>
+   * <br>
+   * Necessary because `report_uid` is not marked as an IDENTITY column, nor does it have a
+   * corresponding sequence, so we have to use a custom query.
+   */
+  @Query("SELECT (MAX(r.id.reportUid)+1) FROM Report r")
+  Long getNextReportId();
```
mcmcgrath13: (sugg, b): need to use the `IdGeneratorService` for this functionality instead

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -0,0 +1,67 @@
...
+      builder.addTime(now).addUserUid(user.getId());
+    }
+
+    ReportConstants.ReportGroup group =
```
mcmcgrath13: (q, nb): the fancy automagic json-and-things don't take care of this for us I take it?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -0,0 +1,67 @@
...
+        Arrays.stream(ReportConstants.ReportGroup.values())
+            .filter(g -> g.toString().equals(request.group()))
+            .findFirst()
+            .orElseThrow(
+                () -> new IllegalArgumentException("Invalid report group: " + request.group()));
+
+    return builder
+        .dataSource(dataSource)
+        .descTxt(request.description())
+        .isModifiableIndicator('N') // consistently "N" in DB, so just continuing that pattern
```
mcmcgrath13, line 43: (nit, b): the `filter_mode` also needs to be hardcoded to `'B'` it seems

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdminReportRequest.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+public record AdminReportRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @Positive Long dataSourceId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @Positive Long libraryId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String sectionCode,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull Long ownerId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String group,
```
mcmcgrath13: (q, nb): what happens if we make this a `ReportConstants.ReportGroup` and not a `String? Is that a thing?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SaveAsReportRequest.java
```diff
@@ -0,0 +1,11 @@
+package gov.cdc.nbs.report.models;
+
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+public record SaveAsReportRequest(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String sectionCode,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportConstants.ReportGroup group,
```
mcmcgrath13: (nit, nb-for-this-pr): add validation annotations

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/UpsertFilterRequest.java
```diff
@@ -0,0 +1,12 @@
...
+import gov.cdc.nbs.report.ReportConstants;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+public record UpsertFilterRequest(
+    Long id, // Update if present, create if not
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
+        Long filterCodeUid, //  Pertains to FilterCode.ID
+    Long columnUid,
+    ReportConstants.SelectType selectType,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean isRequired) {}
```
mcmcgrath13:
```suggestion
    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) boolean isRequired) {}
```
I think by making it `boolean` it gets coerced to false if it's null or something? this came up on another PR and @brick-green did some digging

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+    this.filterCodeRepository = filterCodeRepository;
+  }
+
+  //  Suppressing java:S131 ("switch" statements should have "default" clauses) because
+  // there isn't anything to actually do in the default case when switching against
+  // the filter request's filterCodeUid.
+  //  Suppressing java:S5411 ("Use a primitive boolean expression here.") because `isRequired`
+  // is a Boolean for API flexibility.
+  public ReportFilter build(UpsertFilterRequest filter, Report report) {
+    Character statusCd = 'A';
```
mcmcgrath13: (nit, nb): should/could this use the `Status.Active` enum?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdminReportRequest.java
JordanGuinn: That's what I tried initially, and it expected the actual referenced enum values, not the underlying string equivalents.  I think because I had to make the enum somewhat custom, since `Reporting Facility` has a space in it.

Can probably simplify it if we change the expected group prop to come in as all caps and pascal case or something, but unsure how we feel about that change UI-side

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+                    new IllegalArgumentException(
+                        "Filter code not found for UID: " + filter.filterCodeUid()));
+
+    Integer minValueCount = null;
+    Integer maxValueCount = null;
+
+    // Note that this logic mapping filterCodeUids to corresponding min value counts and
+    // max value counts was pulled **verbatim** from NBS 6, to maintain parity with 6 as
+    // closely as possible.
+    switch (filter.filterCodeUid().intValue()) {
```
mcmcgrath13: (nit, not-at-all-blocking): I think we're gonna want to also add functionality for going the other way to send single/multi on the api to the UI - maybe we want to start a util for dealing with this funky situation?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
JordanGuinn: Not unless we adjust the shape of the request slightly https://github.com/CDCgov/NEDSS-Modernization/pull/3238#discussion_r3326066097

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+            maxValueCount = 1;
+          }
+        }
+      }
+    }
+
+    ReportFilter.ReportFilterBuilder filterBuilder =
+        ReportFilter.builder().report(report).filterCode(filterCode).statusCd(statusCd);
+
+    if (minValueCount != null) {
```
mcmcgrath13: (q, nb): what happens if we just pass the null here (since it is a nullable column)? does jakarta get angry?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+
+    if (filter.isRequired()) {
+      filterBuilder.filterValidation(
+          ReportFilterValidation.builder()
+              .reportFilterInd('Y')
+              .statusCd(statusCd)
+              .statusTime(now)
+              .build());
+    } else {
+      filterBuilder.filterValidation(null);
```
mcmcgrath13, line 71: does this automagically delete an existing filter validator? (if so - could you add a comment on what is happening since it's a bit too implicit for readability purposes)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
mcmcgrath13: eeeks - though looks like we'll do the same on 7, so I guess this is a "for now, we live with it" situation

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/Report.java
```diff
@@ -20,7 +20,7 @@
 @Entity
 @Table(name = "Report", catalog = "NBS_ODSE")
 public class Report {
-  @NonNull @EmbeddedId private ReportId id;
+  @EmbeddedId @NonNull ReportId id;
```
mcmcgrath13: (q, nb): what was the `private` doing here and what does switching it get us?

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportFilter.java
```diff
@@ -44,7 +44,7 @@ public class ReportFilter {
   @OneToMany(mappedBy = "reportFilter", fetch = FetchType.LAZY)
   private List<FilterValue> filterValues;
 
-  @OneToOne(mappedBy = "reportFilter", fetch = FetchType.LAZY)
+  @OneToOne(mappedBy = "reportFilter", fetch = FetchType.LAZY, orphanRemoval = true)
```
mcmcgrath13, line 49: (nit, b): add a comment here about the orphan removal and what it's doing

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportFilterValidation.java
```diff
@@ -21,8 +18,8 @@ public class ReportFilterValidation {
   @Column(name = "report_filter_validation_uid", nullable = false)
   private Long id;
 
-  @NonNull @OneToOne(fetch = FetchType.LAZY)
-  @JoinColumn(name = "report_filter_uid")
+  @OneToOne(fetch = FetchType.LAZY)
+  @JoinColumn(name = "report_filter_uid", nullable = false)
```
mcmcgrath13, line 22: (q, nb): what's the difference between the annotation an the paramt ot he annotation?

testing/regression/cypress/step_definitions/reports-api/create.steps.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }).as("apiResponse");
+}
+
+When(/^I send a POST request to the create endpoint with a valid report$/, () => {
+  const validReport = {
+    dataSourceId: VALID_DATA_SOURCE_UID,
+    libraryId: VALID_REPORT_LIBRARY_ID,
+    reportTitle: "Test Report Title",
+    sectionCode: "Test Section Code",
```
mcmcgrath13:
```suggestion
    sectionCode: VALID_SECTION_CD,
```
(nit, nb): so we can send a request for real, set this to a real section

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
JordanGuinn: How did you land on that, for future reference?  In our conversation earlier we had discussed it being okay to exclude

testing/regression/cypress/step_definitions/reports-api/create.steps.js
mcmcgrath13: also, we should probably validate the section cd is good in the API since we are/were getting 200's with junk data?

testing/regression/cypress/step_definitions/reports-api/create.steps.js
```diff
@@ -0,0 +1,128 @@
...
+    ownerId: 0,
+    group: "Private",
+    filterRequests: [],
+    description: "Test Description",
+  };
+  
+  makeCreateRequest(validReport);
+});
+
+When(/^I send a POST request to the create endpoint with missing dataSourceId$/, () => {
```
mcmcgrath13, line 36: (nit, nb): should we also test invalid data source and library id's?

testing/regression/cypress/step_definitions/reports-api/edit.steps.js
```diff
@@ -0,0 +1,122 @@
...
+    url: `${Cypress.config().baseUrl}nbs/api/report/configuration/${VALID_REPORT_UID}/${VALID_DATA_SOURCE_UID}`,
+    body,
+    failOnStatusCode: false,
+    headers: {
+      "Content-Type": "application/json"
+    }
+  }).as("apiResponse");
+}
+
+When(/^I send a PUT request to the edit endpoint with a valid report$/, () => {
```
mcmcgrath13, line 19: (nit, not-at-all-blocking): this is maybe not worth the squeeze, but should we edit the report we created in the create steps? to make sure we don't mess too much with the state of the DB in cumulative e2e running - changing something we added feels much more safe than changing something part of the base golden db

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
mcmcgrath13: it probably is okay to exclude to be fair, but since we're hardcoding the modifiable, it's consistent to hardcode that in `ReportDt.java` in 6, when it validates the report it also sets a bunch of fields (consistent with what we're seeing in the DB)

image

libs/database-entities/src/main/java/gov/cdc/nbs/repository/ReportRepository.java
```diff
@@ -3,5 +3,15 @@
...
 
-public interface ReportRepository extends JpaRepository<Report, ReportId> {}
+public interface ReportRepository extends JpaRepository<Report, ReportId> {
+  /**
+   * Generates the next available `report_uid` in the DB; to be used during report creation.<br>
+   * <br>
+   * Necessary because `report_uid` is not marked as an IDENTITY column, nor does it have a
+   * corresponding sequence, so we have to use a custom query.
+   */
+  @Query("SELECT (MAX(r.id.reportUid)+1) FROM Report r")
```
brick-green: q, b: two questions:

- is there a possibility for a race condition if two people try to create a report at the same time?
- Will this work if there aren't any reports? I _think_ `MAX()` will return `null` in that case

The second question feels like a simple fix (maybe use COALESCE?) but the first may be trickier and a pretty unlikely edge case anyways as users aren't making reports that often. Should we even try to solve for it?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdminReportRequest.java
mcmcgrath13: The UI is also going to have to hardcode/use an enum of some sort, so if that's a list vs a mapping of pretty names to API names doesn't seem like a big deal to me. The generated code seems to deal with enums and forces the UI to use those (at least in some of the cases such as the `AdvancedQuery.combinator` setup) - maybe we copy that setup?

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportFilterValidation.java
JordanGuinn: `nullable = false` is for the underlying database column, whereas `@NonNull` is for the field on the Java entity.  I removed `@NonNull` because on report filter creation, there will be times where I don't have the actual report filter yet to set on the validation (because neither have been inserted), but I _am_ setting the to-be-created validation itself on the report during the build phase, so that underlying DB constraint should still be honored.

testing/regression/cypress/step_definitions/reports-api/create.steps.js
JordanGuinn: We can, though we're already doing something similar in the controller tests

testing/regression/cypress/step_definitions/reports-api/edit.steps.js
JordanGuinn: Yeah that shouldn't be too bad - will bundle all the CRUD tests into a single feature so we can run them meaningfully sequentially

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdminReportRequest.java
JordanGuinn: If we're okay with it in the UI, then works for me

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -0,0 +1,67 @@
...
+      builder.id(new ReportId(reportRepository.getNextReportId(), dataSource.getId()));
+      builder.addTime(now).addUserUid(user.getId());
+    }
+
+    ReportConstants.ReportGroup group =
+        Arrays.stream(ReportConstants.ReportGroup.values())
+            .filter(g -> g.toString().equals(request.group()))
+            .findFirst()
+            .orElseThrow(
+                () -> new IllegalArgumentException("Invalid report group: " + request.group()));
```
brick-green: sug, nb: Can this (maybe along with `reportGroupToDbChar()`) be moved over to the ReportGroup enum? That might mean it should be pulled out of the Constants file to it's own

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -15,5 +15,28 @@ public final class ReportConstants {
...
+    PRIVATE("Private"),
+    REPORTING_FACILITY("Reporting Facility"),
+    PUBLIC("Public"),
+    TEMPLATE("Template");
+
+    private final String text;
+
+    ReportGroup(final String text) {
+      this.text = text;
+    }
```
brick-green: sugg, nb: maybe add in the characters to the enum values? 

```
PRIVATE("Private", "P")
REPORTING_FACILITY("Reporting Facility", "R"),
...
```
Then you wouldn't have to convert between text and char

testing/regression/cypress/step_definitions/reports-api/create.steps.js
JordanGuinn: Is a `section_cd` considered good just if it's `1000`, `1001`, or `1002`?  Also would the UI want to send those underlying numbers, or rather the corresponding text that maps to them?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
```diff
@@ -0,0 +1,130 @@
+package gov.cdc.nbs.report;
+
+import gov.cdc.nbs.entity.odse.*;
```
brick-green: sugg, b: list out the imports instead of the wildcard?

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/Report.java
brick-green: q, nb: should this be fully reverted?

brick-green (commented):
a few Qs and Suggs

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/Report.java
JordanGuinn: Yep, should be removed from the diff now!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
JordanGuinn: Will give that a shot manually - this was either the result of IntelliJ trying to be smart, or our Spotless formatter, definitely not something I intentionally did myself

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
JordanGuinn: Oh interesting - I'm not 100% sure, but this shouldn't have an impact on that in either direction, only handles `ResponseStatusException`.

libs/database-entities/src/main/java/gov/cdc/nbs/repository/ReportRepository.java
JordanGuinn: @mcmcgrath13 pointed out there's [already an existing utility for this in 7](https://github.com/CDCgov/NEDSS-Modernization/blob/main/libs/id-generator/src/main/java/gov/cdc/nbs/id/IdGeneratorService.java), so I just ended up using that!

That doesn't necessarily address those questions specifically, but imo it does make this less of a problem that we wanna solve for as part of _this_ work.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
JordanGuinn: I ended up actually removing the custom `text` field entirely from the enum, as that was causing problems during serialization from the UI (so now just a standard enum).  So this is gone entirely now!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
JordanGuinn: I'm hesitant to do that only because of the troubles I've had during request serialization from the UI, and trying to map `group` to the correct enum value when said enum includes private fields 😬 It's why I ended up removing the custom `text` field entirely

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
JordanGuinn: Confirmed, it was an IntelliJ setting.  Though that does make me wonder if we have a preference on a potential upper limit there?  Like at what point do we prefer `*`, if ever?

brick-green: Fair question. I set my upper limit to 999 lol

But I can't see a good reason to use the wildcard. Maybe imposing a somewhat reasonable limit (like 20ish?) would be an indicator that too much is happening in a single file and that some refactoring in needed. What do you think?

testing/regression/cypress/step_definitions/reports-api/create.steps.js
mcmcgrath13: it's configurable based on the `Report_Section` table. The UI will send the code it maps to. It's a string even though it's a number 🤷

libs/database-entities/src/main/java/gov/cdc/nbs/repository/ReportRepository.java
mcmcgrath13: The Id Generation service exists (hopefully) to avoid these sorts of races. It dibs the ID for us and then we can use it knowing there won't be a clash

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdminReportRequest.java
```diff
@@ -0,0 +1,24 @@
...
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @Positive Long dataSourceId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @Positive Long libraryId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String reportTitle,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String sectionCode,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull Long ownerId,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull ReportConstants.ReportGroup group,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull List<UpsertFilterRequest> filterRequests,
+    String description) {
+  public AdminReportRequest {
+    if (!ReportConstants.SECTION_CODES.contains(sectionCode)) {
```
mcmcgrath13: (nit, b): this needs to be updated since section codes aren't constant

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFilterBuilder.java
JordanGuinn: Works for me!  Will bump mine to 20 - I suppose if you've got more than that many imports from a single location, maybe worth re-evaluating why that is 😁

testing/regression/cypress/step_definitions/reports-api/create.steps.js
```diff
@@ -0,0 +1,135 @@
+import { When, Then } from "@badeball/cypress-cucumber-preprocessor";
+
+const VALID_DATA_SOURCE_UID = 1;
+const VALID_REPORT_LIBRARY_ID = 10000001;
+const VALID_SECTION_CODE = "1000";
+
+export let NEW_REPORT_UID;
+export let NEW_DATA_SOURCE_UID;
```
mcmcgrath13: (nit, nb): we aren't creating a new data source, so can just use the valid one we already know about

mcmcgrath13 (approved):
🪇

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -158,4 +223,37 @@ public ResponseEntity<ReportResult> executeReport(ReportExecutionRequest request
...
+            .findById(request.libraryId())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "No report library found for ID " + request.libraryId()));
+
+    if (!reportSectionRepository.existsBySectionCd(request.sectionCode())) {
+      throw new IllegalArgumentException(
+          "No report section found for code " + request.sectionCode());
+    }
```
brick-green, line 270: q, nb: this validation feels a little out of place for the method `fetchReportMetadata()`. Are there other validations we could group together into a `valiate()` method?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -1,27 +1,21 @@
...
 
+import gov.cdc.nbs.authentication.NbsUserDetails;
 import gov.cdc.nbs.datasource.utils.DataSourceNameConfiguration;
 import gov.cdc.nbs.datasource.utils.DataSourceNameUtils;
-import gov.cdc.nbs.entity.odse.DataSourceColumn;
-import gov.cdc.nbs.entity.odse.DisplayColumn;
-import gov.cdc.nbs.entity.odse.Report;
-import gov.cdc.nbs.entity.odse.ReportId;
-import gov.cdc.nbs.entity.odse.ReportLibrary;
+import gov.cdc.nbs.entity.odse.*;
```
brick-green: wildcard

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -1,27 +1,21 @@
...
-import gov.cdc.nbs.report.models.BasicFilterConfiguration;
-import gov.cdc.nbs.report.models.Library;
-import gov.cdc.nbs.report.models.ReportColumn;
-import gov.cdc.nbs.report.models.ReportConfiguration;
-import gov.cdc.nbs.report.models.ReportDataSource;
-import gov.cdc.nbs.report.models.ReportExecutionRequest;
-import gov.cdc.nbs.report.models.ReportResult;
-import gov.cdc.nbs.report.models.ReportSpec;
+import gov.cdc.nbs.report.mappers.ReportMapper;
+import gov.cdc.nbs.report.models.*;
```
brick-green: wildcard

apps/modernization-api/src/test/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapperTest.java
```diff
@@ -3,12 +3,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
-import gov.cdc.nbs.entity.odse.DataSource;
-import gov.cdc.nbs.entity.odse.DataSourceColumn;
-import gov.cdc.nbs.entity.odse.FilterCode;
-import gov.cdc.nbs.entity.odse.Report;
-import gov.cdc.nbs.entity.odse.ReportFilter;
-import gov.cdc.nbs.entity.odse.ReportId;
+import gov.cdc.nbs.entity.odse.*;
```
brick-green: wildcard

apps/modernization-api/src/test/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapperTest.java
```diff
@@ -3,14 +3,7 @@
...
 
-import gov.cdc.nbs.entity.odse.DataSource;
-import gov.cdc.nbs.entity.odse.DataSourceColumn;
-import gov.cdc.nbs.entity.odse.FilterCode;
-import gov.cdc.nbs.entity.odse.FilterValue;
-import gov.cdc.nbs.entity.odse.Report;
-import gov.cdc.nbs.entity.odse.ReportFilter;
-import gov.cdc.nbs.entity.odse.ReportFilterValidation;
-import gov.cdc.nbs.entity.odse.ReportId;
+import gov.cdc.nbs.entity.odse.*;
```
brick-green: wildcard

brick-green (approved):
Good work here! Found a few more wildcards and had one question

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: No other validations as of right now - I can change this fn from `fetch` to `verify`, if that's helpful?

PR #3243, opened 2026-05-19
[APP-528] feat: add include nulls handling to basic filters

mcmcgrath13:
## Description

Add include null handling to basic filters

* Set it up as side by side fields so it visually looks contiguous with an aria label for screen reader continuity
* Styling doesn't exactly match design and it might not be trivial to do, but that's a later problem I think
* Fixed the internal server error log so we get the stack trace
* Fixed null `includeNulls` handling in the where clause builder

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-528


apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: (b) I believe we need this `basicFilterRequest.includeNulls() != null &&` here as well!

mcmcgrath13: nice catch! pushed

emyl3 (approved):
🚢

brick-green (approved):
✅ ✨

PR #3253, opened 2026-05-20
[APP-636] feat: add column select allowability and plumb through API

mcmcgrath13:
## Description

Add tracking of whether a library allows column selection.

In NBS 6, this is done via the setting of the `Custom` `type` when creating a report, however, in practice a library either supports column selection or does not, so moving this to the library makes more sense here. The `type` is otherwise unused in NBS 6, especially as we are no longer creating images/charts, just CSVs

Notes:
* Renamed some report config since the `report` was redundant since we're in a report config

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-636


apps/modernization-ui/src/apps/report/run/ReportConfigurationPage.tsx
```diff
@@ -39,14 +39,19 @@ const ReportConfigurationPage = ({
...
                     <Card id="advanced-filter" title="Advanced Filter" collapsible={true}>
-                        <AdvancedFilter filter={config.advancedFilter} columns={config.reportColumns} />
+                        <AdvancedFilter filter={config.advancedFilter} columns={config.columns} />
+                    </Card>
+                )}
+                {config.library.allowColumnSelection && (
+                    <Card id="advanced-filter" title="Advanced Filter" collapsible={true}>
+                        <AdvancedFilter filter={config.advancedFilter} columns={config.columns} />
                     </Card>
                 )}
```
mcmcgrath13:
```suggestion
```

krista-skylight (approved):
(q, nb) Overall, this looks good to me but I have a question. Why is the `#TempSasPrograms` table in `03_ODSE_Data_Report_Library_Init.sql` necessary? Can't we just directly insert into `Report_Library`?

emyl3 (approved):
Looks good to me!!

mcmcgrath13:
> (q, nb) Overall, this looks good to me but I have a question. Why is the `#TempSasPrograms` table in `03_ODSE_Data_Report_Library_Init.sql` necessary? Can't we just directly insert into `Report_Library`?

We could probably not use the temp table, but we need to both insert all those values and put in rows for any values we don't know about that are referenced in the `Report` table. This approach allows that in one query, vs two, but it definitely could be approached differently

PR #3256, opened 2026-05-20
APP-635 Allow column selection to pass through to potntl_dup_inv_sum library

krista-skylight:
## Description

Potential Duplicate Investigations report needs to have all columns. This PR is a fix of the `POTNTL_DUP_INV_SUM` library which I previously wrote. The crosswalking of column names happens elsewhere. We just need to make sure the library returns all of them.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-635?atlOrigin=eyJpIjoiYzQxYjE0Y2YzODRlNGJjZmI4ZTRmOGU4NDMwY2UxMjUiLCJwIjoiaiJ9)


apps/report-execution/src/libraries/potntl_dup_inv_sum.py
mcmcgrath13: I don't think these columns will still be availably named this given the mapping is a few CTE's back. Might need to figure out how to do the first three CTEs in one shot?

or inner join based on the unique row ID to get the original columns to where on?

could you add a test where the subset sql is something like

```
SELECT EVENT_DATE as [Event Date], PATIENT_LOCAL_ID as [Patient Local Id], <...other fields you need>
FROM <the data source>
```

and make sure things still work?

mcmcgrath13: another option is we add a mapping of column names to titles to the request specification and add it as a kwarg to the execute call so you can do a lookup as needed for the source=>target column name

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -83,18 +71,64 @@ def execute(
...
+        NOTIFICATION_LAST_UPDATED_USER,
+        INV_RPT_DT,
+        INV_START_DT,
+        CONFIRMATION_DT,
+        CONFIRMATION_METHOD,
+        HSPTL_ADMISSION_DT,
+        CURR_PROCESS_STATE,
+        PATIENT_COUNTY_CODE,
+        JURISDICTION_NM,
+        INIT_NND_NOT_DT
```
mcmcgrath13:
```suggestion
        *
```
We don't know what columns will be in here - it depends on the subset query

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
mcmcgrath13: could you check what happens to the order when an sort column and direction are included when running the report from 6?

krista-skylight:
@mcmcgrath13 @JordanGuinn @brick-green I have a concern with going the approach we discussed yesterday of limiting the column map to only the fields we want. I need `PATIENT_LOCAL_ID`, `DISEASE_CD`, AND `EVENT_DATE` in order to make the calculations in this library, so I always need those included in the subset query. Do you think it's best in that case to include all columns in the map and then find another way to filter? Or hard-code the inclusion of those 3 columns?

No matter what we decide I'll likely need some pairing/assistance with the implementation :pray:

mcmcgrath13:
> @mcmcgrath13 @JordanGuinn @brick-green I have a concern with going the approach we discussed yesterday of limiting the column map to only the fields we want. I need `PATIENT_LOCAL_ID`, `DISEASE_CD`, AND `EVENT_DATE` in order to make the calculations in this library, so I always need those included in the subset query. Do you think it's best in that case to include all columns in the map and then find another way to filter? Or hard-code the inclusion of those 3 columns?
> 
> No matter what we decide I'll likely need some pairing/assistance with the implementation 🙏

If the column isn't in the map, this library won't work, so I'd throw a validation error. But the approach outlined should work assuming that validation is done

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
krista-skylight: @mcmcgrath13  I think by default it is sorting by `PATIENT_LOCAL_ID` in ascending order, but there is no clear secondary sort order. How are sort column and direction passed through in the api? Do you think I will need to include a new argument in the `execute` function for that?

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,19 +7,27 @@ def execute(
...
+    ), 'Required columns are missing from column_map'
     # Only use default if days_value is None (not provided)
     # If days_value is 0, treat it as 0 (not default)
     # days_value = kwargs.get('days_value')
     if days_value is None:
         days_value = 3650
 
+    select_clause = ', '.join(
+        [f'd.{key} AS [{item}]' for key, item in column_map.items()]
+    )
```
mcmcgrath13:
```suggestion
    select_clause = ', '.join(
        [f'[{item}]' for item in column_map.values()]
    )
```
The subset query already does the renaming, so don't need to do that again (won't work)

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -63,10 +58,10 @@ def execute(
                 EVENT_DATE
             ) AS days_since_prev,
             DATEDIFF(day, 
-                EVENT_DATE,
+                [EVENT_DATE],
```
mcmcgrath13:
```suggestion
                [{column_map[EVENT_DATE]}],
```
and everywhere else for all the columns from the subset

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
mcmcgrath13: they will be part of the original sort of the subset query, so as long as you maintain that sort, you're good

mcmcgrath13: or do the same thing with it that sas does

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,90 @@ def execute(
...
     """Potential Duplicate Investigations.
 
     Identifies potential duplicate investigations for the same patient,
     with the same disease, within a user-specified number of days.
     """
+    missing_columns = []
+    for col in ['PATIENT_LOCAL_ID', 'DISEASE_CD', 'EVENT_DATE']:
+        if col not in column_map:
+            missing_columns.append(col)
+    if missing_columns:
```
mcmcgrath13:
```suggestion
    if len(missing_columns) > 0:
```

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -20,65 +21,108 @@ class TestIntegrationNbsSrDupInvLibrary:
...
+        base = {
+            'version': 1,
+            'is_export': True,
+            'is_builtin': True,
+            'report_title': 'Potential Duplicate Investigations',
+            'library_name': 'potntl_dup_inv_sum',
+            'data_source_name': '[RDB].[dbo].[INV_SUMM_DATAMART]',
+            'subset_query': self.subset_query,
+            'days_value': None,
+            'column_map': self.default_column_map.copy(),
```
mcmcgrath13: (q, nb): why do we need to copy it?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -20,65 +21,108 @@ class TestIntegrationNbsSrDupInvLibrary:
...
+            NOTIFICATION_SUBMITTER as [Notification Submitter],
+            NOTIFICATION_LAST_UPDATED_DATE as [Notification Last Updated Date],
+            NOTIFICATION_LAST_UPDATED_USER as [Notification Last Updated By User],
+            INV_RPT_DT as [Date of Report],
+            INV_START_DT as [Investigation Start Date],
+            CONFIRMATION_DT as [Confirmation Date],
+            HSPTL_ADMISSION_DT as [Hospital Admission Date],
+            CURR_PROCESS_STATE as [Investigation Current Process Stage]
+        FROM [RDB].[dbo].[INV_SUMM_DATAMART]
+    """
```
mcmcgrath13:
```suggestion
```

compute from the column map instead to match what should come over the requests

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -20,65 +21,108 @@ class TestIntegrationNbsSrDupInvLibrary:
...
+    def create_spec(self, **overrides):
+        """Create a report specification with defaults from the class properties."""
+        base = {
+            'version': 1,
+            'is_export': True,
+            'is_builtin': True,
+            'report_title': 'Potential Duplicate Investigations',
+            'library_name': 'potntl_dup_inv_sum',
+            'data_source_name': '[RDB].[dbo].[INV_SUMM_DATAMART]',
+            'subset_query': self.subset_query,
```
mcmcgrath13:
```suggestion
            'subset_query': '',
```

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -20,65 +21,108 @@ class TestIntegrationNbsSrDupInvLibrary:
...
+            'days_value': None,
+            'column_map': self.default_column_map.copy(),
+        }
+        base.update(overrides)
+        return ReportSpec.model_validate(base)
```
mcmcgrath13, line 54:
```suggestion
        if base.subset_query == '':
           base.subset_query = f"""
           SELECT {f'{key} as [{alias}]' for key, alias in base.column_map.items()}
           FROM [RDB].[dbo].[INV_SUMM_DATAMART]
           """
        return ReportSpec.model_validate(base)
```

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -156,40 +163,18 @@ def test_execute_report_with_negative_days_value(self):
...
-                'subset_query': """
-                SELECT * FROM [RDB].[dbo].[INV_SUMM_DATAMART] WHERE 1 = 0
-                """,
-                'days_value': 365,
-            }
-        )
+        # Take the base subset_query and add WHERE 1=0 to force empty results
+        empty_subset_query = f'{self.subset_query} WHERE 1 = 0'
+
+        report_spec = self.create_spec(days_value=365, subset_query=empty_subset_query)
```
mcmcgrath13:
```suggestion
        report_spec = self.create_spec(days_value=365)
        report_spec.subset_query += " WHERE 1 = 0"
```

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
krista-skylight: That was a caution since python lists and dicts are mutable, but not really necessary in this case since I'm not making changes to it.

krista-skylight:
@mcmcgrath13 FYI - Seems like the investigation id index fixed the sort order mismatch for the snapshot but now I'm blocked from comparison testing because of [this bug](https://skylight-hq.slack.com/archives/C09HXK6MKPE/p1780006009885619) which @JordanGuinn spent some time helping me out with but we are stumped.

mcmcgrath13:
> @mcmcgrath13 FYI - Seems like the investigation id index fixed the sort order mismatch for the snapshot but now I'm blocked from comparison testing because of [this bug](https://skylight-hq.slack.com/archives/C09HXK6MKPE/p1780006009885619) which @JordanGuinn spent some time helping me out with but we are stumped.

The default should kick in if you remove that filter from the report and then you can still test the library (just with a large days number value)

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -223,3 +143,39 @@ def test_execute_report_verify_no_single_events(self):
...
+        )
+
+        with pytest.raises(MissingColumnError) as exc_info:
+            execute_report(report_spec)
+
+        # Verify the error contains the missing column
+        assert 'DISEASE_CD' in str(exc_info.value)
+
+    def test_execute_report_column_ordering(self):
+        """Verify that the expected columns are consistently ordered as expected."""
```
adaball: suggestion: rename test as the test body doesn't test column ordering but rather tests whether or not expected columns are present.

adaball (approved):
Overall LGTM, one suggestion re: naming of a test.

:shipit:

krista-skylight:
@adaball I accidentally rerequested review from you when that was meant for @mcmcgrath13 !

krista-skylight:
@mcmcgrath13 I finally managed to redo e2e testing and have loaded new files to confluence! sort order and date formats are expectedly different, so I have noted those in conversion notes.

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,101 @@ def execute(
...
     """Potential Duplicate Investigations.
 
     Identifies potential duplicate investigations for the same patient,
     with the same disease, within a user-specified number of days.
+
+    Conversion notes:
+    * The way SAS sorts data depends on the session's encoding. This library
+    will always sort alphabetically by Patient Local ID, Disease Code, and
+    Event Date to ensure consistent results regardless of encoding.
+    * Dates are formatted MM/DD/YYY in this library
```
mcmcgrath13:
```suggestion
    * Dates defer to default formatting
```

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,101 @@ def execute(
...
+    select_clause = ', '.join([f'd.[{item[1]}]' for item in column_map])
+
     full_query = f"""
     WITH subset AS ({subset_query})
     -- Capture SQL Server's physical row order
     , source_order AS (
         SELECT 
             *,
-            ROW_NUMBER() OVER (ORDER BY (SELECT NULL)) AS sas_row_num
+            ROW_NUMBER() OVER (ORDER BY (SELECT NULL)) AS row_num
```
mcmcgrath13: (q,b): is this `row_num` still needed? could we delete this whole CTE and use `subset` directly?

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,101 @@ def execute(
...
     )
-    ORDER BY 
-        d.PATIENT_LOCAL_ID COLLATE Latin1_General_BIN,
-        d.DISEASE_CD COLLATE Latin1_General_BIN,
-        d.EVENT_DATE,
-        d.sas_row_num
+    ORDER BY
+        d.[{col_dict['PATIENT_LOCAL_ID']}],
+        d.[{col_dict['DISEASE_CD']}],
+        d.[{col_dict['EVENT_DATE']}]
```
mcmcgrath13: (nit, b): it looks like the sort from the end user _is_ applied to this library - if we make the first argument of the order by the row_num does that work? or does that also need to be plumbed through to the library spec?

Image

```suggestion
        d.[{col_dict['EVENT_DATE']}] desc
```

looks like the sas code sorts event date down instead of up

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -236,6 +237,17 @@ public ResponseEntity<ReportResult> executeReport(ReportExecutionRequest request
         .contentType(MediaType.APPLICATION_JSON)
         .body(reportSpec)
         .retrieve()
+        .onStatus(
```
mcmcgrath13, line 256: could you revert these error handling changes from this PR? I'm working on refining and testing in a separate PR

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,101 @@ def execute(
...
     Identifies potential duplicate investigations for the same patient,
     with the same disease, within a user-specified number of days.
+
+    Conversion notes:
+    * The way SAS sorts data depends on the session's encoding. This library
+    will always sort alphabetically by Patient Local ID, Disease Code, and
+    Event Date to ensure consistent results regardless of encoding.
+    * Dates are formatted MM/DD/YYY in this library
     """
+    # for easier lookups when order doesn't matter
```
mcmcgrath13, line 28: (sugg, b): This library requires column selection, so we should probably error if column_map is None

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
```diff
@@ -243,12 +233,37 @@ tables:
         data: random.choice(["Open Case", "Closed Case", "Field Follow-up", "Awaiting Interview", "Surveillance Follow-up", None])
         null_percentage: 0.20
       
-      - column_name: PATIENT_COUNTY_CODE
+      - column_name: PATIENT_COUNTY
```
mcmcgrath13, line 236: why were changes to the various columns needed in this PR? intentional or merge cruft?

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -223,3 +143,39 @@ def test_execute_report_verify_no_single_events(self):
...
+        report_spec = self.create_spec()
+
+        result = execute_report(report_spec)
+
+        expected_columns = {
+            'Patient Local Id',
+            'Disease Code',
+            'Event Date',
+            'Investigation Local Id',
+        }
```
mcmcgrath13: (sugg, b): can we check the order as well as the presence

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
krista-skylight: you are right, I can delete

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
krista-skylight: Does lines 28-33 not address that or do you mean I should raise a different error if column_map isn't passed through at all?

apps/report-execution/tests/integration/assets/tablefaker_schema/inv_summ_datamart.yaml
krista-skylight: IIRC I removed some columns after learning about the column selection feature because there are more columns in the INV_SUMM_DATAMART table than are in the available columns selection

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
krista-skylight: added here! https://github.com/CDCgov/NEDSS-Modernization/pull/3256/changes/d8cd978037d1a83bdfdc67be7234297bd9ea44b7

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,100 @@ def execute(
...
     with the same disease, within a user-specified number of days.
+
+    Conversion notes:
+    * The way SAS sorts data depends on the session's encoding. This library
+    will always sort alphabetically by Patient Local ID, Disease Code, and
+    Event Date to ensure consistent results regardless of encoding.
+    * Dates defer to default formatting
     """
+    if not column_map:
+        raise MissingColumnError(['PATIENT_LOCAL_ID', 'DISEASE_CD', 'EVENT_DATE'])
```
mcmcgrath13:
```suggestion
    required_cols = ['PATIENT_LOCAL_ID', 'DISEASE_CD', 'EVENT_DATE']
    if not column_map:
        raise MissingColumnError(required_cols)
```

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,100 @@ def execute(
...
+    # for easier lookups when order doesn't matter
+    col_dict = {m[0]: m[1] for m in column_map}
+
+    missing_columns = []
+    for col in ['PATIENT_LOCAL_ID', 'DISEASE_CD', 'EVENT_DATE']:
```
mcmcgrath13:
```suggestion
    for col in required_cols:
```

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,100 @@ def execute(
...
     # days_value = kwargs.get('days_value')
     if days_value is None:
         days_value = 3650
 
+    select_clause = ', '.join([f'd.[{item[1]}]' for item in column_map])
+
     full_query = f"""
     WITH subset AS ({subset_query})
     -- Capture SQL Server's physical row order
     , source_order AS (
```
mcmcgrath13:
```suggestion
```

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -7,108 +8,100 @@ def execute(
...
-            MMWR_YEAR,
-            NOTIFICATION_STATUS,
-            DISEASE_CD,
-            sas_row_num
-        FROM source_order
-        WHERE EVENT_DATE IS NOT NULL
-            AND PATIENT_LOCAL_ID IS NOT NULL
-            AND DISEASE_CD IS NOT NULL
+        SELECT *
+        FROM source_order
```
mcmcgrath13:
```suggestion
   , clean_data AS (
        SELECT *
        FROM subset 
```

mcmcgrath13: I think this suggestion is coming across weird due to github funkiness - the idea is to remove the use of `source_order` entirely and just directly reference `subset`

apps/report-execution/src/errors.py
```diff
@@ -18,6 +18,16 @@ def __init__(self, library_name: str, is_builtin: bool):
         super().__init__(message, 422)
 
 
+class MissingColumnError(BaseReportExecutionError):
+    """Required columns are missing."""
+
+    def __init__(self, missing_columns: list[str]):
+        message = f"""
+        Required columns {', '.join(missing_columns)} are missing from column_map
```
mcmcgrath13:
```suggestion
        Required columns {', '.join(missing_columns)} are missing from column selection
```

apps/report-execution/tests/integration/libraries/potntl_dup_inv_sum.py
```diff
@@ -223,3 +143,40 @@ def test_execute_report_verify_no_single_events(self):
...
+        result = execute_report(report_spec)
+
+        expected_columns = [
+            'Event Date',
+            'Patient Local Id',
+            'Disease Code',
+            'Investigation Local Id',
+        ]
+        missing_columns = set(expected_columns) - set(result.content.columns)
+        assert not missing_columns, f'Missing expected columns: {missing_columns}'
```
mcmcgrath13, line 179:
```suggestion
```
I think this is now redundant with the below check

mcmcgrath13 (commented):
some minor comments but overall looks good - the java styling and tests need to be fixed up before approval/merge

mcmcgrath13 (approved):
🗺️

PR #3264, opened 2026-05-27
[APP-534] BAS_DAYS filter value added to report spec

brick-green:
## Description

This PR maps the value of a BAS_DAYS filter to the daysValue field on report spec so that the value can be passed on to the report execution service.

This also moves the filter type constants from where clause service to `ReportConstants`

## Tickets

* [APP-534](https://cdc-nbs.atlassian.net/jira/software/c/projects/APP/boards/626?selectedIssue=APP-534)

[APP-534]: https://cdc-nbs.atlassian.net/browse/APP-534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportSpec.java
```diff
@@ -8,4 +8,5 @@ public record ReportSpec(
     @JsonProperty(value = "report_title", required = true) String reportTitle,
     @JsonProperty(value = "library_name", required = true) String libraryName,
     @JsonProperty(value = "data_source_name", required = true) String dataSourceName,
-    @JsonProperty(value = "subset_query", required = true) String subsetQuery) {}
+    @JsonProperty(value = "subset_query", required = true) String subsetQuery,
+    Integer daysValue) {}
```
mcmcgrath13: does this need the `@JsonProperty` annotation, or does it automatically get snake_case'd to match what python expects?

Image

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -83,10 +83,52 @@ public ReportSpec build() {
...
+                    filterConfig.filterType() != null
+                        && ReportConstants.BAS_DAYS.equals(filterConfig.filterType().type()))
+            .map(BasicFilterConfiguration::reportFilterUid)
+            .toList();
+
+    // Inspect the runtime execution request to find the user-submitted filter matching
+    // the targeted "BAS_DAYS" UIDs
+    String rawDaysValue =
+        reportExecRequest.basicFilters().stream()
+            .filter(request -> basDaysFilterUids.contains(request.reportFilterUid()))
```
mcmcgrath13: can these be combined into one filter then map?

mcmcgrath13: ahhh - I see it now - different streams

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -83,10 +83,52 @@ public ReportSpec build() {
...
+      return null;
+    }
+
+    // Parse the text value into an Int. Malformed entries will bubble up as an argument validation
+    // failure.
+    try {
+      return Integer.valueOf(rawDaysValue);
+    } catch (NumberFormatException e) {
+      throw new IllegalArgumentException(
+          "The requested filter value must be a valid integer: " + rawDaysValue);
```
mcmcgrath13: (nit, nb): specify the filter name instead of just "requested"

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportSpec.java
brick-green: definitely needed the `@JsonProperty`

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportSpecBuilderTest.java
```diff
@@ -307,6 +318,149 @@ void build_should_include_where_clause_when_present() {
...
+    when(request.basicFilters()).thenReturn(filterRequests);
+
+    DataSourceNameUtils dataSourceNameUtils = mockDataSourceNameUtils();
+
+    ReportSpecBuilder reportSpec =
+        new ReportSpecBuilder(request, reportConfig, dataSourceNameUtils, whereClauseService);
+
+    assertThatThrownBy(reportSpec::build)
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("The requested filter value must be a valid integer: not_number");
```
mcmcgrath13: this test is out of date

mcmcgrath13 (approved):
📆

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -83,10 +83,62 @@ public ReportSpec build() {
...
+    if (reportExecRequest.basicFilters() == null || reportConfig.basicFilters() == null) {
+      return null;
+    }
+
+    // Find the basic filter configuration for "BAS_DAYS" if it exists.
+    BasicFilterConfiguration basDaysConfig =
+        reportConfig.basicFilters().stream()
+            .filter(
+                filterConfig -> ReportConstants.BAS_DAYS.equals(filterConfig.filterType().type()))
+            .findFirst()
```
krista-skylight, line 107: (q, nb) why do we need to use `findFirst()`? does this parse as an array if the user puts in multiple integers with commas?

krista-skylight (approved):
have one question but otherwise lgtm!

JordanGuinn (approved):
✅

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
brick-green: the `basicFilters()` are a list then the `.filter()` will pare down to the `BAS_DAYS` which will only ever be a single filter. But the type is still `list` at that point. Since we know it will only be a single value in that list I just use the `findFirst()` to essentially convert from `List<BasicFilterConfiguration>` to `BasicFilterConfiguration`

PR #3267, opened 2026-05-29
[APP-368] feat: Update subset_sql to use advanced filters

emyl3:
## Description

- Translate the existing [advanced filters code from NEDSSDev](https://github.com/cdcent/NEDSSDev/blob/ef89cd6d83915ebe74e2e9900d419dd2175ed0b8/source/gov/cdc/nedss/report/util/ReportSQLBuilder.java#L416-L490) to work in mod
- Also includes [handling if there are columns from the RDB.Lab_Result_Val table](https://github.com/cdcent/NEDSSDev/blob/ef89cd6d83915ebe74e2e9900d419dd2175ed0b8/source/gov/cdc/nedss/report/util/ReportSQLBuilder.java#L499-L501)

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-368)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -26,24 +46,69 @@ public class WhereClauseService {
...
+    AdvancedQueryResult advancedQueryResult =
+        buildAdvancedQueryResult(reportConfig, executionRequest.advancedFilter());
+    String advWhereFragment = advancedQueryResult.query();
+    boolean hasLabResultVal = advancedQueryResult.hasLabResultVal();
+
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+
+    if (hasLabResultVal) {
+      basicWhereFragment = String.format("%s WHERE %s", labResultQuery, basicWhereFragment);
```
mcmcgrath13: (nit, b-ish): why change the basic Where Fragment here instead of creating a fragment and then adding it to the finalWhere independent of the basic where stuff - feels muddy that these are intermixed. We have three segments and each is built, check actually exists and added if it exists

mcmcgrath13: or is there something I'm missing in the logic of how these need to be combined?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdvancedQueryResult.java
```diff
@@ -0,0 +1,7 @@
+package gov.cdc.nbs.report.models;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+public record AdvancedQueryResult(
```
mcmcgrath13: (sugg, nb): This data structure seems unneeded to me - why not have the lab result checking be its own thing instead of folded in here? (or fully folded in and then we don't need to return both)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -12,13 +12,25 @@
...
 
+  /**
+   * Generates a string that contains the colName in brackets which protects against SQL reserved
+   * words
+   */
+  public String convertToSQLColName(String colName, String colType) {
+    String sqlColName = "[" + colName + "]";
+    if (colType.equals("DATETIME")) {
+      return "CAST(" + sqlColName + " AS DATE)";
+    }
```
mcmcgrath13, line 33: (q, b): why do this casting? can sql not deal with date/datetime comparisons natively? If it is needed can you add a comment on why?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -15,5 +16,37 @@ public final class ReportConstants {
...
+  public static final String NE = "NE"; // not equals
+  public static final String IN = "IN"; // is null
+  public static final String NN = "NN"; // not null
+  public static final String SW = "SW"; // starts with
+  public static final String CO = "CO"; // contains
+  public static final String BW = "BW"; // between
+  public static final String LT = "LT"; // less than
+  public static final String GT = "GT"; // greater than
+  public static final String LE = "LE"; // less than or equal to
+  public static final String GE = "GE"; // greater than or equal to
```
mcmcgrath13: (sugg, very-nb): would it be cleaner to make this an `Operator` enum instead?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -15,5 +16,37 @@ public final class ReportConstants {
...
+
+  public static final Set<String> RDB_LAB_RESULT_VAL_COLS =
+      Set.of(
+          "numeric_result_val",
+          "ref_range_frm",
+          "ref_range_to",
+          "coded_result_val",
+          "coded_result_val_desc",
+          "text_result_val",
+          "RESULT_UNITS");
```
mcmcgrath13: (q, nb): is the mixed case-ness real? and/or should we go all lower or upper for consistency?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
mcmcgrath13: I think I'm seeing the logic, but this still feels a bit roundabout since the basic and advanced need to be and'ed together no matter what (assuming existence), then the lab thing wraps all of that, so could we construct the full where, but then if the lab thing needs to be done, handle that at the very end (similar to 6)?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -95,27 +160,109 @@ public String buildBasicWhereFragment(
...
+
+      for (AdvancedQuery rule : ruleGroup.rules()) {
+        AdvancedQueryResult innerAdvQueryRes = buildAdvancedQuery(config, rule);
+        String innerRuleSql = innerAdvQueryRes.query();
+        if (!innerRuleSql.isEmpty()) {
+          joiner.add(innerRuleSql);
+        }
+        if (innerAdvQueryRes.hasLabResultVal()) {
+          containsLabResultValCol = true;
+        }
```
```suggestion
        containsLabResultValCol |= innerAdvQueryRes.hasLabResultVal();
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -95,27 +160,109 @@ public String buildBasicWhereFragment(
...
+      }
+
+      return new AdvancedQueryResult(joiner.toString(), containsLabResultValCol);
+    }
+
+    return new AdvancedQueryResult("", containsLabResultValCol);
```
mcmcgrath13: (q, nb): this is dead code? should we error instead?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -95,27 +160,109 @@ public String buildBasicWhereFragment(
...
+  }
+
+  private AdvancedQueryResult buildFormattedAdvancedCriteria(
+      ReportConfiguration config, AdvancedQuery.Rule rule) {
+    boolean hasLabResultValCol;
+
+    String operator = rule.operator().toUpperCase();
+    ReportColumn column =
+        findColumn(config, rule.columnId()).orElseThrow(IllegalArgumentException::new);
+    hasLabResultValCol = RDB_LAB_RESULT_VAL_COLS.contains(column.name());
```
mcmcgrath13: (q, b): is the case sensitive? (can we make sure it is a case-agnostic check)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -157,33 +304,58 @@ private String buildBasicFilterCriteria(
...
+  private String buildBasicFilterCriteria(
+      BasicFilterRequest basicFilterRequest, ReportColumn column) {
+    return buildFilterCriteria(
+        basicFilterRequest.values(), column, basicFilterRequest.includeNulls(), false);
+  }
+
+  private String buildAdvFilterCriteria(AdvancedQuery.Rule rule, ReportColumn column) {
+    List<String> values = Arrays.asList(rule.value().split("\\|"));
+    boolean isNNOperator = rule.operator().equals(NE);
+    return buildFilterCriteria(values, column, isNNOperator, isNNOperator);
```
```suggestion
    boolean isNEOperator = rule.operator().equals(NE);
    return buildFilterCriteria(values, column, isNEOperator, isNEOperator);
```

(q) why do we need to add the `or is null` clause for this case?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -193,12 +365,105 @@ private String buildBasicTimeRangeCriteria(
...
+   */
+  private String buildAdvLikeCriteria(AdvancedQuery.Rule rule, ReportColumn column) {
+    StringBuilder criteria = new StringBuilder("(");
+    boolean isContains = rule.operator().contains(CO);
+
+    return criteria
+        .append(fieldFormatter.convertToSQLColName(column.name(), column.sourceTypeCode()))
+        .append(" LIKE ")
+        .append(String.format("'%s", isContains ? "%" : ""))
+        .append(rule.value())
```
mcmcgrath13: (sugg, b): This needs to handle sql injection (unless I'm missing that's already handled?

mcmcgrath13: (`formatField` should take care of this if used here

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
emyl3: Updated to use a enum!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
emyl3: the mixed case-ness is not real 😅 I will switch this to all upper!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: I was trying to match the [behavior here](https://github.com/cdcent/NEDSSDev/blob/ef89cd6d83915ebe74e2e9900d419dd2175ed0b8/source/gov/cdc/nedss/report/util/ReportSQLBuilder.java#L611-L612)! 👀

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: changed it up a little to use LIKE CONCAT('%', value, '%') so I can use the formatField!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
mcmcgrath13: cool! I had assumed SQL would do this for us and this was due to sas dumbnes, but it turns out sql is also like this 😭

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -12,13 +12,25 @@
 public class FieldFormatter {
   public String formatField(String type, String value) {
     return switch (type.toUpperCase()) {
-      case "STRING" -> "'" + value.replace("'", "''") + "'";
+      case "STRING", "DATETIME" -> "'" + value.replace("'", "''") + "'";
       case "DATE" -> convertToSQLDate(value);
```
mcmcgrath13: (q, b): why convert `DATETIME` to string? (and is this making any assumptions based on the current mod implementation which hasn't dealt with date serialization/formatting yet as that will come for free with the BNS 7 components?)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -26,24 +36,69 @@ public class WhereClauseService {
 
   private final FieldFormatter fieldFormatter;
 
+  private static final String LAB_RESULT_QUERY_VAL =
+      "root_ordered_test_pntr IN (SELECT root_ordered_test_pntr FROM %s";
```
```suggestion
      "WHERE root_ordered_test_pntr IN (SELECT root_ordered_test_pntr FROM %s %s)";
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -26,24 +36,69 @@ public class WhereClauseService {
...
+    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
+    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+
+    if (hasLabResultVal) {
+      basicWhereFragment = String.format("%s WHERE %s", labResultQuery, basicWhereFragment);
+    }
     finalWhere.add(basicWhereFragment);
+    if (!advWhereFragment.isEmpty()) {
+      finalWhere.add(
+          String.format("%s%s", advancedQueryResult.query(), hasLabResultVal ? ")" : ""));
```
```suggestion
    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
    
    finalWhere.add(basicWhereFragment);
    finalWhere.add(advancedQueryReqult.query());

    if (hasLabResultVal) {
      String rdbDataSource = dataSourceNameUtils.buildDataSourceName("nbs_rdb.lab_test_report");
      return LAB_RESULT_QUERY_VAL.formatted(rdbDataSource, finalWhere.toString());
    }
```

or something - does this get us to the same place?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -26,24 +36,69 @@ public class WhereClauseService {
...
   public String buildWhereClause(
-      ReportConfiguration reportConfig, ReportExecutionRequest executionRequest) {
-
-    // StringJoiner provides the "WHERE " prefix and " AND " delimiters between filter statements
-    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+      ReportConfiguration reportConfig,
+      ReportExecutionRequest executionRequest,
+      DataSourceNameUtils dataSourceNameUtils) {
+    String rdbDataSource = dataSourceNameUtils.buildDataSourceName("nbs_rdb.lab_test_report");
+    String labResultQuery = String.format(LAB_RESULT_QUERY_VAL, rdbDataSource);
```
```suggestion
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
mcmcgrath13: same with the one above (but if this doesn't feel readable, feel free to ignore)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -95,27 +150,109 @@ public String buildBasicWhereFragment(
...
+  }
+
+  private AdvancedQueryResult buildFormattedAdvancedCriteria(
+      ReportConfiguration config, AdvancedQuery.Rule rule) {
+    boolean hasLabResultValCol;
+
+    Operator operator = Operator.valueOf(rule.operator().toUpperCase());
+    ReportColumn column =
+        findColumn(config, rule.columnId()).orElseThrow(IllegalArgumentException::new);
+    hasLabResultValCol = RDB_LAB_RESULT_VAL_COLS.contains(column.name().toUpperCase());
```
```suggestion
    boolean hasLabResultValCol = RDB_LAB_RESULT_VAL_COLS.contains(column.name().toUpperCase());
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -95,27 +150,109 @@ public String buildBasicWhereFragment(
...
+      return new AdvancedQueryResult(joiner.toString(), containsLabResultValCol);
+    }
+
+    throw new IllegalArgumentException("Invalid advanced filter");
+  }
+
+  private AdvancedQueryResult buildFormattedAdvancedCriteria(
+      ReportConfiguration config, AdvancedQuery.Rule rule) {
+    boolean hasLabResultValCol;
+
```
```suggestion
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -147,7 +284,7 @@ private String buildBasicFilterCriteria(
     if (hasValues) {
       criteria
           .append(colName)
-          .append(" IN (")
+          .append(String.format(" %sIN (", negateCriteria ? "NOT " : ""))
```
```suggestion
          .append(negateCriteria ? " NOT" : "")
          .append(" IN (")
```
minor readability suggestion

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -157,33 +294,58 @@ private String buildBasicFilterCriteria(
...
+  private String buildBetweenCriteria(
+      List<String> values, ReportColumn column, boolean includeNulls) {
     if (values.size() != 2 && !includeNulls) return "";
 
-    // Transform raw client values into type-formatted strings
-    List<String> formattedValues = fieldFormatter.convertToSQLFromDateRange(values);
+    String colType = column.sourceTypeCode();
+    List<String> formattedValues;
+
+    if (colType.equals("DATE")) {
```
mcmcgrath13: (q, nb): what's the difference between the date and datetmie column types in practice?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -157,33 +294,58 @@ private String buildBasicFilterCriteria(
...
+      formattedValues = fieldFormatter.convertToSQLFromDateRange(values);
+    } else {
+      formattedValues =
+          values.stream()
+              .filter(Objects::nonNull)
+              .map(v -> fieldFormatter.formatField(column.sourceTypeCode(), v))
```
```suggestion
              .map(v -> fieldFormatter.formatField(colType, v))
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -193,12 +355,105 @@ private String buildBasicTimeRangeCriteria(
...
+  private String buildNullCriteria(ReportColumn column, boolean negateCriteria) {
+    return String.format(
+        "%s IS %sNULL",
+        fieldFormatter.convertToSQLColName(column.name(), column.sourceTypeCode()),
+        negateCriteria ? "NOT " : "");
+  }
+
+  private String buildAdvNullCriteria(AdvancedQuery.Rule rule, ReportColumn column) {
+    boolean negateCriteria = Operator.valueOf(rule.operator()).equals(Operator.NN);
+    return "(" + buildNullCriteria(column, negateCriteria) + ")";
```
mcmcgrath13, line 375: (q, nb): why parenthesize here?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -193,12 +355,105 @@ private String buildBasicTimeRangeCriteria(
...
+   * @return a SQL fragment.
+   */
+  private String buildAdvLikeCriteria(AdvancedQuery.Rule rule, ReportColumn column) {
+    StringBuilder criteria = new StringBuilder("(");
+    boolean isContains = Operator.valueOf(rule.operator()).equals(Operator.CO);
+
+    return criteria
+        .append(fieldFormatter.convertToSQLColName(column.name(), column.sourceTypeCode()))
+        .append(" LIKE CONCAT(")
+        .append(String.format("'%s', ", isContains ? "%" : ""))
```
```suggestion
        .append(" LIKE CONCAT('")
        .append(isContains ? "%" : "")
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -193,12 +355,105 @@ private String buildBasicTimeRangeCriteria(
...
+        .append(fieldFormatter.formatField(column.sourceTypeCode(), rule.value()))
+        .append(", '%')")
+        .append(")")
```
mcmcgrath13, line 400: (q, nb): do we need these outer parens?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
```diff
@@ -408,4 +463,203 @@ void should_throw_exception_when_input_values_result_in_zero_formatted_values()
...
+        Arguments.of(
+            List.of(18, 19, 20, 21),
+            "(([COLUMN_STRING] LIKE CONCAT('', 'foo', '%')) AND ([COLUMN_STRING] LIKE CONCAT('%', 'foo', '%')) AND ([COLUMN_STRING] IN ('foo')) AND ([COLUMN_STRING] NOT IN ('foo') OR [COLUMN_STRING] IS NULL))"),
+        Arguments.of(List.of(), ""),
+        Arguments.of(null, ""));
+  }
+
+  private List<AdvancedQuery.Rule> create_adv_query_rules() {
+    String intNumVal = "1";
+    String dateVal = "2026-05-28";
```
mcmcgrath13:
```suggestion
    String dateVal = "05/28/2026";
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdvancedQueryResult.java
emyl3: Removed!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: No longer relevant since implementation has changed! 😅

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
emyl3: OOO yikes... yes I was converting it to string since I saw that the rule was being returned from the frontend as 
```
{
  "id": "12bf0cd2-9d36-4f94-b9f1-ec44be676270",
   "operator": "LT",
   "columnId": 3,
    "value": "2026-05-28"
}
```

So when I switch it to the NBS 7 components we should be getting a timestamp or a date like https://github.com/CDCgov/NEDSS-Modernization/pull/3267#discussion_r3352043201?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: Updated!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: Yes! Took most of this in my latest commit! Thank you!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: Yes! Because the first parens on line 422 closes the CONCAT's parens that opens on 389 
Then, the outer parens on 423, we need to close the parens opened on line 384. Hope that makes sense 😅

mcmcgrath13: it does! my question is a little more philosophical - it seems like this generates statements of the shape `([field] LIKE CONCAT('%foo%'))` and I'm wondering why not have `[field] LIKE CONCAT('%foo%')` since it is semantically the same

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
mcmcgrath13: yup! which is how dates are stored in NBS 6 today and will continue that way

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -26,25 +36,68 @@ public class WhereClauseService {
...
 
     String basicWhereFragment =
         buildBasicWhereFragment(reportConfig, executionRequest.basicFilters());
 
+    String advWhereFragment =
+        buildAdvancedQueryResult(reportConfig, executionRequest.advancedFilter());
+
+    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+
     finalWhere.add(basicWhereFragment);
```
mcmcgrath13: (sugg/q, b): does this need the same non-null/empty check as the advanced filter?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: used this with slight modification -- thank you!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: It might be just me but with all the AND and ORs and for super nested scenarios it made it easier for me to read but I can remove it if it adds clutter or confusion for others!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: Ooo yes! good idea! Thank you!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
emyl3: This was to make sure if we are checking a column that is a datetime and we are given a date the different operators still work... 

For example, if we are checking the `add_time` column that is a `datetime` with a value of `2016-07-20 23:14:53.647` only after casting it would return something if it equaled the date provided of `07/20/2016' etc... I hope that makes sense. I can leave a more succinct comment in the code 😅

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: Updated!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: I am honestly not sure... I just saw the new type when translating it from the original code in 6 😅

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: Same reason as https://github.com/CDCgov/NEDSS-Modernization/pull/3267#discussion_r3356795829 😅 Let me know what you think... 👀

emyl3:
@mcmcgrath13 Addressed comments -- let me know what you think and I can make changes based on that! Thank you!!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
mcmcgrath13: I can argue it either way and if it helped your reading, that's good enough for me!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
mcmcgrath13: that makes sense - comment please!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
mcmcgrath13:
```suggestion
      case "STRING" -> "'" + value.replace("'", "''") + "'";
case "DATE", "DATETIME" -> convertToSQLDate(value);
```

I think?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -157,33 +271,58 @@ private String buildBasicFilterCriteria(
...
+      List<String> values, ReportColumn column, boolean includeNulls) {
     if (values.size() != 2 && !includeNulls) return "";
 
-    // Transform raw client values into type-formatted strings
-    List<String> formattedValues = fieldFormatter.convertToSQLFromDateRange(values);
+    String colType = column.sourceTypeCode();
+    List<String> formattedValues;
+
+    if (colType.equals("DATE")) {
+      formattedValues = fieldFormatter.convertToSQLFromDateRange(values);
```
mcmcgrath13: (q, b): should the `DATETIME` values also go through this path? if not, why not? (and comment in the code please 🙏 )

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -157,33 +271,58 @@ private String buildBasicFilterCriteria(
...
+          values.stream()
+              .filter(Objects::nonNull)
+              .map(v -> fieldFormatter.formatField(colType, v))
+              .toList();
+    }
 
     StringBuilder criteria = new StringBuilder("(");
 
-    String colName = "[" + column.name() + "]";
+    String colName = fieldFormatter.convertToSQLColName(column.name(), colType);
```
mcmcgrath13, line 325: (nit, super-nb): would it make sense to add a `formatToSQL` method on the column object? or is it maybe cleaner to keep with field formatting 🤔

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
emyl3: ooo yes! that makes more sense and makes all the date handling consistent!!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: OOO yes good catch!!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
emyl3: I left it in that other file in case it needs to be used outside of the where clause service

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
emyl3: Done!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
emyl3: Updated!!

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
```diff
@@ -414,4 +467,201 @@ void should_throw_exception_when_input_values_result_in_zero_formatted_values()
...
+        Arrays.asList(
+            Map.of("columnId", 2L, "operator", Operator.EQ, "value", intNumVal),
+            Map.of("columnId", 2L, "operator", Operator.NE, "value", intNumVal),
+            Map.of("columnId", 2L, "operator", Operator.GT, "value", intNumVal),
+            Map.of("columnId", 2L, "operator", Operator.LT, "value", intNumVal),
+            Map.of("columnId", 2L, "operator", Operator.GE, "value", intNumVal),
+            Map.of("columnId", 2L, "operator", Operator.LE, "value", intNumVal),
+            Map.of("columnId", 2L, "operator", Operator.NN, "value", ""),
+            Map.of("columnId", 2L, "operator", Operator.IN, "value", ""),
+            Map.of("columnId", 2L, "operator", Operator.BW, "value", "1,2"),
```
mcmcgrath13, line 605: can we add coverage of between with date vals?

mcmcgrath13 (approved):
minor coverage nit, but otherwise looks good!

apps/modernization-api/src/test/java/gov/cdc/nbs/report/WhereClauseServiceTest.java
emyl3: I have [one here](https://github.com/CDCgov/NEDSS-Modernization/pull/3267/changes/BASE..83133752468bae784e2286b33e21697c65210def#diff-61dcbd7538d2047502938ebc8906f850301529ec8a89e4029d61e34e54991404R584) with [this rule](https://github.com/CDCgov/NEDSS-Modernization/pull/3267/changes/BASE..83133752468bae784e2286b33e21697c65210def#diff-61dcbd7538d2047502938ebc8906f850301529ec8a89e4029d61e34e54991404R610) 👀

mcmcgrath13: I see it now... 🤦

emyl3: hahah no worries there are A LOT of rule combinations included 🫠 😅

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
JordanGuinn: Is this to say that we'll _only_ be supporting date filtering, and will _always_ disregard the time portion of any datetime query param potentially provided?  If so, maybe worth noting that we're doing so intentionally, and that `DATETIME` filtering itself is not supported

JordanGuinn (approved):
🔥 🚀

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
emyl3: In the UI, it doesn't look like we are allowing the option to filter by datetime at all since a user can only enter a date. I'll add that extra note to the comment.

PR #3268, opened 2026-05-29
APP-675 New params for report API

krista-skylight:
## Description

Implement a new JSON-ey object that will keep track of parameters like report days for libraries like QA07.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-675?atlOrigin=eyJpIjoiN2Q0MGFmZTQxNmQ5NDcyZWE5NDI1OGUzN2UwMDAwNDkiLCJwIjoiaiJ9)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportConfiguration.java
```diff
@@ -1,22 +1,33 @@
...
 public record ReportConfiguration(
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportDataSource dataSource,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Library library,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String title,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED)
-        List<BasicFilterConfiguration> basicFilters,
+    ReportDataSource dataSource,
+    Library library,
+    String title,
+    List<BasicFilterConfiguration> basicFilters,
```
mcmcgrath13: this file shouldn't need to be changed at all - revert

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportConfigurationOptions.java
```diff
@@ -0,0 +1,21 @@
+package gov.cdc.nbs.report.models;
+
+import java.util.List;
+
+public class ReportConfigurationOptions {
+    private List<Long> defaultColumnUids;
+    private Integer reportDays;
+
+    public ReportConfigurationOptions() {}
```
mcmcgrath13: revert this file as well - don't need to do anything with the column uids and we don't want to be specific about the params in the java

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Library.java
```diff
@@ -2,18 +2,21 @@
...
 import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.Map;
 
 public record Library(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String runner,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String libraryName,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean isBuiltin,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean allowColumnSelection) {
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean allowColumnSelection,
+    @Schema(description = "Optional JSON parameters for the report") Map<String, Object> reportParams) {
```
mcmcgrath13: The java doesn't need to know anything about these params, so I'd probably leave this as a string for simplicity and let the python parse it into a dict

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportSpec.java
```diff
@@ -9,4 +10,5 @@ public record ReportSpec(
     @JsonProperty(value = "library_name", required = true) String libraryName,
     @JsonProperty(value = "data_source_name", required = true) String dataSourceName,
     @JsonProperty(value = "subset_query", required = true) String subsetQuery,
-    @JsonProperty(value = "days_value") Integer daysValue) {}
+    @JsonProperty(value = "days_value") Integer daysValue,
+    @JsonProperty(value = "report_params") Map<String, Object> reportParams) {}
```
mcmcgrath13:
```suggestion
    @JsonProperty(value = "report_params") String reportParams) {}
```

apps/modernization-api/src/main/resources/db/report/execution/01_ODSE_Structure_Report_Library_Create.sql
```diff
@@ -17,12 +17,15 @@ BEGIN
...
+        is_builtin_ind          char(1)         NOT NULL,
+        report_params           NVARCHAR(MAX)   NULL,              
         add_time                datetime        NOT NULL,
         add_user_id             bigint          NOT NULL,
         last_chg_time           datetime        NOT NULL,
         last_chg_user_id        bigint          NOT NULL,
 
         CONSTRAINT PK_Report_Library PRIMARY KEY (library_uid)
+        CONSTRAINT CK_Report_Library_ReportParams_IsJSON
+            CHECK (ISJSON(report_params) = 1 OR report_params IS NULL)
```
mcmcgrath13:
```suggestion
```

nice find! but unfortunately it was introduced to TSQL too recently so we can't use it

apps/modernization-api/src/main/resources/db/report/execution/01_ODSE_Structure_Report_Library_Create.sql
```diff
@@ -17,12 +17,15 @@ BEGIN
         desc_txt                varchar(300)    NOT NULL,
         runner                  varchar(10)     NOT NULL,
         column_select_ind       char(1)         NOT NULL,
-        is_builtin_ind          char(1)         NOT NULL,              
+        is_builtin_ind          char(1)         NOT NULL,
+        report_params           NVARCHAR(MAX)   NULL,
```
mcmcgrath13:
```suggestion
        library_params           nvarchar(MAX)   NULL,              
```
These params are really at the library level and not the report level - maybe this is clearer?

apps/report-execution/src/models.py
```diff
@@ -17,6 +17,7 @@ class ReportSpec(BaseModel):
     data_source_name: str = Field(min_length=1)
     subset_query: str = Field(min_length=1)
     days_value: int | None = None  # Specific to potntl_dup_inv_sum
+    report_params: dict[str, Any] | None = None
```
```suggestion
    library_params: Json[Any] | None = pydantic.Field(default_factory=dict)
```

This should parse the JSON for us

and add an import for `from pydantic.types import Json`

mcmcgrath13: (untested - gotten from https://github.com/pydantic/pydantic/discussions/8446)

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/JsonToMapConverter.java
```diff
@@ -0,0 +1,34 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.Converter;
+
+import java.util.Map;
+
+@Converter(autoApply = true)
+public class JsonToMapConverter implements AttributeConverter<Map<String, Object>, String> {
+
+    private static final ObjectMapper mapper = new ObjectMapper();
```
mcmcgrath13: can delete this once you just leave the JSON a string in Java

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportLibrary.java
```diff
@@ -38,6 +40,10 @@ public class ReportLibrary {
   @NonNull @Column(name = "is_builtin_ind", length = 1, nullable = false)
   private Character isBuiltinInd;
 
+  @Column(name = "report_params", columnDefinition = "NVARCHAR(MAX)", nullable = true)
+  @Convert(converter = JsonToMapConverter.class)
+      private Map<String, Object> reportParams;
```
mcmcgrath13:
```suggestion
  @Column(name = "library_params", columnDefinition = "NVARCHAR(MAX)", nullable = true)
      private String libraryParams;
```

apps/modernization-api/src/main/resources/db/report/execution/01_ODSE_Structure_Report_Library_Create.sql
mcmcgrath13: also in practice, we don't expect these params to be huge - is there any cost to doing the nvarchar max instead of a more limited number? also why nvarchar instead of varchar like the other fields?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/Library.java
```diff
@@ -9,7 +9,8 @@ public record Library(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String name,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String description,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean isBuiltin,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean allowColumnSelection) {
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Boolean allowColumnSelection,
+    @Schema(description = "Optional JSON parameters for the report") String libraryParams) {
```
adaball:
```suggestion
    @Schema(
            requiredMode = Schema.RequiredMode.NOT_REQUIRED,
            description = "Optional JSON parameters for the report")
        String libraryParams) {
```

**suggestion:** label `libraryParams` as `NOT_REQUIRED`.

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportLibrary.java
```diff
@@ -38,6 +38,9 @@ public class ReportLibrary {
   @NonNull @Column(name = "is_builtin_ind", length = 1, nullable = false)
   private Character isBuiltinInd;
 
+  @Column(name = "library_params", length = 300, nullable = true)
```
adaball:
```suggestion
  @Column(name = "library_params", length = 300)
```
**nit:** `nullable` is already set to `true` by default.

apps/modernization-api/src/main/resources/db/report/execution/01_ODSE_Structure_Report_Library_Create.sql
```diff
@@ -17,7 +17,8 @@ BEGIN
         desc_txt                varchar(300)    NOT NULL,
         runner                  varchar(10)     NOT NULL,
         column_select_ind       char(1)         NOT NULL,
-        is_builtin_ind          char(1)         NOT NULL,              
+        is_builtin_ind          char(1)         NOT NULL,
+        library_params          varchar(300)    NULL,
```
adaball, line 21: **issue:** ‏Your addition to this migration will not be executed on databases that are already initialized by `modernization-api` because the table already exists.

Even though Liquibase is set to re-run migrations when changes are detected (as per the [runOnChange attribute](https://docs.liquibase.com/secure/reference-guide-5-1-1/changelog-attributes/runonchange)):

1. The `CREATE TABLE` statement is gated by an if statement checking if it exists already
2. Even if this gate wasn't there this migration would fail given that the table already exists

You'll likely need to create a new migration in the `apps/modernization-api/src/main/resources/db/report/execution` directory with the contents looking something like:

```sql
USE [NBS_ODSE]

ALTER TABLE [dbo].[Report_Library]
ADD library_params varchar(300) NULL;
```

I tested this locally and was able to confirm that your change did not take when running the Liquibase migrations.  I also tested a new migration with the above query and it worked.

apps/report-execution/pyproject.toml
```diff
@@ -6,7 +6,7 @@ readme = "README.md"
 requires-python = ">=3.14"
 dependencies = [
     "fastapi[standard]~=0.136",
-    "mssql-python>=1.8.0",
+    "mssql-python>=1.7.1",
```
adaball, line 9: **question:** should we ‏re-version the `mssql-python` library just because MacOS dev is broken?  That might be the best option, but it should likely be in a dedicated PR along with a regenerated `uv.lock` file.

adaball (changes requested):

apps/modernization-api/src/main/resources/db/report/execution/01_ODSE_Structure_Report_Library_Create.sql
mcmcgrath13: @adaball in general we treat the migrations as atomic per release, so even though this won't work for dev's out of the box, we want it to be clear/minimal-number-of-files for the users who adopt it. The expectation is devs just blow up their db and start over

mcmcgrath13 (commented):
generally looking good! can you add some python test(s) and pass on the parsed library params to execute?

adaball (approved):
Looks like Sonar is failing, but this LGTM :shipit:

mcmcgrath13 (approved):
📚 

(assuming sonar and all checks are happy pre-merge)

PR #3269, opened 2026-06-01
[APP-663] Standarize CSV date and datetime formatting

adaball:
## Description

This PR adds logic to standardize formatting for Python date and datetime instances when outputting to CSV.  The defaults for each are as follows:
 - datetime: `mm/dd/yyyy hh:mm:ss`
 - date: `mm/dd/yyyy`

These are changeable at runtime using the `REPORT_EXPORT_DATE_FORMAT` and `REPORT_EXPORT_DATETIME_FORMAT` env vars.

This PR also updates the following libraries to use Date and/or Datetime instances instead of custom formatting in their SQL:
- nbs_sr_08
- qa_01
- qa_03

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-663)


apps/report-execution/src/models.py
```diff
@@ -47,12 +50,43 @@ def get_unique_column(self, col_name: str) -> list[Any]:
...
+    Standardizes Python date and datetime instances to the provided strftime constants,
+    defaulting to:
+     - datetime: mm/dd/yyyy hh:mm:ss
+     - date: mm/dd/yyyy
+    """
+    # strftime constants for Python date and datetime when outputting to CSV
+    csv_date_strftime = get_str_env_or_default('CSV_DATE_STRFTIME', '%m/%d/%Y')
+    csv_datetime_strftime = get_str_env_or_default(
+        'CSV_DATETIME_STRFTIME', '%m/%d/%Y %H:%M:%S'
+    )
```
mcmcgrath13: (nit, b): instead of making this CSV can we name these a bit more specifically and generically something like `REPORT_EXPORT_DATE_FORMAT` and `REPORT_EXPORT_DATETIME_FORMAT`

apps/report-execution/src/models.py
```diff
@@ -47,12 +50,43 @@ def get_unique_column(self, col_name: str) -> list[Any]:
...
+
+    # update table data to have properly formatted dates and datetimes
+    table.data = [map(convert_dates, tpl) for tpl in table.data]
+
     # Short cut to valid CSV - can swap out later if performance dictates
     # or serialize to CSV at a different location
-    df = DataFrame.from_records(table.data, columns=table.columns, coerce_float=True)
-    # remove trailing new line
-    return df.to_csv(index=False, float_format='{:.20g}', lineterminator='\r\n')[:-2]
+    df = pd.DataFrame.from_records(table.data, columns=table.columns, coerce_float=True)
```
mcmcgrath13: (nit, b): it feels weird to have this function mutate the original arg - can we save off the data and use without mutating

```suggestion
    data = [map(convert_dates, tpl) for tpl in table.data]

# Short cut to valid CSV - can swap out later if performance dictates
# or serialize to CSV at a different location
    df = pd.DataFrame.from_records(data, columns=table.columns, coerce_float=True)
```

apps/report-execution/src/utils.py
mcmcgrath13, line 27:
```suggestion
res = get_str_env_or_default(env_var, default)
```

or not worth it/annoying because of types?

apps/report-execution/src/utils.py
```diff
@@ -35,6 +35,15 @@ def get_int_env_or_default(env_var: str, default: int):
         return default
 
 
+def get_str_env_or_default(env_var: str, default: int):
```
mcmcgrath13:
```suggestion
def get_str_env_or_default(env_var: str, default: str):
```

apps/report-execution/tests/integration/libraries/qa_03.py
```diff
@@ -47,7 +49,7 @@ def test_execute_report_check_data(self, snapshot):
         assert record[0].startswith('CAS')
         assert record[2].startswith('PSN')
         assert int(record[3]) >= 100
-        assert len(record[5]) == 10
+        assert record[5] == datetime.datetime(2022, 1, 24, 0, 0)
```
mcmcgrath13:
```suggestion
        assert isinstance(record[5], datetime.datetime)
```

want to check shape, not literal value here

adaball: Ah yes, good call.  Fixed.

apps/report-execution/src/models.py
adaball: Can do: https://github.com/CDCgov/NEDSS-Modernization/pull/3269/commits/5b87dcb7064817fad69e1476b212a67b6a2a5398

apps/report-execution/src/models.py
adaball: Good call: https://github.com/CDCgov/NEDSS-Modernization/pull/3269/commits/5144592c1270d5ed89d86f26f088bf10579968b5

Interestingly the `pyrefly` type checking required that I explicitly cast the mapped results to a list.

apps/report-execution/src/utils.py
adaball: I think because of types I'd keep it as-is.

mcmcgrath13 (approved):
📆

PR #3270, opened 2026-06-02
[APP-690] feat: start report admin routing

mcmcgrath13:
## Description

Add routing from NBS 6's report listing to the new modernized report configuration admin pages. This required adding a redirect route since we need to go through an `/nbs` page so 7 can do auth before going to a `/report` page. I made this a generic redirect-er instead of following the pattern of a redirect-er per-path

Notes:
* The add page doesn't actually exist yet, so this is an error until that does exist (shortly) - you'll need to go to nbs 6 directly to add a new report in the meantime
* Added e2e's for the view report and routing. Added the testing-library package so we get nicer/more semantic selectors available. Skipped adding a page since I didn't see a point, but happy to be convinced there's a point
* Switched from `ValueView` => `ValueField` on the filter viewing render since that has nicer a11y properties and I didn't want to mess with `ValueView` until I understand why it is the way it is better
* Renamed the `PATIENT_FILE` order to `MODERNIZED_SERVICE` as it was already being used more broadly and is generally a "middle" order that makes sense more broadly

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-690


mcmcgrath13:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=CDCgov_NEDSS-Modernization&pullRequest=3270) **Quality Gate failed**
> Failed conditions  [70.3% Coverage on New Code](https://sonarcloud.io/component_measures?id=CDCgov_NEDSS-Modernization&pullRequest=3270&metric=new_coverage&view=list) (required ≥ 90%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=CDCgov_NEDSS-Modernization&pullRequest=3270)

The code that sonar says isn't covered is definitely covered... 🤷

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportRedirectController.java
```diff
@@ -0,0 +1,24 @@
...
+    prefix = "nbs.ui.features.report.execution",
+    name = "enabled",
+    havingValue = "true")
+class ReportRedirectController {
+
+  /** Intercepts legacy report add requests and re-routes to mod. Needed for auth to work. */
+  @Hidden
+  @GetMapping("/nbs/redirect/report/**")
+  RedirectView reportRedirect(HttpServletRequest request) {
+    var parts = request.getRequestURL().toString().split("/nbs/redirect");
```
emyl3: `String[]` instead of `var` here?

apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportAdminRouteLocatorConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+ *   <li>Path equal to {@code /nbs/ViewReport.do
+ *   <li>Query param for data_source_uid
+ *   <li>Query param for report_uid
+ * </ul>
+ * <ul>
+ *   <li>Request is a GET
+ *   <li>Path equal to {@code /nbs/NewReport.do
+ * </ul>
+ */
+@Configuration
```
emyl3, line 32: General question (nb): I remember when we did this for the `ReportExecutionRouteLocatorConfiguration` we had to copy cookies to make the auth work... would we also need that here? 🤔

emyl3 (approved):
One small change and a general question but approving!

apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportAdminRouteLocatorConfiguration.java
mcmcgrath13: We don't for this one, but in a similar vein we need the redirect route here. We were able to avoid the redirect route on the config api requests because of the cookie copying, but that doesn't apply to this situation

mcmcgrath13: basically when you go from 6 to 7 you need to go via a page that starts with `/nbs` as the `JSESSION` cookie is scoped to the `/nbs` path so if you go straight to `/report` then the cookie isn't available and we can't log you in to 7 (which adds its cookies at the top level `/`)

krista-skylight:
(q, maybe b): are cypress tests failing expectedly? FYI-I tried running one of the new ones you wrote and got this failure which didn't seem to happen in CI, but I know cypress is finicky.

image

mcmcgrath13:
> (q, maybe b): are cypress tests failing expectedly? FYI-I tried running one of the new ones you wrote and got this failure which didn't seem to happen in CI, but I know cypress is finicky.
> 
> image

That _should_ work (and was working for me locally earlier), but looks like there's some merge cruft that happened in the cypress tests I need to clean up

krista-skylight (approved):
:rocket:

PR #3271, opened 2026-06-02
[APP-692] feat: add report sort scaffolding

mcmcgrath13:
## Description

Add the database table for `Report_Sort_Column` and plumb through the values to the config and add fields for them on the request as well

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-691


libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/Report.java
```diff
@@ -40,6 +40,11 @@ public class Report {
   @OneToMany(mappedBy = "report", fetch = FetchType.LAZY)
   private List<DisplayColumn> displayColumns;
 
+  // should be 1:1 in practice, but the DB implies there could be
+  // more and NBS 6 fetches all then takes the first
+  @OneToMany(mappedBy = "report", fetch = FetchType.LAZY)
```
JordanGuinn, line 45: 😭

apps/modernization-ui/src/generated/models/ReportExecutionRequest.ts
```diff
@@ -9,7 +9,15 @@ export type ReportExecutionRequest = {
     dataSourceUid: number;
     isExport: boolean;
     columnUids?: Array<number>;
+    sortColumnUid?: number;
```
JordanGuinn: (nit, nb): is it worth bundling these together into a single optional `sort` param?  Just so sort itself remains optional, but if you specify sort, you specify both.  Unless we're gonna have defaults for each?

JordanGuinn (approved):

apps/modernization-ui/src/generated/models/ReportExecutionRequest.ts
mcmcgrath13: I was also contemplating this and couldn't decide it it was overkill or not - they only make sense as a pair, but at the same time we're only going to add the order by if the column is there and then the logic is "are you ascending? if not, descending" - so we don't reallly require the direction in practice. In NBS6 this is kinda shipped to/from the UI on the display column, which was a path I didn't want to repeat. I think I've talked myself into consolidating it 😂

PR #3274, opened 2026-06-03
fix: trim column names and titles for safety

mcmcgrath13:
## Description

Trim column names and titles to make sure whitespace doesn't cause problems when formatting/using them


adaball (approved):
:shipit:

PR #3275, opened 2026-06-03
fix: display column join to report

mcmcgrath13:
## Description

I noticed when debugging the potential duplicate join things that the columns weren't pre-populating like they should. This fixes this

The tests will come when e2e's start happening for the report workflow (soon)


brick-green (approved):
🤝

emyl3 (approved):

PR #3266, opened 2026-05-29
[APP-670] feat: report admin config viewing

mcmcgrath13:
## Description

Create the page for viewing a report configuration as an admin and start the scaffolding for the rest of the CRUD.

Notes:
* Consolidated the horizontal field and value field since they should be the same-ish style wise
* Moved report layout things to the top level of the folder and made a bit more generic
* Added needed options endpoints to the back end
  * Note the report library endpoint isn't easily unit testable right now as the config api doesn't get the liquibase stuff applied 🥴 ) - this will get tested in a fast follow PR which will do the gateway routing and add some e2e testing
* Added needed fields to the `ReportConfiguration` the backend returns (and renamed `libraryName` to `name` for consistency with other api-models)
* Left some TODO's for the follow on create/edit as this PR was big enough for now and they weren't critical to this work

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-670

## Checklist before requesting a review
- [ ] PR focuses on a single story
- [ ] Code has been fully tested to meet acceptance criteria
- [ ] PR is reasonably small and reviewable (Generally less than 10 files and 500 changed lines)
- [ ] All new functions/classes/components reasonably small
- [ ] Functions/classes/components focused on one responsibility
- [ ] Code easy to understand and modify (clarity over concise/clever)
- [ ] PRs containing TypeScript follow the [Do's and Don'ts](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)
- [ ] PR does not contain hardcoded values (Uses constants)
- [ ] All code is covered by unit or feature tests

PR #3277, opened 2026-06-03
[APP-369] Add program area, jurisdiction, and reporting facility permission to subset query

brick-green:
## Description

This PR adds a permission fragment to the subset sql via the `WhereClauseService` for jurisdiction, program area, and facility.  

## Tickets

* [APP-369](https://cdc-nbs.atlassian.net/jira/software/c/projects/APP/boards/626?selectedIssue=APP-369)

[APP-369]: https://cdc-nbs.atlassian.net/browse/APP-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportConfiguration.java
```diff
@@ -8,6 +8,7 @@ public record ReportConfiguration(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) ReportDataSource dataSource,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Library library,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String title,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Character shared,
```
mcmcgrath13: (nit, b-ish): this is added in https://github.com/CDCgov/NEDSS-Modernization/pull/3266 as well, but called `group` - can we stick to that wording? (the "shared" is very db-ey vs ux-ey). We'll also want to change both of these PRs to use the enum instead of a string/character

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportDataSource.java
```diff
@@ -3,9 +3,15 @@
 import gov.cdc.nbs.entity.odse.DataSource;
 import io.swagger.v3.oas.annotations.media.Schema;
 
-public record ReportDataSource(@Schema(requiredMode = Schema.RequiredMode.REQUIRED) String name) {
+public record ReportDataSource(
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) String name,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Character jurisdictionSecurity,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Character facilitySecurity) {
```
mcmcgrath13: (nit, nb): in other cases where we're consuming an indicator field, we've translated it to a boolean on the model

```suggestion
    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Character hasJurisdictionSecurity,
    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Character hasFacilitySecurity) {
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportDataSource.java
```diff
@@ -3,9 +3,15 @@
...
 
   public ReportDataSource(DataSource entity) {
-    this(entity.getDataSourceName());
+    this(
+        entity.getDataSourceName(),
+        entity.getJurisdictionSecurity(),
+        entity.getReportingFacilitySecurity());
```
mcmcgrath13:
```suggestion
        "Y".equals(entity.getJurisdictionSecurity().toString()),
        "Y".equals(entity.getReportingFacilitySecurity().toString()));
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -15,6 +15,18 @@ public final class ReportConstants {
   public static final String SQL_AND = " AND ";
   public static final String SQL_WHERE = "WHERE ";
 
+  public static final String REPORTINGOPERATION = "REPORTING";
+
+  public static final String VIEWREPORTTEMPLATE = "VIEWREPORTTEMPLATE";
+  public static final String VIEWREPORTPRIVATE = "VIEWREPORTPRIVATE";
+  public static final String VIEWREPORTPUBLIC = "VIEWREPORTPUBLIC";
+  public static final String VIEWREPORTREPORTINGFACILITY = "VIEWREPORTREPORTINGFACILITY";
```
mcmcgrath13: (thought, nb): I wonder if we want to start a more central permissions constants file for these sorts of things vs have it part of the report constants 🤔

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
-    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+    List<String> activeClauses = new ArrayList<>();
 
     String basicWhereFragment =
         buildBasicWhereFragment(reportConfig, executionRequest.basicFilters());
+    if (!basicWhereFragment.isBlank()) {
+      activeClauses.add(basicWhereFragment);
+    }
+
+    String permissionFragment = buildPermissionFragment(reportConfig, reportConfig.shared());
```
```suggestion
    String permissionFragment = buildPermissionFragment(reportConfig);
```

(sugg, nb): seems redundant to pass a field of the thing we're already passing the whole of

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
+    }
+
+    if (permissionClauses.isEmpty()) {
+      return "";
+    }
+
+    return "(" + String.join(SQL_AND, permissionClauses) + ")";
+  }
+
+  private String getJurisProgramCriteria(Character jurisdictionSecurity, Character sharedStatus) {
```
```suggestion
  private String getJurisProgramCriteria(boolean hasJurisdictionSecurity, ReportGroup group) {
```
(nit, b-probably): can we use the enum instead?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
+
+    Permission permission = mapSharedToPermission(sharedStatus);
+    if (permission == null) {
+      return "";
+    }
+
+    PermissionScope scope = this.scopeResolver.resolve(permission);
+    if (scope.any().isEmpty()) {
+      return "";
+    }
```
mcmcgrath13, line 180: (q, b): if this is empty, does that mean they should have no permissions or all permissions? and can we add a comment about this either way

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
+
+    String ids = scope.any().stream().map(String::valueOf).collect(Collectors.joining(", "));
 
-    finalWhere.add(basicWhereFragment);
+    return "(program_jurisdiction_oid IN (" + ids + "))";
+  }
+
+  private String getReportingFacilityCriteria(Character reportingFacilitySecurity) {
```
```suggestion
  private String getReportingFacilityCriteria(boolean hasReportingFacilitySecurity) {
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
+    if (reportingFacilitySecurity == null || reportingFacilitySecurity != 'Y') {
+      return "";
+    }
+
+    Long externalOrgId = SecurityUtil.getUserDetails().getExternalOrgUid();
+    if (externalOrgId == null) {
+      return "";
```
mcmcgrath13: (q, b): similarly, if a user doesn't have a facility, but the data source is secured by facility - should they have no data or all data?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
+    }
+
+    return "(REPORTING_FACILITY_UID = " + externalOrgId + ")";
+  }
 
-    // Only return the WHERE clause if it contains anything beyond the initial "WHERE " prefix
-    return finalWhere.length() > SQL_WHERE.length() ? finalWhere.toString() : "";
+  /** Pure mapper utility isolating the business rules matching flags to permissions. */
+  private Permission mapSharedToPermission(Character shared) {
```
```suggestion
  private Permission mapSharedToPermission(ReportGroup group) {
```

(nit, b): switch to enum

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -37,16 +44,95 @@ public class WhereClauseService {
...
+      case ReportConstants.TEMPLATE_REPORT_GROUP_CHAR ->
+          new Permission(ReportConstants.REPORTINGOPERATION, ReportConstants.VIEWREPORTTEMPLATE);
+      case ReportConstants.PRIVATE_REPORT_GROUP_CHAR ->
+          new Permission(ReportConstants.REPORTINGOPERATION, ReportConstants.VIEWREPORTPRIVATE);
+      case ReportConstants.PUBLIC_REPORT_GROUP_CHAR ->
+          new Permission(ReportConstants.REPORTINGOPERATION, ReportConstants.VIEWREPORTPUBLIC);
+      case ReportConstants.REPORTING_FACILITY_REPORT_GROUP_CHAR ->
+          new Permission(
+              ReportConstants.REPORTINGOPERATION, ReportConstants.VIEWREPORTREPORTINGFACILITY);
+      default -> null;
```
mcmcgrath13: (sugg, b): should this be an error state? with my 😈 hat on - if I put in a weird group in the DB then I unlock all rows instead of securing it all the way - feels like a loophole

mcmcgrath13: maybe forcing this to be enum gets us that as well? since the failing to parse into an enum would be an earlier failure of the whole operation?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: Tested this a bit in 6 and it looks like the Reports tab is not accessible for users without a PA and Jurisdiction set. With that knowledge I think this should be an error state instead of allowing all.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: When I tested this in 6 I found that a facility is not required to access reports. This was different from Juris/PA. So I think leaving this as returning an empty string is correct.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -78,30 +84,110 @@ public String buildWhereClause(
...
+    }
 
     String advWhereFragment =
         buildAdvancedQueryResult(reportConfig, executionRequest.advancedFilter());
-
-    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
-
-    if (basicWhereFragment != null && !basicWhereFragment.isEmpty()) {
-      finalWhere.add(basicWhereFragment);
+    if (advWhereFragment != null && !advWhereFragment.isEmpty()) {
```
JordanGuinn: (nit, nb): Is there any reason we're checking `isBlank()` on the `basicWhereFragment` while checking `isEmpty()` on the `advWhereFragment`?  Is there a distinction?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
JordanGuinn: +1 to a comment indicating what `""` means for both this and `getReportingFacilityCriteria`, or potentially tweaking the method names to try and indicate as much!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
JordanGuinn: (q, nb): I know that we're working against a relatively tight deadline here, so maybe something worth holding off on (at least for the time being), but I wonder if we might also benefit from some e2e tests for these different permutations?  Just to more holistically ensure that permissions are honored by the application

JordanGuinn (approved):
🙏

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: Since this is using the enum now I was able to delete the default state.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: `isEmpty()`: "true if length() is 0, otherwise false"

`isBlank()`: "true if the string is empty or contains only white space codepoints, otherwise false"

those are the descriptions from the methods in `String.java`. I typically default to `isBlank()`. I'll update the check on advWhereFragment

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -16,6 +16,18 @@ public final class ReportConstants {
...
+
+  public static final char PRIVATE_REPORT_GROUP_CHAR = 'P';
+  public static final char REPORTING_FACILITY_REPORT_GROUP_CHAR = 'R';
+  public static final char PUBLIC_REPORT_GROUP_CHAR = 'S';
+  public static final char TEMPLATE_REPORT_GROUP_CHAR = 'T';
```
mcmcgrath13: merge cruft?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -78,30 +84,132 @@ public String buildWhereClause(
...
+    if (!permissionFragment.isBlank()) {
+      activeClauses.add(permissionFragment);
     }
 
     boolean hasLabResultVal = hasLabResultVal(reportConfig, executionRequest.advancedFilter());
     if (hasLabResultVal) {
       String rdbDataSource = dataSourceNameUtils.buildDataSourceName("nbs_rdb.lab_test_report");
-      return LAB_RESULT_QUERY_VAL.formatted(rdbDataSource, finalWhere.toString());
+      return LAB_RESULT_QUERY_VAL.formatted(
+          rdbDataSource, SQL_WHERE + String.join(SQL_AND, activeClauses));
```
mcmcgrath13: (nit, b): from my reading of NBS 6 (below), the security and reporting facility where always go on the "outer" where, so in this lab case, we need to do the lab shenanigans without the permission fragments, and then add the permission fragment at the end

Image

mcmcgrath13 (changes requested):
overall looks good, but that lab where ordering could be semantically different putting the security filters on the inner where vs the outer, so think it's important to match 6 there

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
brick-green: nice catch! Updated the code and the `all filters` test to confirm. Here is an easier to read version of the assert on that test:

```
WHERE 
  root_ordered_test_pntr IN
    (SELECT root_ordered_test_pntr
     FROM [RDB].[dbo].[lab_test_report]
     WHERE ([COLUMN_INTEGER] IN (1))
       AND (([TimeRangeColumn] BETWEEN '2023-01-01' AND '2024-01-01')
            OR ([TimeRangeColumn] IS NULL))
       AND ((CAST([COLUMN_DATETIME] AS DATE) IN ('2026-05-28'))
            OR (([COLUMN_STRING] LIKE CONCAT('%', 'foo', '%'))
                AND (([COLUMN_INTEGER] NOT IN (1)
                      OR [COLUMN_INTEGER] IS NULL)
                     OR (([COLUMN_STRING] LIKE CONCAT('foo', '%'))
                         AND ([COLUMN_STRING] IN ('2019 Novel Coronavirus',
                                                  'AIDS',
                                                  'Acanthamoeba Disease (Excluding Keratitis)'))
                         AND ([COLUMN_INTEGER] IN (1))
                         AND ([COLUMN_INTEGER] >= 1))
                     OR (CAST([COLUMN_DATETIME] AS DATE) > '2026-05-28')
                     OR ([numeric_result_val] = 1))
                AND ([COLUMN_STRING] NOT IN ('2019 Novel Coronavirus',
                                             'AIDS',
                                             'Acanthamoeba Disease (Excluding Keratitis)')
                     OR [COLUMN_STRING] IS NULL)
                AND ([RESULT_UNITS] <> '1'
                     OR ([RESULT_UNITS] IS NULL)))
            OR (CAST([COLUMN_DATETIME] AS DATE) BETWEEN '2026-05-25' AND '2026-05-28')
            OR (CAST([COLUMN_DATETIME] AS DATE) IS NOT NULL)
            OR ([COLUMN_INTEGER] > 1)
            OR ([COLUMN_INTEGER] BETWEEN 1 AND 2))
    )
  AND ((program_jurisdiction_oid IN (50))
       AND (REPORTING_FACILITY_UID = 54321))
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -18,6 +18,18 @@ public final class ReportConstants {
   public static final String SQL_AND = " AND ";
   public static final String SQL_WHERE = "WHERE ";
 
+  public static final String REPORTINGOPERATION = "REPORTING";
+
+  public static final String VIEWREPORTTEMPLATE = "VIEWREPORTTEMPLATE";
+  public static final String VIEWREPORTPRIVATE = "VIEWREPORTPRIVATE";
+  public static final String VIEWREPORTPUBLIC = "VIEWREPORTPUBLIC";
+  public static final String VIEWREPORTREPORTINGFACILITY = "VIEWREPORTREPORTINGFACILITY";
```
mcmcgrath13: (nit, nb): create a public final inner class `Permissions` to put these in - I think there will be more added here over time and that gives them a lil more context

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportConstants.java
```diff
@@ -18,6 +18,18 @@ public final class ReportConstants {
...
+
+  public static final char PRIVATE_REPORT_GROUP_CHAR = 'P';
+  public static final char REPORTING_FACILITY_REPORT_GROUP_CHAR = 'R';
+  public static final char PUBLIC_REPORT_GROUP_CHAR = 'S';
+  public static final char TEMPLATE_REPORT_GROUP_CHAR = 'T';
```
mcmcgrath13:
```suggestion
  public final class GroupChar {
  public static final char PRIVATE_REPORT = 'P';
  public static final char REPORTING_FACILITY_REPORT = 'R';
  public static final char PUBLIC_REPORT = 'S';
  public static final char TEMPLATE_REPORT = 'T';
  }
```

very non-blocking

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -79,30 +85,142 @@ public String buildWhereClause(
...
+      String rdbDataSource = dataSourceNameUtils.buildDataSourceName("nbs_rdb.lab_test_report");
+      String labResultQueryValFragment =
+          LAB_RESULT_QUERY_VAL.formatted(
+              rdbDataSource, SQL_WHERE + String.join(SQL_AND, activeClauses));
+
+      String permissionFragment = buildPermissionFragment(reportConfig);
+      if (!permissionFragment.isBlank()) {
+        return labResultQueryValFragment
+            + SQL_AND
+            + String.join(SQL_AND, buildPermissionFragment(reportConfig));
```
```suggestion
        return labResultQueryValFragment
            + SQL_AND
            + buildPermissionFragment(reportConfig);
```

The permission fragment looks like it's already anded

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -79,30 +85,142 @@ public String buildWhereClause(
...
 
-    StringJoiner finalWhere = new StringJoiner(SQL_AND, SQL_WHERE, "");
+    boolean hasLabResultVal = hasLabResultVal(reportConfig, executionRequest.advancedFilter());
+    if (hasLabResultVal) {
+      String rdbDataSource = dataSourceNameUtils.buildDataSourceName("nbs_rdb.lab_test_report");
+      String labResultQueryValFragment =
+          LAB_RESULT_QUERY_VAL.formatted(
+              rdbDataSource, SQL_WHERE + String.join(SQL_AND, activeClauses));
+
+      String permissionFragment = buildPermissionFragment(reportConfig);
```
mcmcgrath13, line 110:
```suggestion
      // Add permission fragment to outer where
      String permissionFragment = buildPermissionFragment(reportConfig);
```

mcmcgrath13 (approved):
some minor suggestions, but overall looks good!

PR #3279, opened 2026-06-03
(WIP) APP-505: Advanced Query Builder/Parser

JordanGuinn:
## Description

Please include a summary of the changes and any key information a reviewer may need.

## Tickets

* [Jira Ticket](url)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -1,12 +1,16 @@
...
 import gov.cdc.nbs.report.ReportConstants;
 import gov.cdc.nbs.report.models.AdvancedFilterConfiguration;
-import gov.cdc.nbs.report.models.AdvancedQuery;
 import gov.cdc.nbs.report.models.FilterType;
+import java.util.Arrays;
+import java.util.List;
 
 public class AdvancedFilterConfigurationMapper {
+  private static final List<String> KNOWN_OPERATORS =
+      Arrays.asList("GE", "LE", "GT", "LT", "BW", "CO", "SW", "EQ", "NE", "IN", "NN");
```
mcmcgrath13, line 15: heads up that in #3267 @emyl3 has created an enum for this - worth stealing/consolidating on that approach?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -18,10 +22,7 @@ public static AdvancedFilterConfiguration fromReportFilter(ReportFilter filter)
           "Cannot create advanced filter from non where clause builder filter");
     }
 
-    // For the future: Populate this
-    // https://cdc-nbs.atlassian.net/browse/APP-505
-    AdvancedQuery.RuleGroup defaultValue = null;
-
-    return new AdvancedFilterConfiguration(filter.getId(), defaultValue);
+    return new AdvancedFilterConfiguration(
+        filter.getId(), new AdvancedQueryBuilder(filter.getFilterValues()).build());
```
mcmcgrath13: (thought) - when the getting filter values doesn't parse, we want to send an error/warning on the configuration and include some representation of what was in the database. This will probably need to get slightly more complicated as a structure to support that

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+    ReportConstants.QueryCombinators firstCombinator = null;
+
+    // Iterate through FilterValues until we find the first OPERATOR
+    while (hasNext()) {
+      FilterValue filterValue = peek();
+      if (filterValue.getValueType().equals("CLAUSE")) {
+        firstRule = buildClause(filterValue);
+      } else if (filterValue.getValueType().equals("OPERATOR")) {
+        if (filterValue.getOperator().equals("and") || filterValue.getOperator().equals("or")) {
+          firstCombinator = ReportConstants.QueryCombinators.valueOf(filterValue.getOperator());
```
mcmcgrath13: worth (/possible) parsing the database value to/from the enum instead of strings?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+              }
+              break;
+            case "(":
+              if (previous() != null && previous().getOperator().equals(")")) {
+                queryErrors.add(
+                    new AdvancedQueryException(
+                        "Cannot follow a closing parenthesis with an open parenthesis without an OPERATOR in between",
+                        filterValue));
+              }
+              nestDepth++;
```
mcmcgrath13: opening a parens is essentially creating a new rule group - does the nest depth increment do this? I'm not 100% following the usage of nest depth

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+      }
+    }
+
+    if (nestDepth != 0) {
+      queryErrors.add(new AdvancedQueryException("Mismatched parentheses"));
+    }
+
+    if (!queryErrors.isEmpty()) {
+      //  TODO: What exactly do we wanna do here?  Should we still return the partially built
+      // ruleGroup
```
mcmcgrath13: we want to return no rule group and a warning message that displays what we tried to interpret but couldn't

mcmcgrath13: and definitely let everything else return, so a 2xx

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+
+                // And attach it to the existing list of 'OR' group rules
+                rules.add(andGroup);
+              }
+              break;
+            case "(":
+              if (previous() != null && previous().getOperator().equals(")")) {
+                queryErrors.add(
+                    new AdvancedQueryException(
+                        "Cannot follow a closing parenthesis with an open parenthesis without an OPERATOR in between",
```
mcmcgrath13: do we also need to check for operator->operator sequences and clause->clause sequences?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+  public AdvancedQueryBuilder(List<FilterValue> filterValues) {
+    this.filterValues = filterValues;
+  }
+
+  public AdvancedQuery.RuleGroup build() {
+    AdvancedQuery.Rule firstRule = null;
+    ReportConstants.QueryCombinators firstCombinator = null;
+
+    // Iterate through FilterValues until we find the first OPERATOR
+    while (hasNext()) {
```
mcmcgrath13: I'm wondering if a cleaner starting logic is checking for an opening paren and if it exists eating it and starting a rule group or otherwise just starting a rule group?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: > opening a parens is essentially creating a new rule group

That was my initial strategy, but that's not necessarily how it works in 6.  Query portions with needless multiple parens (i.e. `((( DIAGNOSIS_DATE IS NOT NULL )))`) are considered valid in 6, so using `nestDepth` is a means to try to maintain compatibility with that sorta thing in 7 as well, without adding complexity to the AdvancedQuery with additional rule groups that don't actually need to be there.

JordanGuinn: I'd be VERY down to not support those kinds of parens situations, but I wasn't sure if that was on the table?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: I think it depends on how we wanna handle https://github.com/CDCgov/NEDSS-Modernization/pull/3279#discussion_r3357632943

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: Will look into this.  It's kinda tough in this situation because the `operator` field is only `and`/`or` if the `value_type` is `OPERATOR`, so we can't parse into this enum under all circumstances

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
mcmcgrath13: I was imagining this would translate to a a `{id: blah, rules: [{id: blah, rules: [{id: blah, rules: [DIAGNOSIS_DATE IS NOT NULL]}]}]}` which we would then flatten down after

PR #3288, opened 2026-06-04
Update sample.env with more descriptive values

adaball:
## Description

Update the `sample.env` file to be more descriptive with regards to the report export formatting.  Also puts quotes around the `REPORT_EXPORT_*` values in said file.


mcmcgrath13 (approved):
🧹

PR #3289, opened 2026-06-04
[APP-646] feat: add report configuration

mcmcgrath13:
## Description

Add report configuration creation to the sys admin UI

image

Notes:
* When navigating from 7 -> 6 need to use regular links and JS as the react router stuff will not work (assumes truly internal nav when relative routes used)
* Added CI running on merge to main
* The redirect controller needs to work with post as well as get as for some reason hitting the "Create" report button sends a post request to that URL 🥴 
* Added "back to report management" link on the view page as I was looking for it as user and when doing e2e testing

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-646
## Checklist before requesting a review
- [ ] PR focuses on a single story
- [ ] Code has been fully tested to meet acceptance criteria
- [ ] PR is reasonably small and reviewable (Generally less than 10 files and 500 changed lines)
- [ ] All new functions/classes/components reasonably small
- [ ] Functions/classes/components focused on one responsibility
- [ ] Code easy to understand and modify (clarity over concise/clever)
- [ ] PRs containing TypeScript follow the [Do's and Don'ts](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)
- [ ] PR does not contain hardcoded values (Uses constants)
- [ ] All code is covered by unit or feature tests

PR #3296, opened 2026-06-08
fix: repeating block state handling

mcmcgrath13:
## Description

Spinoff from #3289 - the repeating block state handling was calling `onChange` inside the reducer, which causes errors in React. This pre-exists this work and it kinda was working, but also could be why when we tried to upgrade the backing form library we got a bunch of infinite loop type issues

Changes:
* If `onChange` is passed, it is expected to update the data and pass it back in (in practice, this is already happening in all use sites)
* add `itemName` prop and default to `title`, so that the card title doesn't have to match the verbage for the action buttons


emyl3 (approved):
Thank you for cleaning this up!

PR #3297, opened 2026-06-08
fix: design system cleanup and enhancements

mcmcgrath13:
## Description

Spinoff from #3289 

* Add a `ToggleField` and have it support non-visible labels on `Toggle` to avoid double labeling (spun off the labeled utils to support this)
* Add a `disabled` prop to the `Card`
* Get rid of `useCallback` on `Select` which is overengineered and causing errors when options is not defined


apps/modernization-ui/src/apps/deduplication/configuration/import/ImportConfigurationModal.spec.tsx
```diff
@@ -4,9 +4,6 @@ import { BlockingAttribute, MatchingAttribute, MatchMethod } from 'apps/deduplic
 import { ImportConfigurationModal } from './ImportConfigurationModal';
 import { AlgorithmExport } from 'apps/deduplication/api/model/AlgorithmExport';
 
-const onImport = vi.fn();
```
mcmcgrath13, line 7: these tests mysteriously started failing, so tightened up their mocking usage to fix them

brick-green (approved):
👍

PR #3298, opened 2026-06-08
feat: add more options endpoints

mcmcgrath13:
## Description

Spinoff from #3289 - Add new options endpoints for report filters and filterable data source columns

Notes:
* I flip flop on whether the filterable should be a path param or move it to a query param with the thought we might want `displayable` or some other type of view of this data down the road (not in current roadmap) - thoughts?


libs/options-api/src/main/java/gov/cdc/nbs/option/report/datasources/columns/ReportDataSourceColumnController.java
```diff
@@ -0,0 +1,27 @@
...
+  ReportDataSourceColumnController(final ReportDataSourceFilterableColumnFinder finder) {
+    this.finder = finder;
+  }
+
+  @Operation(
+      operationId = "reportDataSourceColumns",
+      summary = "Filterable report columns for a data source",
+      description = "Provides all filterable report columns in a data source in NBS.",
+      tags = "ReportDataSourceFilterableColumns")
+  @GetMapping("/nbs/api/options/report/datasource/columns/filterable/{dataSourceUid}")
```
brick-green: thought, nb: to your question in the description, I think the textbook would recommend going with something like this:


```suggestion
  @GetMapping("/nbs/api/options/report/datasource/{dataSourceUid}/columns?status=filterable")
```


I'm feeling flipfloppy about it too though. I think there's value is being specific when we don't have product requirements for being flexible.

brick-green (approved):
👍

libs/options-api/src/main/java/gov/cdc/nbs/option/report/datasources/columns/ReportDataSourceColumnController.java
mcmcgrath13: seems reasonable to me - I thinking `filterable=true` is probably going to play better with how the database thinks about the dat

mcmcgrath13: the other likely candidate for query param is `displayable` (and no filter at all), but we just don't _need_ it at this point

mcmcgrath13: I started going down the adding it as a param hole and quickly realized the complexity is non-trivial as it's really a three value'd logic (filter on 'Y', filter on 'N', no filter at all), so fixed up the path ordering and called it a day

PR #3300, opened 2026-06-08
remove wrongly committed debug code and file

adaball:
## Description

It would appear I accidentally left some debug lines in a test which caused `sr18snapshot.yaml` to be committed to `main`.

This PR removes that file and the test debug line.


emyl3 (approved):
🚢

PR #3301, opened 2026-06-08
APP-543 Convert QA07

krista-skylight:


apps/report-execution/tests/integration/assets/tablefaker_schema/std_hiv_datamart.yaml
```diff
@@ -47,98 +52,286 @@ tables:
...
         data: foreign_key("[RDB].[dbo].[D_ORGANIZATION]", "organization_key")
+
       - column_name: patient_pregnant_ind
-        data: random.choices(["No", "Yes", None], weights=[90, 5, 5])[0]
+        data: |
+          r = fake.random_int(1, 100)
+          if r <= 90: return "No"
+          elif r <= 95: return "Yes"
+          else: return None
+
```
timoballard, line 261: (q, nb): is this an idiomatic/preferential change or does the other format not actually work the way we want?

krista-skylight: When I was running tests, some old snapshots were seemingly randomly not matching, so I had a suspicion that the tablefaker seed was not working for `random.choice` which is why I switched everything to `fake.random` and that seemed to fix the problem. The rest of it is just a preferential change.

timoballard: oh weird, I would have guessed that `random.choice` is just syntactic sugar for the above

krista-skylight: you know what, I think I am wrong because [tablefaker docs ](https://pypi.org/project/tablefaker/) say it applies to both random and numpy. anyhow, somehow i got snapshots to match :sweat_smile:

timoballard (approved):

adaball (approved):
:shipit:

PR #3303, opened 2026-06-09
feat: update api to use enums, nbs class code

mcmcgrath13:
## Description

Spinoff from #3289 - update the mod API to use the enums for group and select type on the wire. Also some minor cleanup to the creating things code

Notes
* Drop `min` and `max` counts fields in lieu of `selectType`
* Use the ID generator service for `ReportFilter` saving, use the `NBS` type instead of `REPORT`/`REPORTFILTER` to match NBS 6 behavior
* If the saved library is a SAS library, populate the location and type so NBS 6 can still run them


apps/modernization-ui/src/apps/report/management/configuration/ReportConfigurationContent.tsx
```diff
@@ -19,11 +19,9 @@ interface FilterConfig {
 const sizing = 'medium';
 
 // TODO: Move this to the API once #3238 is merged
-const formatType = ({ minValueCount, maxValueCount }: BasicFilterConfiguration) => {
-    if (minValueCount === 1) {
-        if (maxValueCount === -1) return 'Multi';
-        if (maxValueCount === 1) return 'Single';
-    }
+const formatType = ({ selectType }: BasicFilterConfiguration) => {
```
mcmcgrath13, line 22: left the TODO since this gets fully dealt with in the create UI PR - this is just a small-as-possible change to get this PR to not be broken

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/ValueCountCalculator.java
```diff
@@ -63,4 +63,13 @@ public static ReportValueCounts fromFilterRequest(UpsertFilterRequest filterRequ
 
     return new ReportValueCounts(minValueCount, maxValueCount);
   }
+
+  public static ReportConstants.SelectType toSelectType(ReportValueCounts counts) {
+    if ((Integer.valueOf(1)).equals(counts.minValueCount)) {
```
emyl3, line 68: To clarify... we are just trying to account for select type here and not other combinations of min/max values right? 👀

mcmcgrath13: correct - the others aren't functionally used as far as I can tell

emyl3 (approved):
🚢

PR #3304, opened 2026-06-09
[APP-407] SR05 Functional Testing

adaball:
## Description

This PR is to confirm that functional testing is working and data parity is achieved for NBS SR05.

Confirming data parity between SAS and Python versions:
Screenshot 2026-06-09 at 09 34 17

Changes:
- Added conversion note
- Updated percentage `SELECT` statement to more closely match SAS output

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-407)

**NOTE:** Not using the new library checklist as this PR is just for fixing an existing library


apps/report-execution/tests/integration/libraries/nbs_sr_05.py
```diff
@@ -131,7 +131,7 @@ def test_execute_report_no_current_year(self):
 
             assert record is not None
             # current year data should be zero, older not
-            assert int(record[0][:-1]) < 0
+            self._verify_percentage(record[0])
```
mcmcgrath13: it looks like verify percentage is checking this is a percent, but we've lost the < 0/is negative part of the test - could you add that back?

mcmcgrath13 (approved):
one minor testing nit, but looks good!

apps/report-execution/tests/integration/libraries/nbs_sr_05.py
adaball: So the query's output is following the SAS script's `percent9.0` formatting for the calculated percentage.  As such the result is not actually going to be displayed as negative but rather like `(15)%` (instead of `-15%`).

I went ahead and improved the `_verify_percentage` function to perform more sanity checks on the percentage.

https://github.com/CDCgov/NEDSS-Modernization/pull/3304/commits/8bb144e6483a82d4508ee78b382d5c6dfe7ab470

PR #3311, opened 2026-06-11
[APP-630] Convert PA03

timoballard:
- [x] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelogreport/execution/03_ODSE_Data_Report_Library_Init.sql`

## Tickets

* [Jira Ticket](url)


apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,257 @@
...
+    'A2 - Associate 2',
+    'A3 - Associate 3',
+}
+VALID_PROCESSING_DECISIONS = {
+    'Field Follow-up',
+    'Secondary Referral',
+    'Record Search Closure',
+}
+OUTCOME_CODES = ('I1', 'I2', 'I3', 'I4', 'I5', 'I6', 'I7')
+VALID_REFERRAL_BASES = tuple(sorted(PARTNER_BASES | SOCIAL_BASES | ASSOCIATE_BASES))
```
mcmcgrath13, line 27: @adaball are these also useful for pa01? should we pull them out to a common constants-y file?

apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    valid_referral_bases_sql = _sql_string_list(VALID_REFERRAL_BASES)
+
+    cases_query = f"""
+    WITH shd AS (
+        {subset_query}
+    )
+    SELECT DISTINCT
+        shd.INV_LOCAL_ID,
+        shd.INIT_FUP_INTERNET_FOLL_UP_CD
```
mcmcgrath13, line 138: (q, nb): this query looks substantially simpler than pa03_new - why are we able to do so much less? any risk of missing something in this simplification?

apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,257 @@
...
+        for outcome_code in OUTCOME_CODES:
+            rows.append(
+                (
+                    f'{label} {outcome_code}',
+                    counts.get(outcome_code, 0),
+                )
+            )
+
+    content = Table(
+        columns=['LABEL', 'VALUE'],
```
mcmcgrath13: (sugg, b): this table structure looks much simpler/squished than the design - can we match it? This enables better filtering of sub-sections of the report and clearer labels

https://docs.google.com/spreadsheets/d/1YpXfOimCcqe-vjrnyz6IB8b6t1W6G8Ajok_X8gac4Cg/edit?gid=728286166#gid=728286166

timoballard: oh my, I completely spaced on the design. Should be pretty straightforward to reshape things, will do

apps/report-execution/src/libraries/pa_03.py
timoballard: If we were replicating the dataset export that SAS does, this query would need to more closely match `pa3_new`. But since we're only doing the final metrics report, we can trim down to the fields that are relevant to those calculations and omit the others (e.g. `provder_last_name`)

apps/report-execution/src/libraries/pa_03.py
timoballard: reshaped and updated the example file in the SAS Library catalogue:

image

apps/report-execution/src/libraries/pa_03.py
mcmcgrath13: could you leave a comment on that in case we ever need to go back and also do the export and are wondering why this isn't like pa03_new?

apps/report-execution/src/libraries/pa_03.py
mcmcgrath13: nice! could we format the index to match the numeric formatting from the PDF?

mcmcgrath13: given our general CSV float formatting, I'm a bit surprised we're seeing that many trailing 0's to begin with 🤔

apps/report-execution/src/libraries/pa_03.py
timoballard: added https://github.com/CDCgov/NEDSS-Modernization/pull/3311/commits/0986a5510370932e881ea9a19366d28767e6f8eb

apps/report-execution/src/libraries/pa_03.py
timoballard: fixed in https://github.com/CDCgov/NEDSS-Modernization/pull/3311/commits/e442b8fa65c7f823a1e6815fab6b07096cbd6962

image

apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,316 @@
...
+        ):
+            outcome_ids[row[5]].add(row[0])
+
+    return {code: len(ids) for code, ids in outcome_ids.items()}
+
+
+def _ratio(numerator: int, denominator: int) -> str | None:
+    if denominator == 0:
+        return None
+    return f'{numerator / denominator:.2f}'
```
mcmcgrath13: do you know why we needed to do this and the default CSV float formatting didn't work?

timoballard: I believe [this](https://github.com/CDCgov/NEDSS-Modernization/blob/b2306652b8266a81bea81051635654b1c66a3392/apps/report-execution/src/models.py#L86) yields up to 20 significant digits, so we were seeing fp precision runoff. Do we want that to be `{:.2g}` ?

timoballard: went ahead and did that in https://github.com/CDCgov/NEDSS-Modernization/pull/3311/commits/b872fbbfd4fae933cca5339afcb87ca7f6e41e63 and rolled back the str conversion on the library side

apps/report-execution/src/models.py
```diff
@@ -83,7 +83,7 @@ def convert_dates(val: Any) -> Any:
 
     csv_str = df.to_csv(
         index=False,
-        float_format='{:.20g}',
+        float_format='{:.2f}',
```
mcmcgrath13: I think we needed the `g` so that numbers would print at as small a precision as they could get away with

timoballard: synced offline: https://github.com/CDCgov/NEDSS-Modernization/pull/3311/commits/3fa55b12c01502c86504b6e078dcb2da46108c07

mcmcgrath13:
could you update the comparison testing files on the SAS catalog?

apps/report-execution/src/libraries/pa_03.py
adaball: To my knowledge these don't map to things in PA01, but it's still early days on it for me so I will check back later.

apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,316 @@
...
+    'A2 - Associate 2',
+    'A3 - Associate 3',
+}
+VALID_PROCESSING_DECISIONS = {
+    'Field Follow-up',
+    'Secondary Referral',
+    'Record Search Closure',
+}
+OUTCOME_CODES = ('I1', 'I2', 'I3', 'I4', 'I5', 'I6', 'I7')
+VALID_REFERRAL_BASES = tuple(sorted(PARTNER_BASES | SOCIAL_BASES | ASSOCIATE_BASES))
```
adaball, line 27: TIL you can use `|` to combine multiple sets into one.

adaball (approved):
Overall LGTM :shipit: 

2 things worth mentioning/asking:

1. I see a slight data descrepency between the PDF and CSV after bringing the whole stack up, loading your tablefaker stuff, and running both the SAS and Python versions (see screenshot)

Image

2. Is there a reason a chose not to do snapshot testing for PA03?  The testing you have is quite thorough so I'm not overly concerned, it's just a break with the normal pattern.

timoballard:
> Overall LGTM :shipit:
> 
> 2 things worth mentioning/asking:
> 
>     1. I see a slight data descrepency between the PDF and CSV after bringing the whole stack up, loading your tablefaker stuff, and running both the SAS and Python versions (see screenshot)
> 
> 
> Image
> 
>     2. Is there a reason a chose not to do snapshot testing for PA03?  The testing you have is quite thorough so I'm not overly concerned, it's just a break with the normal pattern.

1. This turned out to be a [bug in PA03.SAS](https://github.com/cdcent/NEDSSDev/blob/ef89cd6d83915ebe74e2e9900d419dd2175ed0b8/source/sas/pgm/PA03.sas#L411) (note the extra colon) -- 4 is the correct number here. I'm guessing we're not fixing that SAS bug but just documenting the difference somewhere, but will confirm. Thanks!

2. 😱 I was working on the tests without the snapshot because I find them finicky, but then neglected to actually add it in once things were passing. Added in https://github.com/CDCgov/NEDSS-Modernization/pull/3311/commits/6d550b9fb1c4f9543dfb3a4d8f4d1c7cebbfb31f, good catch!

timoballard:
> could you update the comparison testing files on the SAS catalog?

done

adaball (approved):
:shipit:

mcmcgrath13:
It seems like the "Tot No IPS <blah>" are generally the sum of the other rows, but associate has 3 but the total is 4 - why is this?

apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,320 @@
...
+    """PA03 STD Program Report: Internet Partner Services Report.
+
+    Conversion notes:
+    * Keeps the SQL intentionally narrow and performs the SAS metric math in Python.
+    * Returns a long-form table instead of reproducing the SAS PDF layout.
+    * Simplifies SQL queries to only include fields that are relevant to
+    *   calculating the metrics in the summary output. This gives parity
+    *   with the SAS "Run" functionality. If we wish to reach parity with
+    *   SAS "Export" functionality, the SQL queries will need to more closely
+    *   match those in PA03.sas
```
mcmcgrath13:
```suggestion
        calculating the metrics in the summary output. This gives parity
        with the SAS "Run" functionality. If we wish to reach parity with
        SAS "Export" functionality, the SQL queries will need to more closely
        match those in PA03.sas
```

apps/report-execution/src/libraries/pa_03.py
```diff
@@ -0,0 +1,320 @@
...
+    * PA03.SAS contains a bug on line 411. The search string contains an extra ":",
+    *   which causes the value for Total No. IPS Associates in the output PDF
+    *   to always be 0, so the python library will differ from the SAS library
+    *   for that particular field.
```
mcmcgrath13:
```suggestion
        which causes the value for Total No. IPS Associates in the output PDF
        to always be 0, so the python library will differ from the SAS library
        for that particular field.
```

timoballard:
> It seems like the "Tot No IPS " are generally the sum of the other rows, but associate has 3 but the total is 4 - why is this?

Not sure I understand the question - could you rephrase?

mcmcgrath13:
> > It seems like the "Tot No IPS " are generally the sum of the other rows, but associate has 3 but the total is 4 - why is this?
> 
> Not sure I understand the question - could you rephrase?

Why is the sum of the "outcomes" "associate" rows not equal to the value in "Tot No. IPS Associates" row?

And if the SAS were fixed to not have the bug, is that what it would do?

timoballard:
> > > It seems like the "Tot No IPS " are generally the sum of the other rows, but associate has 3 but the total is 4 - why is this?
> > 
> > 
> > Not sure I understand the question - could you rephrase?
> 
> Why is the sum of the "outcomes" "associate" rows not equal to the value in "Tot No. IPS Associates" row?
> 
> And if the SAS were fixed to not have the bug, is that what it would do?

ahh, yeah good question. It comes down to the tablefaker schema [allowing null outcome codes](https://github.com/CDCgov/NEDSS-Modernization/blob/5735cd355520bccd89091994d211c46c2b9debae/apps/report-execution/tests/integration/assets/tablefaker_schema/pa03.yaml#L122). If the outcome code is null, it will count toward the total, but it won't show up in the outcome buckets. That same behavior is true on the SAS side-- when I locally fixed the SAS bug, it showed the same IPS totals of 3/1/4.

I just changed the tablefaker schema to not allow nulls and then the totals line up with the sum of the outcome buckets

mcmcgrath13 (approved):
🕔

PR #3312, opened 2026-06-11
APP-505: Advanced Query Builder/Parser

JordanGuinn:
## Description

This PR introduces a descent parser of sorts for translating `FilterValue` sets into a single nested `AdvancedQuery.RuleGroup`.  If an exception is encountered at any point during parsing, remaining compilation is terminated and a String representation of the equivalent SQL query portion is included in the response.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-505)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -18,10 +22,16 @@ public static AdvancedFilterConfiguration fromReportFilter(ReportFilter filter)
...
 
-    return new AdvancedFilterConfiguration(filter.getId(), defaultValue);
+    AdvancedQuery.RuleGroup value = null;
+    try {
+      value = advQueryBuilder.build();
+    } catch (AdvancedQueryException e) {
+      LOGGER.log(System.Logger.Level.WARNING, "Error occurred while building AdvancedQuery", e);
+    }
+
+    return new AdvancedFilterConfiguration(filter.getId(), value, query);
```
mcmcgrath13: (q,b): how does the UI know if there is an error parsing the query?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdvancedFilterConfiguration.java
```diff
@@ -4,4 +4,13 @@
...
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long reportFilterUid,
-    AdvancedQuery.RuleGroup defaultValue) {}
+    AdvancedQuery.RuleGroup defaultValue,
+    //  The resulting SQL query portion extracted during construction of the RuleGroup.
+    //  Included for debugging purposes if the RuleGroup cannot be constructed.
+    String query) {
+  public AdvancedFilterConfiguration {
+    if (defaultValue == null && (query == null || query.isBlank())) {
+      throw new IllegalArgumentException("Either 'defaultValue' or 'query' must be provided");
+    }
```
mcmcgrath13: (nit, b): I don't think this is necessarily true - you can have no default values saved on a report filter and that's a valid state

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+  public AdvancedQueryBuilder(List<FilterValue> filterValues) {
+    this.filterValues = new ArrayList<>(filterValues);
+  }
+
+  public String generateQueryString() {
+    return "WHERE "
+        + String.join(
+            " ",
+            filterValues.stream()
+                .sorted(Comparator.comparing(FilterValue::getSequenceNumber))
```
mcmcgrath13, line 29: (sugg, nb): should we sort the filterValues on construction instead? this is important for correct building of the rule group as well as this?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+      new FilterValue(
+          2L, null, 1000, ReportConstants.FilterValueType.OPERATOR.toString(), null, ")", null);
+
+  public AdvancedQueryBuilder(List<FilterValue> filterValues) {
+    this.filterValues = new ArrayList<>(filterValues);
+  }
+
+  public String generateQueryString() {
+    return "WHERE "
+        + String.join(
```
mcmcgrath13:
```suggestion
    return String.join(
```

The `WHERE` isn't necessarily true since this is a fragment and it's cleaner to have this empty or not and have other consumers decide how to format/display

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+            filterValues.stream()
+                .sorted(Comparator.comparing(FilterValue::getSequenceNumber))
+                .map(
+                    f -> {
+                      String part;
+
+                      if (List.of("(", ")", "and", "or").contains(f.getOperator())) {
+                        part = f.getOperator();
+                      } else {
+                        part = "COL " + f.getOperator();
```
mcmcgrath13: (sugg, b): lookup the column name and put in the string

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+
+  public String generateQueryString() {
+    return "WHERE "
+        + String.join(
+            " ",
+            filterValues.stream()
+                .sorted(Comparator.comparing(FilterValue::getSequenceNumber))
+                .map(
+                    f -> {
+                      String part;
```
mcmcgrath13: (sugg, nb): maybe restructure this mapper to `if/else` on the `f.getValueType()` and if it's an operator, just print the operator and otherwise do a "([%s] %s %s)" type formatted string for the clauses

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+    if (hasNext()) {
+      throw new AdvancedQueryException("Extra FilterValue left over after parsing: " + peek());
+    }
+
+    AdvancedQuery query = simplify(rootRuleGroup);
+    LOGGER.log(System.Logger.Level.DEBUG, "AdvancedQuery following simplification: " + query);
+
+    if (query instanceof AdvancedQuery.Rule) {
+      return new AdvancedQuery.RuleGroup(
+          UUID.randomUUID().toString(), ReportConstants.QueryCombinators.and, List.of(query));
```
mcmcgrath13: (q, nb): how is this state possible given the parens wrapping on line 85?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+    } else {
+      return (AdvancedQuery.RuleGroup) query;
+    }
+  }
+
+  private AdvancedQuery.RuleGroup startRuleGroup() throws AdvancedQueryException {
+    LOGGER.log(System.Logger.Level.TRACE, "Starting rule group");
```
mcmcgrath13, line 126: (comment, nb): love the use of logging here ❤️

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+    FilterValue closingParen = peek();
+    if (!isCloseParen(closingParen)) throw new AdvancedQueryException("Expected closing paren");
+    advance();
+
+    return ruleGroup;
+  }
+
+  // ( col = 1 AND -> next token should be "(" OR CLAUSE
+  @SuppressWarnings(
+      "java:S3776") // Suppress "cognitive complexity" warning since this is inherently complex
```
mcmcgrath13, line 173: 😭

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  /** Advances the current index to the next {@code FilterValue} in the list */
+  private void advance() {
+    if (hasNext()) {
+      LOGGER.log(
+          System.Logger.Level.TRACE,
+          "Advancing from FilterValue " + current + " to " + current + 1);
+      current++;
+    }
```
mcmcgrath13, line 88: should there be an else here with some sort of debug or warning log that we tried to advance past the end?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+
+            // And attach it to the existing list of 'OR' group rules
+            rules.add(andGroup);
+          }
+        }
+
+      } else if (isOpenParen(filterValue)) {
+        rules.add(startRuleGroup());
+      } else {
+        throw new AdvancedQueryException("') invalid after OPERATOR");
```
mcmcgrath13: this error message may be too narrow - could we also land here on an `and and` type sequence?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+        }
+        return ruleGroup;
+      }
+
+      advance();
+    }
+
+    if (ruleGroup == null) {
+      ruleGroup = new AdvancedQuery.RuleGroup(UUID.randomUUID().toString(), combinator, rules);
+    }
```
mcmcgrath13: (q, nb): when do we end here? With a `a and` sequence? that seems like it should be an error

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -163,8 +167,10 @@ public ReportConfiguration getReport(Long reportUid, Long dataSourceUid) {
                       .filter(
                           f ->
                               f.getFilterCode()
-                                  .getFilterType()
-                                  .equals(ReportConstants.ADV_FILTER_TYPE))
+                                      .getFilterType()
+                                      .equals(ReportConstants.ADV_FILTER_TYPE)
+                                  && f.getFilterValues() != null
+                                  && !f.getFilterValues().isEmpty())
```
mcmcgrath13: (nit, b): empty advanced filters are valid state - revert this change

apps/modernization-api/src/test/java/gov/cdc/nbs/report/models/AdvancedFilterConfigurationTest.java
```diff
@@ -0,0 +1,15 @@
+package gov.cdc.nbs.report.models;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import org.junit.jupiter.api.Test;
+
+class AdvancedFilterConfigurationTest {
+  @Test
+  void should_throw_if_default_value_and_query_are_both_omitted() {
```
mcmcgrath13: this is valid state

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -18,10 +22,16 @@ public static AdvancedFilterConfiguration fromReportFilter(ReportFilter filter)
...
-    AdvancedQuery.RuleGroup defaultValue = null;
+    AdvancedQueryBuilder advQueryBuilder = new AdvancedQueryBuilder(filter.getFilterValues());
+    String query = advQueryBuilder.generateQueryString();
 
-    return new AdvancedFilterConfiguration(filter.getId(), defaultValue);
+    AdvancedQuery.RuleGroup value = null;
+    try {
+      value = advQueryBuilder.build();
+    } catch (AdvancedQueryException e) {
+      LOGGER.log(System.Logger.Level.WARNING, "Error occurred while building AdvancedQuery", e);
```
mcmcgrath13: (sugg, b): can we include the error message on the config so the UI can give that info to the user?

mcmcgrath13:
I tried saving some advanced filter of the form `a and b` in the 6 UI then loading in 7 and got the below
image

I think this is a latent bug - checking and may push up a quick fix here to enable continued testing

mcmcgrath13:
take two

image

image
image

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapper.java
```diff
@@ -33,15 +34,15 @@ public static BasicFilterConfiguration fromReportFilter(ReportFilter filter) {
     // For the future: A list of strings may end up being too simple for all use cases,
     // may need to evolve to be a small object with a key and value
     List<String> defaultValues = null;
-    Boolean defaultIncludeNulls = false;
+    boolean defaultIncludeNulls = false;
     if (filter.getFilterValues() != null) {
       defaultValues =
           filter.getFilterValues().stream()
-              .filter(v -> !v.getOperator().equals("ALLOW_NULLS"))
+              .filter(v -> !"ALLOW_NULLS".equals(v.getOperator()))
```
brick-green, line 41: ❤️

mcmcgrath13:
the regression test failure looks real, but I think it will resolve itself when the empty advanced filters are re-allowed

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: Ah, yeah good call, will do

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
JordanGuinn: My initial thinking was by the lack of `defaultValue`, but it sounds like [that's a viable non-error state](https://github.com/CDCgov/NEDSS-Modernization/pull/3312#discussion_r3414188809), so tbd.

I suppose I could include the error thrown?  Just not sure how helpful that would be to the UI, if we're effectively asking them to recreate anyway, and that might not be the only error present.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdvancedFilterConfiguration.java
JordanGuinn: Oh interesting - so an `ADV_WCB` report filter can have zero corresponding filter values?  Will update - though what does that actually _mean_ in practice?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: I initially avoided looking up the column name here because of this initially only happening in error states, and I wasn't sure of the value of having those col names vs the additional database operations.  Happy to update if we don't mind

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: It _shouldn't be_ in theory, just felt risky to cast `query` directly to `RuleGroup` without at least confirming what we expect.  Could throw an error instead?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: Will add!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
JordanGuinn: We can, though idk how helpful that would be, given (as I understand it) they'd still need to build the filters from scratch anyway, and that may very well not be the only issue with the previous filter state

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapper.java
JordanGuinn: I blame IntelliJ for this, cuz _wut_ 🫠

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/AdvancedFilterConfiguration.java
mcmcgrath13: yup! it means that the report allows the advanced filter to be used, but no default values have been saved to it

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
mcmcgrath13: end users will need to know what columns they were trying to filter on when re-creating the query in the new tool

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
mcmcgrath13: I think post-simplification we could maybe end up here if all we have is one rule group with one rule?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
mcmcgrath13: It may be helpful - I think better to include it

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapper.java
mcmcgrath13: I blame Java

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
mcmcgrath13: including the error is unambiguous and the UI can not show it if it turns out to be not-helpful

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: I'm not actually clear on when we hit encounter this anymore - all the tests pass whether or not it's present, including an additional test I added with an `a and` and `a and b and` sequence.  I can just remove it?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+  private final FilterValue firstOpenParen =
+      new FilterValue(
+          1L, null, 0, ReportConstants.FilterValueType.OPERATOR.toString(), null, "(", null);
+  private final FilterValue lastCloseParen =
+      new FilterValue(
+          2L, null, 1000, ReportConstants.FilterValueType.OPERATOR.toString(), null, ")", null);
+
+  public AdvancedQueryBuilder(List<FilterValue> filterValues, List<DataSourceColumn> columns) {
+    this.filterValues =
+        new ArrayList<>(
```
mcmcgrath13, line 27: (q, very-nb): why use an ArrayList here instead of the base list that was returned from `toList`? (eli5)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+                            .orElse(null);
+                    String columnName;
+
+                    if (column == null) {
+                      LOGGER.log(
+                          System.Logger.Level.WARNING,
+                          "Column not found for UID: " + f.getColumnUid());
+                      columnName = "UNKNOWN_COLUMN";
+                    } else {
+                      columnName = column.getColumnName();
```
mcmcgrath13:
```suggestion
                      columnName = column.getColumnTitle();
```

So we match the drop down options

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+                    }
+
+                    return String.join(" ", columnName, f.getOperator(), f.getValueTxt());
```
mcmcgrath13:
```suggestion
                    return "([%s] %s '%s')".formatted(columnName, f.getOperator(), f.getValueTxt());
```

for easier human parsing when these are bunched together in a string

mcmcgrath13 (commented):
Image

👏 👏 😭 🎉

mcmcgrath13 (approved):

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
JordanGuinn: Was encountering immutability errors just using `toList()`

emyl3:
Please ignore!! my frontend was not up to date 😅 


I keep getting a 422 error when I use rule groups
Screenshot 2026-06-16 at 13 48 11

This is what the request body looks like
```
{
    "isExport": false,
    "reportUid": 6,
    "dataSourceUid": 1,
    "basicFilters": [
        {
            "reportFilterUid": 16,
            "values": [
                "2006",
                "2026"
            ]
        },
        {
            "reportFilterUid": 14,
            "values": [
                "11065"
            ]
        },
        {
            "reportFilterUid": 15,
            "values": [
                "13"
            ]
        }
    ],
    "advancedFilter": {
        "reportFilterUid": 32,
        "value": {
            "id": "f4e9863a-9bb8-4d6c-83f5-b0843e66fe40",
            "combinator": "AND",
            "rules": [
                {
                    "id": "1ec5dbdd-e10a-4574-ba50-d5eda4a6016a",
                    "operator": "NN",
                    "columnId": 1,
                    "value": ""
                },
                {
                    "id": "ae256444-05bc-42f5-829f-85808ebcb1a0",
                    "combinator": "and",
                    "rules": [
                        {
                            "id": "03bbb5dc-39c4-4f17-93db-ba5bc173a139",
                            "operator": "NN",
                            "columnId": 2,
                            "value": ""
                        }
                    ]
                }
            ]
        }
    }
}
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/AdvancedQueryBuilder.java
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  /**
+   * Recursively construct a singular nested {@code RuleGroup} from the list of {@code FilterValue}s
+   * provided.
+   *
+   * @throws AdvancedQueryException if any of the FilterValue set is invalid
+   */
+  public AdvancedQuery.RuleGroup build() throws AdvancedQueryException {
+    LOGGER.log(System.Logger.Level.DEBUG, "Building query from: " + filterValues);
```
emyl3: (q, nb) Are we trying to print like a stringified version of `filterValues`? Will this work as-is? Or do we need to do something special to `filterValues`?

JordanGuinn: Oh interesting - I'm not sure how Jakarta Entities are logged by default, so I'll go ahead and just reference the `FilterValue` IDs for now

emyl3 (approved):
Thank you for all your work on this! 😭 🦾

PR #3315, opened 2026-06-12
[APP-526] feat: Use NBS components for Advanced Filter UI

emyl3:
## Description

- Use NBS components from design system for the advanced filter
- This PR does not address displaying the error handling in a nice way. All validation errors are in an Alert.

### Screenshot 📸 
Screenshot 2026-06-12 at 16 03 54
Screenshot 2026-06-12 at 16 04 00

## Tickets
* [APP-526](https://cdc-nbs.atlassian.net/browse/APP-526?atlOrigin=eyJpIjoiNWE2N2Y1NmNjNjIxNDViMzhhNWE0NmQxMTgzOTJmNzYiLCJwIjoiaiJ9
)

[APP-526]: https://cdc-nbs.atlassian.net/browse/APP-526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-ui/src/design-system/select/single/Select.tsx
```diff
@@ -22,7 +23,7 @@ const renderOptions = (options: Selectable[], placeholder?: string) => (
         )}
         {options?.map((item, index) => (
             <option key={index} value={item.value} aria-label={item.label}>
-                {item.name}
+                {useLabel ? item.label : item.name}
```
emyl3: ⚠️ needed this because the options list that querybuilder returns makes more sense to use the label value than the name

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
```diff
@@ -0,0 +1,126 @@
...
+                grid-column: span 5;
+            }
+
+            .trash-icon {
+                grid-column: span 1 / -2;
+            }
+        }
+
+        .queryBuilder-value {
+            grid-column: span 15;
```
emyl3: not sure how to get the value input to take up the remaining width 😭

apps/modernization-ui/src/design-system/date/criteria/range/DateRangeField.tsx
```diff
@@ -25,39 +17,22 @@ export type DateRangeFieldProps = {
...
-        [onChange, value?.between]
-    );
+        if (type === 'from') {
+            onChange({ between: { to: value?.between.to || undefined, from: v } });
+        }
+    };
 
     const dateRangePickerRef = useRef<HTMLDivElement>(null);
 
-    useLayoutEffect(() => {
```
emyl3, line 48: ⚠️ Had to remove since it was causing wonkiness when having multiple date range pickers

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
```diff
@@ -0,0 +1,126 @@
...
+
+            svg {
+                fill: colors.$error;
+            }
+
+            &:hover {
+                svg {
+                    fill: colors.$error-dark;
+                }
+            }
```
mcmcgrath13: (nit, maybe-b): can we use the "destructive" option instead of custom styling here? 

another trash/delete icon in `ManageSectionTile.tsx` looks like

```

                                <Button
                                    type="button"
                                    className={styles.iconBtn}
                                    outline
                                    disabled={onAction}
                                    onClick={() => {
                                        setSelectedForDelete(section);
                                        setOnAction(true);
                                    }}
                                >
                                    <Icon.Delete data-testId="deleteIcon" style={{ cursor: 'pointer' }} size={3} />
                                </Button>
```

I wonder if we should be coalescing on this format?

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
```diff
@@ -0,0 +1,126 @@
...
+
+            button {
+                grid-column: span 5;
+            }
+
+            .trash-icon {
+                grid-column: span 1 / -2;
+            }
```
mcmcgrath13: (nit, nb): could we use grid area templates with names instead?

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        .queryBuilder-value {
+            grid-column: span 15;
+        }
+
+        // Preview Where Clause styling
+        details {
+            div {
+                max-width: 90rem;
```
mcmcgrath13: (nit, b): can the `max-width` move to the layout instead? I notice we repeat this value. I that doesn't work, can we pull it out to a variable?

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
```diff
@@ -0,0 +1,126 @@
...
+                grid-column: span 1 / -2;
+            }
+        }
+
+        .queryBuilder-value {
+            grid-column: span 15;
+        }
+
+        // Preview Where Clause styling
+        details {
```
mcmcgrath13: (nit, b): the where clause preview isn't a summary/detail in the design, can we make it always display and match that?

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -302,27 +307,32 @@ const validateRule = (rule: RuleGroupTypeAny | RuleType | string, result: Valida
...
-                } else {
-                    const [startInt, endInt] = parts.map((v) => parseInt(v));
-                    if (startInt > endInt) setInvalid(id, 'High value must be greater than or equal to low value');
+            if (rule.value.trim() === '') {
+                setInvalid(id, 'Both low and high values required.');
+            }
+
+            const parts: string[] = rule.value.split(',');
+
+            if (parts[0] === '' || parts[1] === '') {
```
mcmcgrath13: (thought, very-nb): would it make sense to re-use the `DateRangeFilter.tsx`'s validator function here?

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -302,27 +307,32 @@ const validateRule = (rule: RuleGroupTypeAny | RuleType | string, result: Valida
...
+                setInvalid(id, 'Both low and high values required.');
+            }
+
+            if (isDate(parts[0]) && isDate(parts[1])) {
+                const [startDt, endDt] = parts.map((v) => new Date(v));
+
+                if (startDt > endDt) {
+                    setInvalid(id, 'High value must be greater than or equal to low value.');
```
mcmcgrath13: (nit, b): invalid dates aren't handled here and need to be

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -302,27 +307,32 @@ const validateRule = (rule: RuleGroupTypeAny | RuleType | string, result: Valida
...
                 }
+            } else {
+                const [startInt, endInt] = parts.map((v) => parseInt(v));
```
mcmcgrath13: (nit, b): need to check for bad parse (nan maybe?)

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -367,7 +377,14 @@ const AdvancedFilter = ({ filter, columns }: { filter: AdvancedFilterConfigurati
...
                         autoSelectOperator={false}
                         autoSelectValue={false}
-                        controlElements={{ dragHandle: ShiftableDragHandle, valueEditor: ValueEditor }}
+                        controlElements={{
+                            dragHandle: ShiftableDragHandle,
+                            valueEditor: ValueEditor,
+                            fieldSelector: ValueSingleSelector,
+                            operatorSelector: ValueSingleSelector,
+                            combinatorSelector: ValueSingleSelector,
+                            actionElement: AdvancedFilterButton,
```
mcmcgrath13: (nit, b): can we set the things below to control the trash can situation instead of doing via styling

Image

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilterButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { ActionProps } from 'react-querybuilder';
+import { Button } from '../../../../../design-system/button';
+
+const getLabelName = (props: ActionProps) => {
+    if (props.className === 'ruleGroup-addGroup') {
+        return 'Add rule group';
+    } else {
+        return props.title;
+    }
+};
```
mcmcgrath13: (nit, nb): could we use `translations` prop instead of intervening here? https://react-querybuilder.js.org/docs/components/querybuilder#translations

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilterButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+const AdvancedFilterButton = (props: ActionProps) => {
+    switch (props.className) {
+        case 'ruleGroup-addRule':
+        case 'ruleGroup-addGroup':
+            return (
+                <Button
```
mcmcgrath13: (nit, nb): maybe use the `addRuleAction` and `addGroupAction` https://react-querybuilder.js.org/docs/components/querybuilder#translations props instead of a switch in this component

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueEditor.tsx
```diff
@@ -1,12 +1,14 @@
...
+const ValueEditor = (props: ValueEditorProps<ValueSetMetadata & FullField & FullOperator>) => {
     switch (props.type) {
-        case 'select':
+        case 'text':
+            return <ValueInput {...props} />;
         case 'multiselect':
-            return <ValueSetSelector {...props} />;
+            return <ValueSetSelector {...props} required />;
         default:
             return <DefaultValueEditor {...props} />;
```
mcmcgrath13: should we ever hit this default? maybe put a comment that we shouldn't?

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSetSelector.tsx
```diff
@@ -32,12 +32,15 @@ const CODE_DESC_CD = {
     DESCRIPTION: 'd',
 };
 
-const ValueSetSelector = (props: ValueEditorProps<ValueSetMetadata & FullField>) => {
+type Required = {
+    required?: boolean;
```
mcmcgrath13: (nit/q, nb): is required always true? (looks like it from the hard coded `required` prop) - if so, can we just hard code the required here instead of earlier and then remove this extra plumbing?

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useId } from 'react';
+import { FullField, ValueEditorProps } from 'react-querybuilder';
+import { SingleSelect } from '../../../../../design-system/select';
+import { Selectable } from '../../../../../options';
+
+const getLabel = (props) => {
+    if (props.className === 'rule-operators') return 'Logic';
+    return props.title || '';
+};
```
mcmcgrath13: (nit, b): use translation instead

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const getPlaceholder = (props) => {
+    if (props.className === 'ruleGroup-combinators') return ''; // use no placeholder
+    return undefined; // use default placeholder
+};
```
mcmcgrath13: (nit, b): use translations instead

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+        if (value === null) return;
+        props.handleOnChange(value.value ?? '');
+    };
+
+    let currentSelection: Selectable | null;
+
+    if (typeof props.value === 'string') {
+        currentSelection = props.options.find((opt) => opt.value === props.value) || null;
+    } else {
+        currentSelection = props.value;
```
mcmcgrath13: (q, nb): what is the non-string type here? could we instead just always try to find the option and default to null?

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    let options;
+    switch (props.className) {
+        case 'rule-operators':
+            options = props.fieldData.operators
+                .map((operator) => {
+                    return props.options.find((opt) => opt.name === operator.name);
+                })
+                .filter(Boolean);
+            break;
```
mcmcgrath13:
```suggestion
            options = props.options.filter((opt) => !!props.fieldData.operators.find(operator => operator.name == opt.name));
            break;
```
(nit, very-nb): I find this ordering more readable

apps/modernization-ui/src/apps/report/run/report-configuration-page.module.scss
```diff
@@ -0,0 +1,15 @@
+@use 'styles/borders';
+@use 'styles/buttons';
+@use 'styles/colors';
+
+.card {
+    header {
+        div {
+            max-width: 90rem;
+
+            span:first-of-type {
```
mcmcgrath13: (nit, b): this is going to be a much broader layout thing than just report module - can you add a prop to the `Card` component instead of `widthContent` or similar and update the styles there instead?

apps/modernization-ui/src/design-system/date/criteria/range/DateRangeField.tsx
mcmcgrath13: (q, nb): Given that, is the `dateRangePickerRef` still needed?

apps/modernization-ui/src/design-system/date/criteria/range/DateRangeField.tsx
```diff
@@ -25,39 +17,22 @@ export type DateRangeFieldProps = {
...
 
 const DateRangeField = ({ id, value, sizing, onChange, onBlur, label, required }: DateRangeFieldProps) => {
-    const [range, setRange] = useState<DateRange | undefined>(value?.between);
-
-    useEffect(() => {
-        setRange(value?.between);
-    }, [value, setRange]);
+    const handleFieldOnChange = (v, type) => {
+        if (type === 'to') {
+            onChange({ between: { to: v, from: value?.between.from || undefined } });
```
mcmcgrath13: (q, b): does this logic mean if `v` is '', then we can set `to` to `''` on the first field change, but then if we change `from`, `v` is going to change to `undefined`? that feels a bit fussy - can we drop the `|| undefined`?

apps/modernization-ui/src/design-system/field/VerticalField.tsx
```diff
@@ -17,14 +17,21 @@ type Props = {
...
+
 const VerticalField = ({ className, htmlFor, label, helperText, required, error, warning, children }: Props) => (
     <span className={classNames(styles.entry, className)}>
-        <span className={styles.labels}>
-            <label className={classNames({ [styles.required]: required })} htmlFor={htmlFor}>
-                {label}
-            </label>
-            {helperText && <HelperText id={`${htmlFor}-hint`}>{helperText}</HelperText>}
-        </span>
+        {displayLabel(label, required) && (
```
mcmcgrath13:
```suggestion
        {(!!label || required) && (
```

apps/modernization-ui/src/design-system/field/VerticalField.tsx
```diff
@@ -17,14 +17,21 @@ type Props = {
     children: ReactNode;
 };
 
+const displayLabel = (label, required) => {
+    if (label) return Boolean(label);
+    return required && Boolean(label);
+};
```
mcmcgrath13:
```suggestion
```

apps/modernization-ui/src/design-system/select/single/Select.tsx
mcmcgrath13: (q, b): can we change how we set up the options instead? (this is typically how we handle this with other components using select)

mcmcgrath13:
image

I have IDE squiggles - do you?

mcmcgrath13:
image

mcmcgrath13:
when I select a field, then try to go back to the un-selected state, nothing happens

mcmcgrath13:
image

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
emyl3: Fixed! Thank you, Mary!

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
emyl3: Thank you for doing that [here](https://github.com/CDCgov/NEDSS-Modernization/pull/3315/changes#diff-78e5373f894860bd9bbd6fcbce1918bbbc0da0bedb89240c7486dd2e0cf5489aR72)! TIL! 💡

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
emyl3: Moved it up to the layout class instead and continued to work as expected!

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
emyl3: Done!

apps/modernization-ui/src/design-system/field/VerticalField.tsx
emyl3: This doesn't work since I want to only display the required asterisk if a label is present for the between clauses where I want the overarching label to not be present if there are the sub label "From" and "To" if that makes sense...

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
emyl3: [overridden](https://github.com/CDCgov/NEDSS-Modernization/pull/3315/changes#diff-ad37db62917784eb488db7d5abcbc7771540bb783f620dfd1c943d50a45c03a6R355)! Thank you!

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilterButton.tsx
emyl3: Done! https://github.com/CDCgov/NEDSS-Modernization/pull/3315/changes#diff-ad37db62917784eb488db7d5abcbc7771540bb783f620dfd1c943d50a45c03a6R352-R357

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSetSelector.tsx
emyl3: Removed and hardcoded

apps/modernization-ui/src/design-system/date/criteria/range/DateRangeField.tsx
emyl3: removed!

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
emyl3: used the validator there and separated it out into a util file

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
emyl3: Done!

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
emyl3: Done!

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
emyl3: Updated!

apps/modernization-ui/src/design-system/select/single/Select.tsx
emyl3: Removed this implementation and did [this instead](https://github.com/CDCgov/NEDSS-Modernization/pull/3315/changes#diff-2302bbe1ab21bbde81ab8b68ca3b9df854b2bce3315cc286a1c42ff882e30c7eR29-R32)

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueEditor.tsx
emyl3: we shouldn't hit this! Comment added!

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
emyl3: Removed!

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
emyl3: There isn't haha oops. Removed!

emyl3:
> when I select a field, then try to go back to the un-selected state, nothing happens

This should be fixed now!

emyl3:
> image

This should be fixed!

apps/modernization-ui/src/design-system/date/criteria/range/DateRangeField.tsx
emyl3: Cleaned this up!

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
emyl3: Done!

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilterButton.tsx
emyl3: Done!

apps/modernization-ui/src/apps/report/run/filters/advanced/advanced-filter.module.scss
emyl3: I used the destructive prop on the button and removed that styling!
The button in  `ManageSectionTile.tsx` looks slightly different than the designs so I am leaning towards keeping the current implementation as-is. What do you think?

emyl3:
> image

I am no longer seeing squiggles with the new implementation. Let me know if you are still seeing it 👁️ 👁️

emyl3:
> image
> 
> I have IDE squiggles - do you?

Let me know if you still see them with the new implementation!

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -348,13 +336,13 @@ const AdvancedFilter = ({ filter, columns }: { filter: AdvancedFilterConfigurati
     } = useController<ReportExecuteForm, 'advancedFilter'>({
         name: 'advancedFilter',
         defaultValue: filter.defaultValue ? advancedFilterConfigToQuery(filter.defaultValue, columns) : EMPTY_QUERY,
-        rules: { validate: validateAdvancedFilter },
+        rules: { validate: (values) => validateAdvancedFilter(columns, values) },
```
mcmcgrath13, line 228:
```suggestion
        rules: { validate: (val) => validateAdvancedFilter(columns, val) },
```
(nit, very-nb): for some reason the `s` really bugs be as it isn't a plural thing, `value` is a bit overloaded, but not technically problematic. `query` is another option?

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -367,7 +355,17 @@ const AdvancedFilter = ({ filter, columns }: { filter: AdvancedFilterConfigurati
                         autoSelectField={false}
                         autoSelectOperator={false}
                         autoSelectValue={false}
-                        controlElements={{ dragHandle: ShiftableDragHandle, valueEditor: ValueEditor }}
+                        translations={translationOverrides}
+                        controlElements={{
+                            dragHandle: ShiftableDragHandle,
+                            valueEditor: ValueEditor,
+                            valueSelector: ValueSingleSelector,
+                            // actionElement: RemoveButton,
```
mcmcgrath13:
```suggestion
```

apps/modernization-ui/src/apps/report/run/filters/advanced/RemoveButton.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { ActionProps } from 'react-querybuilder';
+import { Button } from '../../../../../design-system/button';
+
+const RemoveButton = (props: ActionProps) => {
+    return (
+        <Button
+            icon="delete"
+            type={'button'}
+            className={'trash-icon'}
```
mcmcgrath13:
```suggestion
            type="button"
            className="trash-icon"
```

apps/modernization-ui/src/apps/report/run/filters/advanced/RemoveButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+            tertiary
+            destructive
+            sizing={'small'}
```
mcmcgrath13:
```suggestion
            sizing="small"
```

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -284,52 +309,6 @@ const validator: QueryValidator = (q) => {
     return result;
 };
 
-const isDate = (val: string) => !!val.match(/\d{4}-\d{2}-\d{2}/);
```
emyl3, line 287: ⚠️ moved to separate validator file

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,74 @@
...
+    const setInvalid = (id: string, reason: string) => {
+        result[id]['valid'] = false;
+        result[id]['reasons'] = [reason];
+    };
+    if (isRuleType(rule)) {
+        const id = rule['id'];
+        const field = rule['field'];
+        const operator = rule['operator'];
+        const value = rule['value'];
+        const label = rule['label'];
```
mcmcgrath13:
```suggestion
        const {id, field, operator, value, label} = rule;
```

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,74 @@
...
+            if (isDateFormat(parts[0]) || isDateFormat(parts[1])) {
+                rangeErrorMsg = validateDateRange(parts, label);
+            } else {
+                rangeErrorMsg = validateNumericRange(parts, label);
+            }
+
+            if (rangeErrorMsg) setInvalid(id, rangeErrorMsg);
+            return;
+        } else if (BINARY_OPERATORS.find((name) => name === operator)) {
+            if (value === '' || value.includes(',')) {
```
mcmcgrath13: (q, b): why check for the `,` seems like including a comma in a true string contains should be fine

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,74 @@
...
+                setInvalid(id, getMissingValErrorMsg(label, false));
+            }
+        }
+    } else if (isRuleGroupType(rule)) {
+        const ruleGroupRules = rule['rules'];
+        ruleGroupRules.forEach((r) => validateRule(r, result));
```
mcmcgrath13:
```suggestion
        rule.rules.forEach((r) => validateRule(r, result));
```

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+    number: NumericInput,
+    text: TextInputField,
+} as const;
+
+const getConvertedRange = (props): DateBetweenCriteria | NumberBetweenCriteria => {
+    if (props.operator === 'between' && typeof props.value === 'string' && props.value) {
+        if (props.operator === 'between' && props.value) {
+            const [from = '', to = ''] = props.value.split(',');
+            return { between: { from, to } };
+        }
```
mcmcgrath13: (nit, b): the inner if seems to fully repeat the outer - clean up

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+    }, [props.handleOnChange]);
+
+    // clean up value when value input is removed from DOM
+    useLayoutEffect(() => {
+        return () => {
+            if (!props.value) {
+                onChangeRef.current('');
+            }
+        };
+    }, []);
```
mcmcgrath13: (q, b): why do we need a ref for this?

mcmcgrath13: the blocking part is this whole chunk needs much more comment (but if we could not use the ref, also great)

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useId } from 'react';
+import { FullField, ValueEditorProps } from 'react-querybuilder';
+import { SingleSelect } from '../../../../../design-system/select';
+import { Selectable } from '../../../../../options';
+
+const ValueSingleSelector = (props: ValueEditorProps<FullField>) => {
+    const id = useId();
+    const title = props['title'] ?? '';
+    let options = props['options'] ?? [];
```
mcmcgrath13:
```suggestion
    const title = props.title ?? '';
    const options = props.options ?? [];
```

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    const handleOnChange = (value: Selectable | null) => {
+        if (value === null) return;
+        props.handleOnChange(value['value'] ?? '');
```
mcmcgrath13:
```suggestion
        props.handleOnChange(value.value ?? '');
```

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+    };
+
+    let currentSelection: Selectable | null;
+
+    currentSelection = options.find((opt) => opt.value === props.value) || '';
```
mcmcgrath13:
```suggestion
    const currentSelection: Selectable | null = options.find((opt) => opt.value === props.value) || null;
```
maybe on the `'' => null` part, but `''` isn't a selectable

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    if (props.className === 'rule-operators') {
+        let availableOperators = props['fieldData']['operators'];
```
mcmcgrath13:
```suggestion
        const availableOperators = props.fieldData.operators;
```

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+        options = options.filter(
```
mcmcgrath13: (nit, b): we shouldn't reassign to options, it came from props, so this has a weird code smell

maybe create a new 'availableOptions` or something and use it here and line 28, which has the same code smell

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
```diff
@@ -0,0 +1,50 @@
...
+            // retain placeholder and operator options available on field selection
+            (opt) => !!availableOperators.find((operator) => operator.name == opt.name || opt.name === '~')
+        );
+    }
+
+    options = options.map((opt) => ({
+        ...opt,
+        name: opt.label,
+    }));
```
mcmcgrath13: (q, nb): why do we need to do this re-name here - should we process the incoming options differently?

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,65 @@
...
+            return `${str} value of "${value[ind]}" is not a valid number for ${field}.`;
+        }
+    }
+
+    if (fromNum > toNum) return getBeforeErrorMsg(field, false);
+};
+
+export const getRangeValErrorMsg = (field: string, isInvalidVal: boolean) => {
+    const invalidTypeMsg = isInvalidVal ? 'valid ' : '';
+    return `Enter ${invalidTypeMsg}from and to values for ${field}.`;
```
mcmcgrath13: (q, nb): when we this not be invalid?

apps/modernization-ui/src/design-system/field/VerticalField.tsx
mcmcgrath13: my brain computer says this is the same logic - where does it break down?

mcmcgrath13: or should it just be `!!label` since for the `displayLabel` if label is truthy, it will always return true, and if label is falsey it will always return false?

apps/modernization-ui/src/design-system/input/range/NumberInputRange.tsx
```diff
@@ -0,0 +1,35 @@
+import { Field, FieldProps } from 'design-system/field';
+import { NumberRangeField, NumberRangeFieldProps } from './NumberRangeField.tsx';
+
+type NumberInputRangeProps = FieldProps & NumberRangeFieldProps;
+
+/**
+ * @return {JSX.Element}
+ */
+const NumberInputRange = ({
```
mcmcgrath13: (nit, b): this seems to overlap with `NumericInput` from `apps/modernization-ui/src/design-system/input/numeric/NumericInput.tsx` - can we use that instead? (maybe adapt as needed?)

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueSingleSelector.tsx
emyl3: I believe if we do that then the preview where clause values get messed up... 😓

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueInput.tsx
emyl3: I needed a way to clear the value when the value input changes because the logic operator is updated so the validations made sense 😓 What do you think?

mcmcgrath13: doing something more in line with https://github.com/mcmcgrath13/react-querybuilder/blob/mcm/dnd-kit-key-sort/packages/react-querybuilder/src/components/ValueEditor.tsx#L313 would be much more idiomatic react

apps/modernization-ui/src/apps/report/run/report-configuration-page.module.scss
emyl3: Done!

apps/modernization-ui/src/design-system/field/VerticalField.tsx
emyl3: Ahhh yes!! Thank you it was just `!!label`

apps/modernization-ui/src/apps/report/run/filters/advanced/AdvancedFilter.tsx
```diff
@@ -37,120 +40,6 @@ const EMPTY_QUERY: QbRuleGroup = {
     rules: [{ id: crypto.randomUUID(), field: '~', operator: '~', value: '' }],
 };
 
-type NbsOperator = Operator & { nbsCd: string };
```
emyl3, line 40: ⚠️ moved to a separate operators.ts file

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
emyl3: Shouldn't hit this but just in case: https://github.com/CDCgov/NEDSS-Modernization/pull/3315/changes#diff-251967c1386470a4a89526ccd126821bc9e2fc439d9b666a087db13815556a9cR32

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueInput.tsx
emyl3: Done!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
emyl3: Removed!

apps/modernization-ui/src/design-system/input/range/NumberInputRange.tsx
emyl3: Updated!

apps/modernization-ui/src/apps/report/run/filters/advanced/AddButton.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { ActionProps } from 'react-querybuilder';
+import { Button } from '../../../../../design-system/button';
```
JordanGuinn, line 3: (thought, super-nb): ENTIRELY unrelated to this PR, but it'd be great if we could get path aliasing working at some point, and remove the need to traverse up the codebase like this for some of our dependencies!

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueEditor.tsx
```diff
@@ -1,14 +1,14 @@
-import { ValueEditor as DefaultValueEditor, FullField, ValueEditorProps } from 'react-querybuilder';
+import { FullField, FullOperator, ValueEditorProps } from 'react-querybuilder';
 import { ValueSetSelector } from './ValueSetSelector';
 import { ValueSetMetadata } from './AdvancedFilter';
+import { ValueInput } from './ValueInput.tsx';
 
-const ValueEditor = (props: ValueEditorProps<ValueSetMetadata & FullField>) => {
+const ValueEditor = (props: ValueEditorProps<ValueSetMetadata & FullField & FullOperator>) => {
```
JordanGuinn: (nit, nb): Thoughts on renaming our `ValueEditor` something that isn't identical to what's provided by `react-querybuilder`?  Just to indicate that they are _slightly_ different, and remove the potential for anyone using the `react-querybuilder` version by mistake

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const validateRule = (rule: RuleGroupTypeAny | RuleType | string, result: ValidationResultMap) => {
+    const setInvalid = (id: string, reason: string) => {
+        result[id]['valid'] = false;
+        result[id]['reasons'] = [reason];
+    };
+    if (isRuleType(rule)) {
+        const { id, field, operator, value, label } = rule;
+
+        if (!id) return; // no key for the map, shouldn't happen in practice
```
JordanGuinn: (nit, nb): I wonder if we want to log a warning or throw or something here, if this shouldn't happen in practice?

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,70 @@
...
+        // default valid
+        result[id] = { valid: true };
+
+        // empty rules are fine
+        if (!field || field === '~') return;
```
JordanGuinn, line 25: (q, nb): Out of curiosity, why are empty rules fine?  Do they get compacted at a later point prior to hitting the API?

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const id = useId();
+    const { handleOnChange, inputType, operator, title, value } = props;
+    const labelName = title ?? '';
+    const isBetween = operator === BETWEEN_OPERATOR_NAME;
+    const InputComponent = isBetween ? RANGE_COMPONENTS[inputType] : SINGLE_COMPONENTS[inputType];
+
+    let convertedValue = isBetween ? getConvertedRange(props) : (value ?? '');
+
+    // eslint-disable-next-line max-len
+    // adapated from https://github.com/mcmcgrath13/react-querybuilder/blob/87a991b124fa9060431ac8e1e8f42b789a5ddecb/packages/react-querybuilder/src/components/ValueEditor.tsx#L312-L322
```
JordanGuinn:
```suggestion
    // adapted from https://github.com/mcmcgrath13/react-querybuilder/blob/87a991b124fa9060431ac8e1e8f42b789a5ddecb/packages/react-querybuilder/src/components/ValueEditor.tsx#L312-L322
```

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,65 @@
+export const validateDateRange = (value, field) => {
+    if (rangeValuesMissing(value)) return getRangeValErrorMsg(field, false);
```
JordanGuinn, line 6: (thought, nb): I feel like there's _gotta_ be a way to assert the presence of `value[0]` and `value[1]` here with an [assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) on `rangeValuesMissing`, but a thought for another time!

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    const fromDate = new Date(value[0]!); // can't be undefined because of above checks
+    const toDate = new Date(value[1]!); // can't be undefined because of above checks
+    for (const { date, ind, str } of [
+        { date: fromDate, ind: 0, str: 'From' },
```
JordanGuinn: (nit, nb): Could we just reference the underlying value in the array, rather than including an index in the list we're iterating against?

Also, given this is just two entries, it could arguably be a bit simpler if it was just two `if` statements, without the for loop entirely?

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const isNumberFormat = (val) => {
+    if (typeof val === 'string') {
+        return !!val.match(/^-?\d+(\.\d+)?$/);
+    }
+
+    return typeof val === 'number';
+};
+
+export const isDateFormat = (val) => {
```
JordanGuinn: (q, nb): I think [date-fns](https://date-fns.org/) [supports something like this](https://date-fns.org/v4.4.0/docs/isMatch) out of the box?  Assuming it supports the formats we're matching against here

JordanGuinn (commented):
Handful of questions/thoughts, but this is looking good!!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
emyl3: It gets ignored and we do not send an empty rule over.

apps/modernization-ui/src/apps/report/run/filters/advanced/AddButton.tsx
emyl3: Yes! Agree! Do you prefer I create a separate ticket to keep track of this or just a good thing to keep in mind when we are out of the crunch period? 👀

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
emyl3: added a console.warn! Thank you!!

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueEditor.tsx
emyl3: Renamed to ValueEditorSwitch 😅 let me know if you have a different name in mind 👀

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
emyl3: Took your first suggestion and passed the value directly instead of the index! Thank you!!

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
emyl3: Modified it to use date-fns to validate the date and caught some edge cases in the process -- thank you!!

emyl3:
@JordanGuinn @brick-green ready for re-review!

apps/modernization-ui/src/apps/report/run/filters/advanced/AddButton.tsx
JordanGuinn: A tech debt ticket couldn't hurt!  Thanks Elisa 🙏

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueEditor.tsx
JordanGuinn: Works for me!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,79 @@
...
+            let rangeErrorMsg;
+            if (type === 'DATETIME') {
+                rangeErrorMsg = validateDateRange(parts, label);
+            } else {
+                rangeErrorMsg = validateNumericRange(parts, label);
+            }
+
+            if (rangeErrorMsg) setInvalid(id, rangeErrorMsg);
+            return;
+        } else if (BINARY_OPERATORS.find((name) => name === operator)) {
```
JordanGuinn: (the tiniest nit, nb):


```suggestion
        } else if (BINARY_OPERATORS.includes(operator)) {
```

apps/modernization-ui/src/apps/report/run/filters/advanced/ValueInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const isBetween = operator === BETWEEN_OPERATOR_NAME;
+    const InputComponent = isBetween ? RANGE_COMPONENTS[inputType] : SINGLE_COMPONENTS[inputType];
+
+    let convertedValue = isBetween ? getConvertedRange(props) : (value ?? '');
+
+    // eslint-disable-next-line max-len
+    // adapted from https://github.com/mcmcgrath13/react-querybuilder/blob/87a991b124fa9060431ac8e1e8f42b789a5ddecb/packages/react-querybuilder/src/components/ValueEditor.tsx#L312-L322
+    useEffect(() => {
+        // clear input when changing from between to other operators
+        if (![BETWEEN_OPERATOR_NAME].includes(operator) && value.includes(',')) {
```
JordanGuinn: (nit, nb):

```suggestion
        if (BETWEEN_OPERATOR_NAME !== operator && value.includes(',')) {
```

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,88 @@
...
+const isNumberFormat = (val) => {
+    if (typeof val === 'string') {
+        return !!val.match(/^-?\d+(\.\d+)?$/);
+    }
+
+    return typeof val === 'number';
+};
+
+export const isDateFormat = (val: string): boolean => {
+    const formats = ['MM/dd/yyyy', 'MM/yyyy', 'yyyy'];
```
JordanGuinn: (super tiny nit, nb): Maybe extract this `formats` const to the top of the file, or at least out of this function?  Just so it's not reinstantiated every time this method is invoked, given it's static.

JordanGuinn (approved):
Looks good to me! 🚀

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
emyl3: Done!

apps/modernization-ui/src/apps/report/run/filters/advanced/AddButton.tsx
emyl3: Here is the ticket! https://cdc-nbs.atlassian.net/browse/APP-776

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,79 @@
...
+            } else {
+                rangeErrorMsg = validateNumericRange(parts, label);
+            }
+
+            if (rangeErrorMsg) setInvalid(id, rangeErrorMsg);
+            return;
+        } else if (BINARY_OPERATORS.includes(operator)) {
+            if (typeof value === 'string') {
+                if ((value as string).trim() === '') {
+                    setInvalid(id, getMissingValErrorMsg(label, false));
```
brick-green, line 58: q, maybe b: should there be a `return` here so that the next `if` for DATETIME doesn't need to run?

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,79 @@
+import { isRuleGroupType, isRuleType, RuleGroupTypeAny, RuleType, ValidationResult } from 'react-querybuilder';
+import { getRangeValErrorMsg, isDateFormat, validateDateRange, validateNumericRange } from '../utils/rangeValidator.ts';
+import { BINARY_OPERATORS } from './operators.ts';
+
+export type ValidationResultMap = Record<string, ValidationResult>;
+
+export const validateRule = (rule: RuleGroupTypeAny | RuleType | string, result: ValidationResultMap) => {
```
brick-green: q, maybe b: rule has three types here `RuleGroupTypeAny | RuleType | string` but it looks like the code only checks for RuleType and RuleGroupType. What is the `string` type for?

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
```diff
@@ -0,0 +1,79 @@
...
+            if (rangeErrorMsg) setInvalid(id, rangeErrorMsg);
+            return;
+        } else if (BINARY_OPERATORS.includes(operator)) {
+            if (typeof value === 'string') {
+                if ((value as string).trim() === '') {
+                    setInvalid(id, getMissingValErrorMsg(label, false));
+                }
+
+                if (type === 'DATETIME') {
+                    if (!isDateFormat(value)) setInvalid(id, getMissingValErrorMsg(label, false));
```
brick-green: thought, b: the missing value error message doesn't seem right here since it might be due to an malformed but present date.

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (fromDate > toDate) return getBeforeErrorMsg(field, true);
+    return;
+};
+
+export const validateNumericRange = (value, field) => {
+    if (rangeValuesMissing(value)) return getRangeValErrorMsg(field, false);
+
+    const fromNum = parseInt(value[0]); // can't be undefined because of above checks
+    const toNum = parseInt(value[1]); // can't be undefined because of above checks
```
brick-green: q, nb: will the range values always be integers? Then either way can you add a test case for decimals?

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export const getRangeValErrorMsg = (field: string, isInvalidVal: boolean) => {
+    const invalidTypeMsg = isInvalidVal ? 'valid ' : '';
+    return `Enter ${invalidTypeMsg}from and to values for ${field}.`;
+};
+
+export const getBeforeErrorMsg = (field: string, isDate: boolean) => {
+    const typeMsg = isDate ? 'date' : 'value';
+    return `From ${typeMsg} must be before to ${typeMsg} for ${field}.`;
```
brick-green:
```suggestion
    return `From ${typeMsg} must be before To ${typeMsg} for ${field}.`;
```

brick-green (commented):
A few questions on the validation files

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
emyl3: Done and updated other instances of that as well!

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.ts
emyl3: Not necessarily! Yes, I've added a test for decimals!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
emyl3: OOO nice catch updated the message!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
emyl3: OOps I didn't need that so it has been removed!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.ts
emyl3: Oh! Good call!

emyl3:
@brick-green ready for re-review! I also caught YET ANOTHER edge case where having nested empty rule groups or rule groups with more empty rule groups would cause issues so I fixed that up and wrote a test for that... 😓

brick-green (approved):
🚀

apps/modernization-ui/src/apps/report/run/filters/advanced/AddButton.tsx
mcmcgrath13: We do have path aliasing! kinda - it's janky and problematic in some ways and maybe needs to be fixed up in the future though, but that's a future problem

All of the top level folders can be directly accesssed so you can do `design-system/button` instead of all the `../..` ing

PR #3317, opened 2026-06-15
[APP-628] Convert PA02 Library

krista-skylight:
- [ ] Add the migration to `apps/modernization-api/src/main/resources/db/report-execution-changelog.yml`. It should be added to the latest `changeSet` since the last release - this could require a new `changeSet` if there isn't one since last release
- [ ] Keep an eye out for things in the sas library that are/could be centralized to pre- or post-processing or a util would be helpful for (e.g. common formatting needs)


apps/report-execution/tests/integration/assets/tablefaker_schema/std_hiv_datamart.yaml
```diff
@@ -40,19 +40,20 @@ tables:
...
+      - column_name: referral_basis
+        data: |
+          # For test rows, set to NULL to use fl_fup_exam_dt directly
+          if row_id < 6 or fake.random_int(1, 100) >= 95:
+              return None
+          else:
+              return fake.random_element(['P1 - Partner, Sex', 'P2 - Partner, Needle-Sharing','P3 - Partner, Both', "T1 - Positive Test", "T2 - Morbidity Report"])
 
   - table_name: "[RDB].[dbo].[D_PROVIDER]"
     row_count: 250
```
adaball, line 52: suggestion: lower this count so the snapshot yaml and output size becomes more reasonable.

In PA01 I worked with 5 providers and it limited the PDF output to ~12 pages while still accomplishing the goal of testing the data.  I can see from running PA02 SAS report the output is > 220 pages.

apps/report-execution/src/libraries/pa_02.py
```diff
@@ -0,0 +1,362 @@
...
+        metric_labels.append((f'Dispo {letter}:', var_name))
+
+    if report_type == 'STD':
+        metric_labels.append(('Dispo E:', 'var_ac_p'))
+
+    metric_labels.append(('Open:', 'var_ad_p'))
+    metric_labels.append(('Non-assigned Dispos:', 'var_ae_p'))
+
+    # ------------------------------------------------------------------
+    # 7. Build final table (no ALL rows)
```
adaball, line 297: question: is there a reason you're not including "ALL" rows?

I can see in the [Complex STD Report Outputs spreadsheet](https://docs.google.com/spreadsheets/d/1YpXfOimCcqe-vjrnyz6IB8b6t1W6G8Ajok_X8gac4Cg/edit?gid=1552969357#gid=1552969357) that PA02 has rows for "ALL".

apps/report-execution/src/libraries/pa_02.py
```diff
@@ -0,0 +1,362 @@
...
+            content=Table(
+                columns=[
+                    'PROVIDER_QUICK_CODE_new',
+                    'colname',
+                    'colval',
+                    'colval2',
+                    'colval3',
+                    'colval4',
+                    'pname_l',
+                    'i',
```
adaball: suggestion: since these values are used multiple times throughout this file, pull them into a constant.

apps/report-execution/src/libraries/support/pa_02/queries.py
```diff
@@ -0,0 +1,180 @@
...
+    WITH a AS ({subset_query}),
+    base AS (
+        SELECT DISTINCT a.INVESTIGATOR_DISP_FL_FUP_KEY
+        FROM a
+        INNER JOIN [RDB].[dbo].INVESTIGATION b
+            ON a.INVESTIGATION_KEY = b.INVESTIGATION_KEY
+        WHERE a.INVESTIGATOR_FL_FUP_KEY != 1
+          AND a.INVESTIGATOR_DISP_FL_FUP_KEY IS NOT NULL
+    )
+    SELECT INVESTIGATOR_DISP_FL_FUP_KEY FROM base
```
adaball: suggestion: remove the `base` CTE wrapping along with line 40 and just use the query within the `base` CTE.

apps/report-execution/tests/integration/libraries/pa_02.py
```diff
@@ -0,0 +1,425 @@
...
+        for provider, count in metrics_per_provider.items():
+            assert count == expected_metrics, (
+                f'Provider {provider} has {count} metrics, expected {expected_metrics}'
+            )
+
+    def test_execute_report_check_data(self, snapshot):
+        """Snapshot test for STD report."""
+        report_spec = self.create_spec()
+        result = execute_report(report_spec)
+        assert result.content_type == 'table'
```
adaball: suggestion: add a snapshot test for the HIV variant (it appears that STD is the only variant with a snapshot test)

apps/report-execution/src/libraries/pa_02.py
krista-skylight: You know what, I blindly copied SAS export output forgetting that we had this outputs spreadsheet for how we want to format python exports. I'll need to make this modification along with updating column names. Thanks for catching!

apps/report-execution/tests/integration/assets/tablefaker_schema/std_hiv_datamart.yaml
krista-skylight: I gave this a shot but it got rid of all records for qa07, I think because that one's restrictive in terms of looking for duplicates. I may take another stab at modifying the yaml later to appease qa07 more easily with fewer records.

adaball (commented):
- For the HIV variant I can see that for "Dispo G" through "Dispo Z" Category 1 shows "Exam'd".  In the [data mockup](https://docs.google.com/spreadsheets/d/1YpXfOimCcqe-vjrnyz6IB8b6t1W6G8Ajok_X8gac4Cg/edit?gid=1552969357#gid=1552969357) Category 1 is "Not Examined"
- Looking at the "ALL WORKERS" portion of the exported CSV for the HIV variant I can see there are some differences when compared to the PDF output, relevant lines included here:

```csv
Worker,Category 1,Category 2,Part.,Clus.,Reac.,Other,Total
ALL,Exam'd,Exam'd w/in 3:,0,0,0,0,0
ALL,Exam'd,Exam'd w/in 5:,0,0,0,0,0
ALL,Exam'd,Exam'd w/in 7:,0,0,0,0,0
ALL,Exam'd,Exam'd w/in 14:,1,0,0,0,1
ALL,Non-Assigned,,401,0,248,0,649
```
- Looking at the "ALL WORKERS" portion of the exported CSV for the STD variant I can see there are some differences when compared to the PDF output, relevant lines included here:

```csv
Worker,Category 1,Category 2,Part.,Clus.,Reac.,Other,Total
ALL,Non-Assigned,,401,0,248,0,649
```

krista-skylight:
@adaball I pushed some changes that should fix the label issue. The discrepancies for non-assigned should be due to the known bug. I suspect the exam'd w/in discrepancies are also related to the bug but putting this back into draft mode for now while I investigate!

krista-skylight:
@adaball I pushed a conversion note that should explain the discrepancy!

adaball (approved):
:shipit:

timoballard (approved):

PR #3323, opened 2026-06-15
[APP-691]: Add column sort to subset sql and pass on to python report spec

brick-green:
## Description

Adds the ability to sort report queries by a specific column and direction (ASC/DESC). This maps SortSpec to a generic Map<String, Object> in ReportSpec for the JSON payload. Then the builder validates the sort column and appends an ORDER BY clause.

### Key Changes:

**SortSpec Record:** Added a toMap() helper.

**ReportSpec Update:** Integrated an optional sort_by field mapping to the generic map.

**ReportSpecBuilder Logic:**
  - Generates a SQL `ORDER BY [column] ASC/DESC` clause via a new helper method.
  - Replaced the static String.join assembly with a null-safe stream to eliminate double-space formatting anomalies.
  - Added validation to ensure the sort column exists in the requested list, including a defensive guard to prevent NullPointerException if the requested column list itself is null.

## Tickets

* [APP-691](https://cdc-nbs.atlassian.net/jira/software/c/projects/APP/boards/626?selectedIssue=APP-691)

[APP-691]: https://cdc-nbs.atlassian.net/browse/APP-691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -156,4 +169,18 @@ private Integer extractDaysValue() {
...
     }
   }
+
+  private SortSpec validateSortColumns(List<Long> requestedColumnUids, SortSpec sortSpec) {
+
+    if (sortSpec == null) {
+      return null;
+    }
+
+    if (requestedColumnUids == null || !requestedColumnUids.contains(sortSpec.columnUid())) {
```
emyl3, line 190: (q, nb) Should we just return `null` if the `requestedColumnUids == null` instead of throwing an error?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -42,6 +42,14 @@ private String buildSelectClause(List<ReportColumn> columns) {
             .collect(Collectors.joining(", "));
   }
 
+  private String buildOrderByClause(SortSpec sortBy) {
+    if (sortBy == null) {
+      return "";
+    }
+    ReportColumn sortColumn = findMatchingColumn(sortBy.columnUid());
+    return String.format("ORDER BY [%s] %s", sortColumn.name(), sortBy.direction().name());
```
emyl3: (b) I think we need to do this `UPPER` [function call ](https://github.com/cdcent/NEDSSDev/blob/ef89cd6d83915ebe74e2e9900d419dd2175ed0b8/source/gov/cdc/nedss/report/util/ReportSQLBuilder.java#L524-L541) for some of these columns

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportSpec.java
```diff
@@ -11,5 +12,6 @@ public record ReportSpec(
     @JsonProperty(value = "data_source_name", required = true) String dataSourceName,
     @JsonProperty(value = "subset_query", required = true) String subsetQuery,
     @JsonProperty(value = "column_map", required = false) List<List<String>> columnMap,
+    @JsonProperty(value = "sort_by", required = false) Map<String, Object> sortBy,
```
mcmcgrath13:
```suggestion
    @JsonProperty(value = "sort_by", required = false) Map<String, String> sortBy,
```

can we tighten this to a String?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
mcmcgrath13: This is an error state in the current design, so this seems fine to guard agains

apps/report-execution/src/models.py
```diff
@@ -16,6 +16,7 @@ class ReportSpec(BaseModel):
     library_name: str = Field(min_length=1)
     data_source_name: str = Field(min_length=1)
     subset_query: str = Field(min_length=1)
+    sort_by: tuple[str, str] | None = None
```
```suggestion
    sort_by: Json[Any] | None = Field(default_factory=dict)
```
So that it parses nicely into a dict

mcmcgrath13: could you add some test coverage for this on the python side?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SortSpec.java
```diff
@@ -2,7 +2,20 @@
...
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) Long columnUid,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) SortDirection direction) {}
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) SortDirection direction) {
+
+  /** Converts the record into a generic map representation for external API pipelines. */
+  public Map<String, Object> toMap() {
+    // Note: Map.of will throw a NullPointerException if columnUid or direction is null.
+    return Map.of(
+        "column_uid",
+        this.columnUid,
```
mcmcgrath13: (comment, b): the library needs the column title, not the uid - could you update the mappings and such

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
brick-green: done!

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportSpec.java
brick-green: yep yep

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SortSpec.java
brick-green: okay, got it

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -81,18 +95,29 @@ public ReportSpec build() {
...
-
+        java.util.stream.Stream.of(selectClause, fromClause, whereClause, orderByClause)
+            .filter(clause -> clause != null && !clause.isBlank())
+            .collect(Collectors.joining(" "))
+            .trim();
+
+    Map<String, String> sortByMap =
+        (sortBy != null && sortColumn != null)
+            ? Map.of(
+                "column_name", sortColumn.name(),
```
mcmcgrath13:
```suggestion
                "column_name", sortColumn.title(),
```
The title is needed and not the name as this ordering could happen much later when the name is long gone to history

Though really, the clause just needs to be the clause - would it make sense to instead just pass the `ORDER BY [title] dir` without the `ORDER BY` as the UPPER logic is also going to be needed. I'm leaning probably, but it feels less nice

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -42,6 +43,19 @@ private String buildSelectClause(List<ReportColumn> columns) {
...
+    if (sortBy == null || sortColumn == null) {
+      return "";
+    }
+
+    String targetColumn =
+        "STRING".equals(sortColumn.sourceTypeCode())
+            ? String.format("UPPER([%s])", sortColumn.name())
+            : String.format("[%s]", sortColumn.name());
+
+    return String.format("ORDER BY %s %s", targetColumn, sortBy.direction().name());
```
mcmcgrath13: maybe change this to return just the `%s %s` and then add the order by if it's not null for shoving into the overall query, and put the `%s %s` on the report spec and change it to just a string?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -156,4 +182,18 @@ private Integer extractDaysValue() {
               "The '%s' filter value must be a valid integer: %s", description, rawDaysValue));
     }
   }
+
+  private SortSpec validateSortColumns(List<Long> requestedColumnUids, SortSpec sortSpec) {
```
JordanGuinn: (nit, nb): Feels somewhat awkward that a `validateSortColumns` method returns a `SortSpec`?  Could be:

```suggestion
  private void validateSortColumns(List<Long> requestedColumnUids, SortSpec sortSpec) {
```

Just to clarify that if we've moved past this method invocation in `build`, that the sort columns are valid, and not that a `SortSpec` has been created.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -81,18 +95,29 @@ public ReportSpec build() {
     if (columns != null) {
       columnMap = columns.stream().map(c -> List.of(c.name(), c.title())).toList();
     }
+    SortSpec sortBy = validateSortColumns(reportExecRequest.columnUids(), reportExecRequest.sort());
```
JordanGuinn: (nit, nb): Stemming from [my comment here](https://github.com/CDCgov/NEDSS-Modernization/pull/3323/changes#r3424432170) - this reads like `validateSortColumns` is actually generating the `sortBy`.

Thoughts on something like:

```suggestion
    SortSpec sortBy = reportExecRequest.sort();
    validateSortColumns(reportExecRequest.columnUids(), sortBy);
    
```
?

JordanGuinn (commented):
Just a minor nit, but this looks great to me!! 🚀  Pending others' feedback, of course

JordanGuinn (approved):
🚀

apps/modernization-api/src/test/java/gov/cdc/nbs/report/models/ReportSpecTest.java
```diff
@@ -18,7 +20,8 @@ void should_create_report_spec() {
             "nbs_rdb.investigation",
             "SELECT * FROM [NBS_ODSE].[dbo].[NBS_configuration]",
             null,
-            11,
+            "UPPER(Column Title) ASC",
```
mcmcgrath13:
```suggestion
            "UPPER([Column Title]) ASC",
```

apps/modernization-api/src/test/java/gov/cdc/nbs/report/models/ReportSpecTest.java
```diff
@@ -28,7 +31,8 @@ void should_create_report_spec() {
     assertThat(reportSpec.dataSourceName()).isEqualTo("nbs_rdb.investigation");
     assertThat(reportSpec.subsetQuery())
         .isEqualTo("SELECT * FROM [NBS_ODSE].[dbo].[NBS_configuration]");
-    assertThat(reportSpec.daysValue()).isEqualTo(11);
+    assertThat(reportSpec.sortBy()).isEqualTo("UPPER(Column Title) ASC");
```
mcmcgrath13:
```suggestion
    assertThat(reportSpec.sortBy()).isEqualTo("UPPER([Column Title]) ASC");
```

apps/report-execution/tests/integration/execute_report.py
```diff
@@ -27,6 +27,7 @@ def test_execute_report_valid(self):
                 # Filter operator is used here as it is a stable, small table
                 'data_source_name': '[NBS_ODSE].[dbo].[Filter_operator]',
                 'subset_query': 'SELECT * FROM [NBS_ODSE].[dbo].[Filter_operator]',
+                'sort_by': 'UPPER(Column Title) ASC',
```
mcmcgrath13:
```suggestion
                'sort_by': 'UPPER([Column Title]) ASC',
```

mcmcgrath13 (approved):
super minor nits to make the test data more realistic, but overall looks good!

PR #3334, opened 2026-06-16
[APP-666] feat: add column sort to ui

mcmcgrath13:
## Description

Add column sort to column selection card

image

Notes:
* Also added max width styling to the column selection widget, per design
* Moved out the utils to the root of project as they were more useful in more places
* Fixed a twitchy whitespace issue on horizontal field

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-666


apps/modernization-ui/src/apps/report/run/columns/column-selector.module.scss
```diff
@@ -3,67 +3,71 @@
...
-    .search {
-        background-color: colors.$primary-lightest;
-        width: 100%;
-        padding: 1rem;
-        height: fit-content;
-        position: sticky;
-        top: 0;
-        z-index: 2;
-    }
+    .grid{
```
mcmcgrath13: this diff is terrible, but all that changed was adding an inner `.grid` class, moving the background color to the outer div and putting a max width on the grid

apps/modernization-ui/src/design-system/field/HorizontalField.tsx
```diff
@@ -44,10 +43,13 @@ const HorizontalField = ({
         </div>
         <div className={styles.right}>
             <div className={styles.children}>{children}</div>
-            <div className={styles.message}>
-                {warning && <InlineWarningMessage id={`${htmlFor}-warning`}>{warning}</InlineWarningMessage>}
-                {error && <InlineErrorMessage id={`${htmlFor}-error`}>{error}</InlineErrorMessage>}
-            </div>
+            {warning ||
+                (error && (
```
mcmcgrath13, line 48: there was weird white space twitching when the width of the card changed because this takes up some space even when empty, so I changed it to hide when not needed

emyl3:
Looks like there's a random border being added that I'm not seeing in figma... 👀 
Screenshot 2026-06-16 at 17 12 56

emyl3:
Do we want both of these x's?
Screenshot 2026-06-16 at 17 16 50


Also I'm realizing this is just to add the sort... so if this needs to be another ticket to follow up on that makes sense to me 😅

apps/modernization-ui/src/apps/report/run/columns/SortSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return (
+        <EntryWrapper
+            sizing="medium"
+            label="Report data sorting"
+            htmlFor={groupId}
+            orientation="horizontal"
+            helperText="Sort data by a selected column"
+        >
+            <div role="group" id={groupId} className={styles.layout}>
+                <SingleSelect
```
emyl3, line 56: If I select the placeholder "- Select -" it doesn't update.

mcmcgrath13: I'm not able to reproduce this - can we pair on it tomrrow?

mcmcgrath13:
> Do we want both of these x's? Screenshot 2026-06-16 at 17 16 50
> 
> Also I'm realizing this is just to add the sort... so if this needs to be another ticket to follow up on that makes sense to me 😅

I think this may be browser specific behavior - I'm not seeing it on firefox. Which browser are you using?

mcmcgrath13:
> Looks like there's a random border being added that I'm not seeing in figma... 👀 Screenshot 2026-06-16 at 17 12 56

The `Required *` is added as an action, and that's the standard separator between action and collapse-y on the card, so even though it's not in the design, I think we live with it. Thoughts?

apps/modernization-ui/src/apps/report/run/columns/SortSelector.tsx
emyl3: Yes! Sounds good!

emyl3:
> > Do we want both of these x's? Screenshot 2026-06-16 at 17 16 50
> > Also I'm realizing this is just to add the sort... so if this needs to be another ticket to follow up on that makes sense to me 😅
> 
> I think this may be browser specific behavior - I'm not seeing it on firefox. Which browser are you using?

I am using Chrome. Then please ignore!! 😅

emyl3:
> > Looks like there's a random border being added that I'm not seeing in figma... 👀 Screenshot 2026-06-16 at 17 12 56
> 
> The `Required *` is added as an action, and that's the standard separator between action and collapse-y on the card, so even though it's not in the design, I think we live with it. Thoughts?

Ahh! If it is default behavior then I am good with it!!

apps/modernization-ui/src/apps/report/run/columns/SortSelector.tsx
emyl3: my bad... i think the being 10 commits behind was the culprit 😓

emyl3 (approved):
LGTM! Works as expected! :

PR #3348, opened 2026-06-18
build: switch python base image to rhel ubi-minimal

mcmcgrath13:
## Description

Per https://skylight-hq.slack.com/archives/C09FM5R0A1W/p1781787092628219 discussion on using more standardized/minimal/hardened base images where possible


mcmcgrath13:
Given I'm off next week, @timoballard you are now in charge of this PR 🪄

adaball (approved):
RIP Debian 🪦 

:shipit:

apps/report-execution/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.14-slim
+FROM registry.access.redhat.com/ubi10/ubi-minimal:latest
```
brick-green, line 1: q, b: Is `latest` the safe move here? Should it be pinned to a specific version?

mcmcgrath13: it's what we use for the other dockerfiles, so I copied - but agreed something to think about for overall base image approach

PR #3350, opened 2026-06-18
build: update elastic search base image

mcmcgrath13:
## Description

Has fewer vulns per https://hub.docker.com/_/elasticsearch/tags?name=9.


adaball (approved):
:shipit:

PR #3354, opened 2026-06-18
fix: use user profiles and not uesrs for owner drop down

mcmcgrath13:
## Description

Found a bug where it turns out NBS 6 owner is based on `USER_PROFILE` and _not_ `Auth_User`. Add the option endpoint for the profiles and use it instead

### DIFF
--- a/apps/modernization-ui/src/generated/models/RuleGroup.ts
+++ b/apps/modernization-ui/src/generated/models/RuleGroup.ts
@@ -10,8 +10,8 @@ export type RuleGroup = {
 };
 export namespace RuleGroup {
     export enum combinator {
-        OR = 'OR',
-        AND = 'AND',
+        OR = 'or',
+        AND = 'and',
     }
 }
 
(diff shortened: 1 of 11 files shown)


apps/modernization-ui/src/generated/models/RuleGroup.ts
```diff
@@ -10,8 +10,8 @@ export type RuleGroup = {
 };
 export namespace RuleGroup {
     export enum combinator {
-        OR = 'OR',
```
JordanGuinn, line 13: 👀

mcmcgrath13: running `npm run generate` did this? I'll double check it

mcmcgrath13: fixed

emyl3 (approved):
Looks good to me!

PR #3364, opened 2026-06-23
[APP-707] feat: add save report modal

emyl3:
## Description

- Add the Save Report Modal
- Create a report where the ownerUID is the same as the user you are logging in with to test and see the 'Save' button

## Recording
https://github.com/user-attachments/assets/e9e2e9f8-b9fb-4335-9902-f6e7d00cd4ae

## Tickets

- [APP-707](https://cdc-nbs.atlassian.net/browse/APP-707?atlOrigin=eyJpIjoiNDQzNjZmMWMyZDc1NGY1ZmIyYmZlN2QxYTQ0NDEzMjAiLCJwIjoiaiJ9)

[APP-707]: https://cdc-nbs.atlassian.net/browse/APP-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/modernization-ui/src/apps/report/run/ReportResultPage.tsx
+++ b/apps/modernization-ui/src/apps/report/run/ReportResultPage.tsx
@@ -1,9 +1,8 @@
 import { Button } from 'design-system/button';
 import { ReportLayout } from '../layout/ReportLayout';
 import { ReportConfiguration } from 'generated';
-import { AlertBanner } from 'apps/page-builder/components/AlertBanner/AlertBanner';
 import { LoadingIndicator } from 'libs/loading/indicator';
-import { ReactNode } from 'react';
+import React, { ReactNode, useRef } from 'react';
 import { Heading } from 'components/heading';
 import { permissions, permitsAny, Permitted } from 'libs/permission';
 import { Shown } from 'conditional-render';
@@ -11,23 +10,51 @@ import { useUser } from 'user';
 
 import layoutStyles from '../layout/layout.module.scss';
 import { PERMISSION_GROUP_MAP } from '../constants';
+import { ModalRef } from '@trussworks/react-uswds';
+import { SaveReportModal } from './modals/SaveReportModal.tsx';
+import { AlertMessage } from '../../../design-system/message';
 
 const ReportResultPage = ({
     config,
     error,
     wasExported,
     resultLoading,
+    resultSaving,
+    isRedirecting,
     handleRefineReport,
+    handleSaveReport,
 }: {
     config: ReportConfiguration;
     error: string | null;
     wasExported: boolean;
     resultLoading: boolean;
+    resultSaving: boolean;
+    isRedirecting: boolean;
     handleRefineReport: () => void;
+    handleSaveReport: () => void;
 }) => {
     const {
         state: { user },
     } = useUser();
+    const saveReportModalRef = useRef<ModalRef>(null);
+
+    const getLoadingHeader = () => {
+        if (resultLoading) {
+            return `Your report is ${wasExported ? 'downloading' : 'opening in a new tab'}. Please do not leave
+        this page while your report is generating.`;
+        } else {
+            return 'Your report is saving and you will be redirected to the Manage Reports page once complete.';
+        }
+    };
+
+    const getLoadingBody = () => {
+        if (resultLoading) {
+            return (
+                'This might take several minutes for large reports. To be sure it opens, check that pop-ups ' +
+                ' are enabled in your browser.'
+            );
+        }
+    };
 
     return (
         <ReportLayout
@@ -35,7 +62,7 @@ const ReportResultPage = ({
             actions={
                 <>
                     <Permitted permission={PERMISSION_GROUP_MAP[config.group].selectFilterCriteria}>
-                        <Button onClick={handleRefineReport} secondary={true} disabled={resultLoading}>
+                        <Button onClick={handleRefineReport} secondary={true} disabled={resultLoading || resultSaving}>
                             Refine Report
                         </Button>
                     </Permitted>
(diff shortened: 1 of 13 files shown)


apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
```diff
@@ -150,6 +156,33 @@ const ReportRunPage = () => {
...
+            dataSourceUid: lastReportExecutionRequest.dataSourceUid,
+            requestBody: lastReportExecutionRequest,
+        })
+            .then(() => {
+                setStatus('redirecting');
+                redirectToManageReport();
+            })
+            .catch((err) => {
+                setError(JSON.stringify(err));
+            });
```
brick-green, line 181: q, maybe b: what does the shape of this `err` look like? Is there a message or body or something to be used?

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
```diff
@@ -150,6 +156,33 @@ const ReportRunPage = () => {
         [config]
     );
 
+    const redirectToManageReport = () => {
+        window.location.href = '/nbs/nfc?ObjectType=7&OperationType=116';
```
brick-green: Gemini recommended something like this:
```suggestion
        const navigate = useNavigate();
        navigate('/nbs/nfc?ObjectType=7&OperationType=116');
```

as I understand it `window.location.href ` forces a full page reload which perhaps isn't always necessary. A little out of my depths here though. Thoughts?

brick-green (commented):
Couple little things

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
JordanGuinn: +1 here, I think we lose all application state on a forced page reload like this - would def prefer navigation through the React router if we're able!

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
JordanGuinn: I don't know if it ever made it's way into the UI given it's handled API-side through exception handlers, but 4xx/5xx errors from all reports endpoints use the `ErrorResponseBody` interface for their response bodies, which look like:

```
{
   message: String;
}
```

If we want to use that here!

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
emyl3: I don't think we are able to since that part of the app hasn't been modernized 😢 Which is why I believe we are doing [this similarly here](https://github.com/CDCgov/NEDSS-Modernization/blob/c512cae169c2bda4d71d4da134b85640778f13b4/apps/modernization-ui/src/apps/report/management/configuration/ViewReportConfiguration.tsx#L57)

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
emyl3: OOO Thank you!! I'll update!

testing/regression/cypress/e2e/pages/login.page.js
```diff
@@ -1,58 +1,63 @@
...
-     });
-  }
-
-  verifyLoginPage() {
-    cy.get(".logo img").should("be.visible");
-  }
-}
-
-export const loginPage = new LoginPage();
+const user = Cypress.env("LOGIN_USERNAME");
```
emyl3, line 1: ⚠️ had to get rid of some weird CRLF line endings in this file so it looks like the entire file changed

emyl3:
@brick-green @JordanGuinn @timoballard Ready for re-review! 
edit: oops had to remove a debug line.... but it should be actually good now 😅

brick-green (approved):
noice 🥇

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
JordanGuinn: Do you think it'd be worth tracking and making a tech debt ticket for?  If it's just in these two places, maybe not, but if it expands beyond this, probably worthwhile to at least have proper routing as a work on the board to prioritize at _some_ point down the road

JordanGuinn (approved):
🚀

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
mcmcgrath13: Once the report listing page is modernized, this can be switched to `useNavigate`, but until then, we need to use the full refresh or things do not work

apps/modernization-ui/src/apps/report/run/ReportResultPage.tsx
```diff
@@ -1,41 +1,68 @@
...
+    };
+
+    const getLoadingBody = () => {
+        if (resultLoading) {
+            return (
+                'This might take several minutes for large reports. To be sure it opens, check that pop-ups ' +
+                ' are enabled in your browser.'
+            );
+        }
+    };
```
mcmcgrath13, line 57: (nit, nb-clearly 😂 ): There's no need to do these as functions, it is more react to just create the messages (typically inline if the logic is straightforward enough) or just do this with variable assignment (e..g `const loadingBody = ...` instead of functions)

apps/modernization-ui/src/apps/report/run/ReportResultPage.tsx
```diff
@@ -46,41 +73,57 @@ const ReportResultPage = ({
...
-                        enabled in your browser.
-                    </p>
-                </TextCard>
-            ) : (
-                !error && (
-                    <TextCard>
-                        <Heading level={2}>
-                            Your report has {wasExported ? 'downloaded' : 'opened in a new tab'}.
-                        </Heading>
+            <>
```
mcmcgrath13, line 94: (nit, nb-clearly 😂 ): unnecessary `<>` same on line 97

apps/modernization-ui/src/apps/report/run/ReportResultPage.tsx
```diff
@@ -46,41 +73,57 @@ const ReportResultPage = ({
...
+                {error && (
+                    <div className={'padding-2'}>
```
mcmcgrath13, line 96: (nit, nb-clearly 😂 ): use `prop="string"` when no logic is done (really, we should figure out how to get eslint to do this for us)

apps/modernization-ui/src/apps/report/run/ReportResultPage.tsx
```diff
@@ -1,41 +1,68 @@
...
         state: { user },
     } = useUser();
+    const saveReportModalRef = useRef<ModalRef>(null);
+
+    const getLoadingHeader = () => {
+        if (resultLoading) {
+            return `Your report is ${wasExported ? 'downloading' : 'opening in a new tab'}. Please do not leave
+        this page while your report is generating.`;
+        } else {
+            return 'Your report is saving and you will be redirected to the Manage Reports page once complete.';
```
mcmcgrath13, line 46: (nit, maybe-b-slash-follow-on): in the admin flow, we disable the buttons while the save is happening, but leave the modal up. We should probably do that here as well and then we can remove this report saving state - thoughts? Also, page redirects are really fast, why show that as a state at all vs just doing the thing?

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
```diff
@@ -166,9 +197,12 @@ const ReportRunPage = () => {
         <ReportResultPage
             config={config}
             resultLoading={status === 'submitting'}
+            resultSaving={status === 'saving'}
+            isRedirecting={status === 'redirecting'}
             wasExported={wasExported}
             error={error}
             handleRefineReport={() => setStatus('configuring')}
+            handleSaveReport={handleSaveReport}
```
mcmcgrath13, line 205: can we do this locally on the report result page instead of at the top level? these statuses are unrelated to the overall flow and are local to that page - this page could be untouched

PR #3365, opened 2026-06-23
APP-769: filter unmapped race codes from demographics summary card

nullflux:
## Summary
The patient demographics **Summary** card showed raw `race_category_cd` values (e.g. "UNK", "1") for race codes with no `NBS_SRTE..Code_value_general` RACE_CALCULATED match, while the **Demographics** card correctly showed only matched values. Reported by Montana (SEER #18435).

Root cause: the two cards use different finders.
- `PatientDemographicsSummaryRaceFinder` LEFT JOINed the SRTE code table and fell back to the raw code via `coalesce(..., race_category_cd, '')`, so an unmatched code rendered raw.
- `PatientRaceDemographicFinder` (Demographics) INNER JOINs and drops unmatched codes.

## Change
Switch the Summary race finder to an INNER JOIN, matching the Demographics finder, so unmapped category codes are filtered out instead of shown raw. Drops the now-dead `coalesce` fallbacks.

## Testing
Added a `demographics-summary` Cucumber scenario with a race category code that has no SRTE match. Verified red→green against the `nedssdb` Testcontainer: before the fix `$.races` was `["Asian","ZTEST"]`; after, `["Asian"]`. Full `@patient-demographics-summary` and `@patient-race-demographics` suites pass with no regression.

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/patient/file/demographics/summary/PatientDemographicsSummaryRaceFinder.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/patient/file/demographics/summary/PatientDemographicsSummaryRaceFinder.java
@@ -10,10 +10,10 @@ class PatientDemographicsSummaryRaceFinder {
   private static final String QUERY =
       """
       select
-          coalesce([race].code_short_desc_txt, [patient].race_category_cd, '')
+          [race].code_short_desc_txt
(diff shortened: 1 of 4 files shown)


JordanGuinn (approved):
Thank you! 🙏

emyl3 (approved):
LGTM! Thank you for addressing this and adding tests! 🫡

PR #3366, opened 2026-06-24
APP-770: Filter inactive counties from county options dropdown

nullflux:
## Summary

The NBS 7 patient address **County** dropdown (Add Patient → Address, plus its autocomplete) listed outdated and duplicate counties. Reported by Alaska via NBS Central SEER #18441; expected behavior is active counties only.

## Root cause

The county option queries read `NBS_SRTE..state_county_code_value` with no active-status predicate:

- `CountiesListFinder` (`GET /nbs/api/options/counties/{state}`, the dropdown)
- `CountyOptionResolver` (county autocomplete)

Every other SRTE-backed option finder in `options-api` (jurisdiction, condition, primary language, occupation, program area) filters `status_cd = 'A'`. These two county finders were the only exceptions, so retired/renamed county codes and their old-vs-new duplicates rendered

## Change

Add `status_cd = 'A'` to both county finders, matching the sibling finders.

## Test plan

Red-fix-green on the `options-api` `@EmbeddedNbsDatabase` Cucumber harness (real SQL Server Testcontainer):

- Two new `CountiesOptions.feature` scenarios assert an inactive county is excluded from the list and from autocomplete.
- `CountyMother` now stamps `status_cd` and can seed inactive counties.
- RED (before the finder change): 2 failures. GREEN (after): 76/76 pass.

## Follow-up

Not yet confirmed against live AK SRTE data whether the outdated counties are flagged via `status_cd` or via `effective_to_time` (with `status_cd` still `A`). If the latter, both finders also need `(effective_to_time is null or effective_to_time > getdate())`. Repro SQL is on SEER #18441 for the state to confirm.


brick-green (approved):
no notes 🥇

JordanGuinn (approved):
Much appreciated 👏 🚀

PR #3367, opened 2026-06-24
[APP-664] Use NBS_Configuration table to store dynamic configuration values used in report execution

brick-green:
This PR migrates database alias names and row execution limits (REPORT_MAX_ROW_LIMIT_RUN and REPORT_MAX_ROW_LIMIT_EXPORT)—out of environment vars (.env files / Spring specifications) and handles them via the NBS_configuration table.

Key Changes
  - Configuration Finder (): Implemented database config retrieval utility mirroring the pattern on the Java and python  sides.
  - On python side I adjusted the container creation so that the NBS_configuration table can be populated as needed on test libraries

## Tickets

* [APP-664](https://cdc-nbs.atlassian.net/jira/software/c/projects/APP/boards/626?selectedIssue=APP-664)

[APP-664]: https://cdc-nbs.atlassian.net/browse/APP-664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emyl3 (dismissed):
Looks like there's still references to a hardcoded RDB in the [PA_03 lib](https://github.com/CDCgov/NEDSS-Modernization/blob/e661be09fc7d995bba9f094d49b718d35acf1284/apps/report-execution/src/libraries/pa_03.py#L140)

brick-green:
> Looks like there's still references to a hardcoded RDB in the [PA_03 lib](https://github.com/CDCgov/NEDSS-Modernization/blob/e661be09fc7d995bba9f094d49b718d35acf1284/apps/report-execution/src/libraries/pa_03.py#L140)

ah, yup good eyes. That's the new library Ada is working on. Missed in in the merge from main

brick-green:
Companion PR in NEDSSDB: https://github.com/cdcent/NEDSSDB/pull/50

apps/report-execution/src/config.py
```diff
@@ -0,0 +1,36 @@
...
+from . import errors
+
+
+def get_config_value(trx: Transaction, config_key: str) -> str:
+    """Retrieves a configuration value from NBS_ODSE..NBS_configuration."""
+    query = """
+            SELECT COALESCE(config_value, default_value)
+            FROM NBS_ODSE..NBS_configuration
+            WHERE config_key = ? \
+            """
```
adaball, line 28: q: Are we expecting the configuration info to be present in production?  I don't see it in the golden backup, nor do I see migrations to populate it.

apps/report-execution/src/execute_report.py
```diff
@@ -28,7 +29,7 @@ def execute_report(report_spec: models.ReportSpec):
             library_params=report_spec.library_params,
         )
 
-    check_valid_result(result, report_spec)
+        check_valid_result(result, report_spec, trx)
```
adaball: nit: The indent is off here.

apps/report-execution/src/config.py
adaball: I'm just now seeing the companion PR.

apps/report-execution/src/libraries/nbs_sr_13.py
```diff
@@ -18,13 +19,16 @@ def execute(
     * Changed file name from NBSSR000017.sas to nbr_sr_13.py since 17 was
       associated with label for 13
     """
+    # Dynamically look up the correct name for the SRTE database
+    nbs_srte = get_config_value(trx, 'nbs_srte')
```
mcmcgrath13:
```suggestion
    nbs_srt = get_config_value(trx, 'nbs_srt')
```

apps/report-execution/src/config.py
```diff
@@ -0,0 +1,36 @@
+from src.db_transaction import Transaction
+
+from . import errors
+
+
+def get_config_value(trx: Transaction, config_key: str) -> str:
```
mcmcgrath13, line 14: (sugg, nb): is it worth caching these values so we don't need to do this read on every report that uses one? they should be pretty stable

apps/report-execution/tests/integration/libraries/custom_library.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 
-@pytest.mark.usefixtures('setup_containers')
+@pytest.mark.usefixtures('setup_containers_with_seed_data')
```
mcmcgrath13: (q, b): any reason to not just have the setup containers add seed data always?

apps/report-execution/tests/conftest.py
```diff
@@ -142,6 +150,45 @@ def teardown():
...
+    Runs exactly ONCE per test session using the active connection string boundaries.
+    """
+    conn_string = utils.get_env_or_error('DATABASE_CONN_STRING')
+
+    with db_transaction(conn_string) as trx:
+        # Baseline mappings required across the testing landscape
+        baseline_configs = [
+            ('NBS_RDB', 'RDB', 'RDB'),
+            ('NBS_ODS', 'NBS_ODSE', 'NBS_ODSE'),
+            ('NBS_SRTE', 'NBS_SRTE', 'NBS_SRTE'),
```
mcmcgrath13:
```suggestion
            ('NBS_SRT', 'NBS_SRTE', 'NBS_SRTE'),
```

and `NBS_MSG` is missing (though maybe not relevant for libs at this point?)

apps/report-execution/tests/conftest.py
```diff
@@ -142,6 +150,45 @@ def teardown():
...
+        integration tests.
+
+    Runs exactly ONCE per test session using the active connection string boundaries.
+    """
+    conn_string = utils.get_env_or_error('DATABASE_CONN_STRING')
+
+    with db_transaction(conn_string) as trx:
+        # Baseline mappings required across the testing landscape
+        baseline_configs = [
+            ('NBS_RDB', 'RDB', 'RDB'),
```
mcmcgrath13: (q, nb) if we have a default value, why insert a normal value?

apps/report-execution/tests/utils.py
```diff
@@ -19,42 +19,6 @@ def test_get_env_or_error_missing(self):
         assert exc_info.value.http_code == 500
         assert exc_info.value.message == 'Internal Server Error'
 
-    def test_get_int_env_or_default_valid(self, monkeypatch):
```
mcmcgrath13, line 22: (q, nb): why delete these tests? it looks like the code is still there - we should either keep the tests or remove the util if it's now dead

apps/report-execution/sample.env
```diff
@@ -8,8 +8,6 @@ DATABASE_CONN_STRING="Server=${NBS_DATASOURCE_SERVER};UID=${DATABASE_USER};PWD=$
 
 # Configuration parameters with reasonable defaults set in the code
 # note: `REPORT_EXPORT_*` values must be valid strftime strings (https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)
-# REPORT_MAX_ROW_LIMIT_RUN=10000
-# REPORT_MAX_ROW_LIMIT_EXPORT=100000
 # REPORT_EXPORT_DATE_FORMAT="%m/%d/%Y"
 # REPORT_EXPORT_DATETIME_FORMAT="%m/%d/%Y %H:%M:%S"
```
mcmcgrath13: (sugg, b): these two also need to be config-ified

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
```diff
@@ -0,0 +1,53 @@
...
+  private static final System.Logger LOGGER =
+      System.getLogger(ConfigurationValueFinder.class.getName());
+
+  private final JdbcClient client;
+
+  private static final String CONFIG_QUERY =
+      """
+          SELECT COALESCE(config_value, default_value)
+          FROM NBS_ODSE..NBS_configuration
+          WHERE config_key = ?
```
mcmcgrath13: (q,b): is this case insensitive? should we upper/lower both values to ensure it will always match?

apps/report-execution/src/execute_report.py
brick-green: this indent is intentional. It gives access to the `trx` so that it can be passed in to `check_valid_result`

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
```diff
@@ -26,16 +28,45 @@ public void initialize(final ConfigurableApplicationContext context) {
...
+
+      var command =
+          """
+          export PATH="$PATH:/opt/mssql/bin:/opt/mssql-tools18/bin" && \
+          export SQLCMDPASSWORD='%s' && \
+          export MIGRATIONS_DIR='migrations' && \
+          export SEED_DATA_DIR='seed_data' && \
+          /var/data/run_migrations.sh %s
+          """
+              .formatted(credential, "6.0.19.1");
```
mcmcgrath13: (q,b): the migrations should automatically run to the version set in the env on container start - why do we need this as a separate step?

mcmcgrath13 (commented):
overall question/thought, should we add a common prefix to db name lookups so they group? right now, they could end up scattered all over the config, which seems less than ideal

maybe `REPORT_DB_{key}`?

mcmcgrath13:
(sugg, b): If we require these configs to run reporting, we should add a migration to the liquibase setup for them (then you can revert the nedssdb changes)

apps/report-execution/src/config.py
brick-green: yeah I like this. It wouldn't matter too much for the users exporting via the front-end. But if people connect a dashboard and poll the api this could be great.

apps/report-execution/tests/integration/libraries/custom_library.py
brick-green: No big reason. Just didn't want to add steps that weren't needed for the specific tests. Should I just add it across the board?

apps/report-execution/tests/conftest.py
brick-green: Yeah MSGOUT isn't used in the libs yet

apps/report-execution/tests/conftest.py
brick-green: I did this in the DB seed data so I did here too. Though I should probably mix up the testing config so that the COALESCE is tested

apps/report-execution/sample.env
brick-green: oh right, I forgot about these ones

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
brick-green: SQLServer is case insensitive. But maybe future proof for other DBs?

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
brick-green: I wasn't seeing the seed data in test container db as it was. It didn't look like the migrations were running...how would they run from the container.start()?

apps/report-execution/src/execute_report.py
adaball: Oh whoops, my bad.

apps/report-execution/tests/integration/libraries/custom_library.py
mcmcgrath13: we're generally using/re-using the same containers over and over, so I'd add it across the board for consistency (and smaller diff)

apps/report-execution/tests/conftest.py
mcmcgrath13: for the db seeding, should we include a default? or lean on the default value column?

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
mcmcgrath13: it's built into the container's entrypoint run that it does migrations, does this elide that?

brick-green: oh nice TIL about entrypoint. So that works but the env var was set wrong. Changed `DB_VERSION` to `DATABASE_VERSION` and voila

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
```diff
@@ -0,0 +1,45 @@
...
+SELECT
+    src.config_key, src.config_value, src.default_value,
+    src.version_ctrl_nbr, src.add_user_id, src.add_time,
+    src.last_chg_user_id, src.last_chg_time, src.status_cd, src.status_time
+FROM (
+    VALUES
+    ('REPORT_DB_NBS_RDB', 'RDB', 'RDB', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE()),
+    ('REPORT_DB_NBS_ODS', 'NBS_ODSE', 'NBS_ODSE', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE()),
+    ('REPORT_DB_NBS_SRT', 'NBS_SRTE', 'NBS_SRTE', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE()),
+    ('REPORT_DB_NBS_MSGOUT', 'NBS_MSGOUTE', 'NBS_MSGOUTE', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE()),
```
mcmcgrath13:
```suggestion
    ('REPORT_DB_NBS_MSG', 'NBS_MSGOUTE', 'NBS_MSGOUTE', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE()),
```

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
```diff
@@ -0,0 +1,235 @@
...
+    src.status_time,
+    src.admin_comment,
+    src.system_usage
+FROM (
+     VALUES
+     (
+     'REPORT_DB_NBS_RDB',
+     'RDB',
+     'RDB Database Name',
+     'Database name or catalog alias for the relational tracking DB.',
```
mcmcgrath13:
```suggestion
     'Database name or catalog alias for the reporting DB. Used for report execution.',
```

And add something about using this for report execution across all these

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
```diff
@@ -0,0 +1,235 @@
...
+     'A',
+     GETDATE(),
+     'Governs safety boundaries against interactive runtime overflows.',
+     NULL
+     ),
+     (
+     'REPORT_EXPORT_DATE_FORMAT',
+     '%m/%d/%Y',
+     'Report Export Date Format',
+     'Python strftime template used when sanitizing Date elements.',
```
mcmcgrath13:
```suggestion
     'Python strftime template used when serializing Date elements.',
```

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
```diff
@@ -0,0 +1,235 @@
...
+     'A',
+     GETDATE(),
+     'Defines localized formatting target for date fields.',
+     NULL
+     ),
+     (
+     'REPORT_EXPORT_DATETIME_FORMAT',
+     '%m/%d/%Y %H:%M:%S',
+     'Report Export Datetime Format',
+     'Python strftime template used when sanitizing Datetime elements.',
```
mcmcgrath13:
```suggestion
     'Python strftime template used when serializing Datetime elements.',
```

apps/report-execution/src/config.py
```diff
@@ -0,0 +1,95 @@
...
+# This block is evaluated as True ONLY by IDEs and type checkers,
+# completely skipping execution at application runtime.
+if TYPE_CHECKING:
+    from src.db_transaction import Transaction
+
+# Module-level cache dictionary
+_CONFIG_CACHE: dict[str, str] = {}
+
+
+# FIX: Use quotes around "Transaction" to make it a string forward reference
```
mcmcgrath13: (comment, nb): not following this comment

apps/report-execution/src/config.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+    # Populate the cache with the normalized key
+    resolved_value = str(val).strip()
+    _CONFIG_CACHE[normalized_key] = resolved_value
+
+    return resolved_value
+
+
+# FIX: Use quotes around "Transaction" here as well
```
mcmcgrath13: ditto

apps/report-execution/src/config.py
```diff
@@ -0,0 +1,95 @@
...
+def prime_report_configurations(trx: Transaction) -> None:
```
mcmcgrath13:
```suggestion
def load_report_configurations(trx: Transaction) -> None:
```
maybe? `prime` reads weird to my eye

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
```diff
@@ -1,41 +1,103 @@
...
-        new NbsDatabaseContainer<>(image)
-            .withUsername(username)
-            .withPassword(credential)
-            .withEnv("DB_VERSION", "6.0.19.1")
-            .withImagePullPolicy(PullPolicy.alwaysPull());
+      databaseContainer.start();
 
-    container.start();
+      // Seed the required configuration entries exactly once
+      seedBaselineConfiguration(databaseContainer);
```
mcmcgrath13: (q, nb): does liquibase not handle this for us now that there's a migration all set up?

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.jdbc.core.simple.JdbcClient;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ConfigurationValueFinder {
+
+  private static final System.Logger LOGGER =
+      System.getLogger(ConfigurationValueFinder.class.getName());
+
+  private final JdbcClient client;
```
mcmcgrath13, line 17: (sugg, maybe-b): can we add a cache here as well?

brick-green: yup, added

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
brick-green: yea! I reverted the file and will log a bug ticket

apps/report-execution/src/config.py
brick-green: This was a left over comment from a strategy I abandoned. Deleted both comments

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
brick-green: actually I was looking at the incorrect database. Liquibase was not running but I enabled it

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -82,7 +82,7 @@ public ReportService(
     this.reportMapper = reportMapper;
 
     this.reportExecutionClient = reportExecutionClient;
-    this.dataSourceNameUtils = new DataSourceNameUtils(dataSourceNameConfig);
+    this.dataSourceNameUtils = new DataSourceNameUtils(configurationValueFinder);
```
mcmcgrath13: (thought, very-nb): now that the datasourcename utils use a db thing which is no longer trivial to just pass mock data into, would it make more sense to make that a `@Component` and have the config value finder injected here instead 🤔

apps/report-execution/src/execute_report.py
```diff
@@ -43,7 +46,9 @@ def is_valid_library(library):
     return True
 
 
-def check_valid_result(report_result: typing.Any, report_spec: models.ReportSpec):
+def check_valid_result(
+    report_result: typing.Any, report_spec: models.ReportSpec, trx: Transaction
```
mcmcgrath13: (nit, merge-blocking) - `trx` isn't used any more, so that should be removed from the signature and then the check above can be de-indented

apps/report-execution/sample.env
```diff
@@ -8,8 +8,4 @@ DATABASE_CONN_STRING="Server=${NBS_DATASOURCE_SERVER};UID=${DATABASE_USER};PWD=$
 
 # Configuration parameters with reasonable defaults set in the code
 # note: `REPORT_EXPORT_*` values must be valid strftime strings (https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes)
```
mcmcgrath13:
```suggestion
```

no longer makes sense

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private final Map<String, String> cache = new ConcurrentHashMap<>();
+
+  private static final String CONFIG_QUERY =
+      """
+              SELECT config_value
+              FROM NBS_ODSE..NBS_configuration
+              WHERE config_key = ?
+              """;
```
mcmcgrath13: (nit, nb): this indent looks weird on github, google's doing? or something that can be de-indented?

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/config/CacheLoader.java
```diff
@@ -0,0 +1,25 @@
...
+    this.finder = finder;
+  }
+
+  /**
+   * Fires automatically the exact second the Spring Boot context becomes fully active and ready to
+   * handle incoming web or API requests.
+   */
+  @EventListener(ApplicationReadyEvent.class)
+  public void onApplicationStartup() {
+    this.finder.loadConfigurations();
```
mcmcgrath13, line 23: this is after liquibase runs, correct?

libs/testing/database/src/main/java/gov/cdc/nbs/testing/database/NbsTestDatabaseInitializer.java
```diff
@@ -36,6 +36,11 @@ public void initialize(final ConfigurableApplicationContext context) {
     System.getLogger(NbsTestDatabaseInitializer.class.getCanonicalName())
         .log(System.Logger.Level.INFO, () -> "[url]: %s".formatted(url));
 
-    TestPropertyValues.of("spring.datasource.url=" + url).applyTo(context.getEnvironment());
+    TestPropertyValues.of(
+            "spring.datasource.url=" + url,
+            "spring.liquibase.enabled=true",
+            "spring.liquibase.driver-class-name=com.microsoft.sqlserver.jdbc.SQLServerDriver",
+            "nbs.ui.features.report.execution.enabled=true")
```
mcmcgrath13: (nit, probably-b): these can/should be set in the `application-test.yml` - change there instead?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
brick-green: good call! Done

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
```diff
@@ -0,0 +1,231 @@
...
+SCOPE:
+  - REPORT_DB_* : Database catalog aliases used for routing reports dynamically.
+  - REPORT_MAX_ROW_LIMIT_* : Maximum row thresholds enforced to prevent out-of-memory errors
+                                during run or export lifecycles.
+
+Insert If Not Exists. This script compares target records against existing database keys. If a
+configuration key is already present, it skips insertion entirely.
+====================================================================================================
+*/
+
```
adaball, line 16: issue: You'll need to add a `USE` statement, otherwise this query won't have any DB to reference.

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
```diff
@@ -0,0 +1,231 @@
...
+     GETDATE(),
+     99999999,
+     GETDATE(),
+     'A',
+     GETDATE(),
+     'Governs safety boundaries against massive tabular data exports.',
+     NULL
+     ),
+     (
+     'REPORT_MAX_ROW_LIMIT_RUN',
```
adaball, line 154: (q, nit): Should this be a `MERGE` or `UPDATE` instead of an `INSERT`?

I can see that `REPORT_MAX_ROW_LIMIT_RUN` and `REPORT_MAX_ROW_LIMIT_EXPORT` already exist in the DB due to the updates that you made in `NEDSSDB` seed data:

```
~/repos/NEDSSDB $ grep -rn REPORT_MAX_ROW ./*
./src/seed_data/09_database_name_mapping.sql:14:    ('REPORT_MAX_ROW_LIMIT_EXPORT', '100000', '100000', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE()),
./src/seed_data/09_database_name_mapping.sql:15:    ('REPORT_MAX_ROW_LIMIT_RUN', '10000', '10000', 1, 99999999, GETDATE(), 99999999, GETDATE(), 'A', GETDATE())
```

so technically speaking these 2 items will not be affected by this query.

If the assumption is that none of these items will exist in prod, then this query is fine, but if not we'd benefit from using `MERGE` or `UPDATE` to make sure the data is fresh.

libs/datasource-utils/src/test/java/gov/cdc/nbs/datasource/utils/DataSourceNameUtilsTest.java
```diff
@@ -17,16 +17,26 @@
...
 class DataSourceNameUtilsTest {
-  @Mock DataSourceNameConfiguration config;
+
+  private static final Map<String, String> DATABASE_MAPPINGS =
+      Map.of(
+          "report_db_nbs_ods", "NBS_ODSE",
+          "report_db_nbs_odse", "NBS_ODSE",
+          "report_db_odse", "NBS_ODSE",
+          "report_db_ods", "NBS_ODSE");
+
```
adaball, line 27: issue: Some of the keys defined here are missing from the liquibase migration.

The only entry from this map that actually exists in the liquibase migration is `REPORT_DB_NBS_ODS`, so technically this test is exercising on non-existing data (unless they somehow already exist in prod DBs).

I may be missing something here but I just want to make sure we're exercising on data that shouldn't be accepted.

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/config/CacheLoader.java
brick-green: as I understand it, yes. Spring will initialize all the Beans before the `ApplicationReadyEvent`. So `reportExecutionLiquibase` will run before that

apps/report-execution/src/config.py
```diff
@@ -0,0 +1,93 @@
...
+    subsequent lookups. Automatically handles case-insensitivity.
+    """
+    normalized_key = config_key.upper()
+
+    # check the cache
+    if normalized_key in _CONFIG_CACHE:
+        return _CONFIG_CACHE[normalized_key]
+
+    query = """
+            SELECT config_value
```
adaball, line 25: suggestion: Use `default_value` when `config_value` is `NULL` or empty.

Something like:

```sql
SELECT COALESCE(NULLIF(config_value, ''), default_value)
```

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
```diff
@@ -0,0 +1,119 @@
...
+  private static final System.Logger LOGGER =
+      System.getLogger(ConfigurationValueFinder.class.getName());
+
+  private final JdbcClient client;
+
+  private final Map<String, String> cache = new ConcurrentHashMap<>();
+
+  private static final String CONFIG_QUERY =
+      """
+      SELECT config_value
```
adaball, line 23: suggestion: Use `default_value` when `config_value` is `NULL` or empty.

Something like:

```sql
SELECT COALESCE(NULLIF(config_value, ''), default_value)
```

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
```diff
@@ -0,0 +1,119 @@
...
+      LOGGER.log(
+          System.Logger.Level.WARNING,
+          "Configuration key ({0}) exists in NBS_Configuration, but both config_value and default_value are null",
+          key);
+      return "";
+    }
+  }
+
+  /**
+   * Batch-loads and caches all keys required by the application runtime. Throws an exception if any
```
mcmcgrath13:
```suggestion
   * Batch-loads and caches all keys known to be used by the application runtime. Throws an exception if any
```

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
```diff
@@ -0,0 +1,119 @@
...
+
+    String[] requiredKeys = {
+      "REPORT_DB_NBS_RDB",
+      "REPORT_DB_NBS_ODS",
+      "REPORT_DB_NBS_SRT",
+      "REPORT_DB_NBS_MSG",
+      "REPORT_MAX_ROW_LIMIT_RUN",
+      "REPORT_MAX_ROW_LIMIT_EXPORT",
+      "REPORT_EXPORT_DATE_FORMAT",
+      "REPORT_EXPORT_DATETIME_FORMAT"
```
mcmcgrath13: (q, nb): these aren't used by mod api - should we remove them here? or is it still a generally valuable failure point to say "hey! your config isn't set up right for reporting" in one place vs splitting that between here and the report execution service?

mcmcgrath13 (approved):
a couple minor nits/questions, but overall looks good!

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
brick-green: It's working without the `USE` statement but I didn't understand why. Turns out liquibase targets NBS_ODSE specifically specifically in the connection url. I think it's a good practice to add the USE in though!

Here's the line from `application.yml` that LiquibaseConfig targets

```
  datasource:
    url: jdbc:sqlserver://${nbs.datasource.server}:${nbs.datasource.port};database=nbs_odse;encrypt=true;trustServerCertificate=true;
````

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
brick-green: I reverted that PR in the DB repo in favor of just this liquibase script. I don't expect the keys to be there in prod. If they are there I don't think we should overwrite those values though. I think that would mean the STLT user set those up already with their intended values.

libs/datasource-utils/src/main/java/gov/cdc/nbs/datasource/utils/ConfigurationValueFinder.java
brick-green: I think I prefer removing them to keep the responsibilities separate

apps/report-execution/src/config.py
brick-green: Looked into the 'default_value' column and it looks like it's mostly used in the rest of the app as documentation for what the `config_value` should look like. Talked to Mary about it too and we ended up leaning to just using the config_value and failing if it's not set

apps/modernization-api/src/main/resources/db/report/execution/04_ODSE_Data_Report_Config_Insert.sql
adaball: Oh interesting!

adaball (approved):
:shipit:

PR #3368, opened 2026-06-24
APP-667: Save Report endpoint

JordanGuinn:
## Description

This PR adds an endpoint for overwriting (or "saving") existing reports, specifically the report's filter and sort data.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-667)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportSpecBuilder.java
```diff
@@ -13,7 +13,7 @@
  * Execution service for subsequent report generation. It assembles report metadata, includes
  * togglable report parameters and (arguably most importantly) compiles a SQL statement to be
  * invoked during report execution, accounting for all supplied parameters as well as relevant
- * permissions for the individual invoking said report.
+ * permissions for the individual invoking that report.
```
JordanGuinn, line 16: IntelliJ was complaining _loudly_ about this for some reason??

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,134 @@
...
+import gov.cdc.nbs.report.models.BasicFilterRequest;
+import java.util.ArrayList;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
+public class FilterValueMapper {
+  private final IdGeneratorService idGenerator;
+
+  private int sequenceNumber = 1;
```
mcmcgrath13: (q, maybe-b): my vague recollection from some other PR is components are singletons, so this sequence number would be for every mapper and will maybe just keep incrementing forever - is that true? or does this reset per `fromAdvancedFilterRequest` invocation?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,134 @@
...
+
+  public FilterValueMapper(IdGeneratorService idGenerator) {
+    this.idGenerator = idGenerator;
+  }
+
+  public List<FilterValue> fromBasicFilterRequest(
+      ReportFilter basicFilter, BasicFilterRequest request) {
+    return request.values().stream()
```
mcmcgrath13: (comment, b): date ranges have a special format and get split into two parts - this needs to be handled

Also allow nulls need to be serialized

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,134 @@
...
+        filterValues.add(clause);
+
+        if (i < root.rules().size() - 1) {
+          FilterValue operator = buildOperatorFilterValue(advancedFilter, root.combinator());
+          filterValues.add(operator);
+        }
+      } else if (rule instanceof AdvancedQuery.RuleGroup r) {
+        List<FilterValue> ruleGroupValues =
+            fromAdvancedFilterRequest(
+                advancedFilter, new AdvancedFilterRequest(request.reportFilterUid(), r));
```
mcmcgrath13: (sugg, nb): would adding a private method that does the recursion simplify this at all? so we don't need to go back and forth between filter requests and then that can handle the count of the sequence?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,134 @@
...
+    return buildParenFilterValue(advancedFilter, ")");
+  }
+
+  private FilterValue buildParenFilterValue(ReportFilter advancedFilter, String paren) {
+    FilterValue parenValue =
+        FilterValue.builder()
+            .id(generateFilterValueId())
+            .reportFilter(advancedFilter)
+            .valueType(ReportConstants.AdvancedFilterValueType.OPERATOR.toString())
+            .operator(paren)
```
mcmcgrath13: this is almost the same as `buildOperatorFilterValue` - can they be combined? maybe have the operator one call to this one and rename paren to operator?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
JordanGuinn: Does the `WhereClauseService` not already handle `includeNulls`?  And if we still need to do it here for some reason, are we talking about appending an `OR` and a `NE null` advanced FilterValue to the set of BasicFilter values we're building?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -134,6 +156,25 @@ private Report upsertReport(
     return reportRepository.save(report);
   }
 
+  /**
+   * The `save` action, which overwrites the various filter/sort details of a given report, without
+   * changing the actual report mechanics themselves.
+   */
+  @Transactional
+  public Report saveReport(ReportExecutionRequest request, ReportId reportId) {
```
mcmcgrath13: (sugg, nb): since we go fetch the report when doing the ownership check, should we 404 there and change this to truly expect a non-null `Report` instead of a `ReportId`? (or still do the null check here, but pass in the already fetched report?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
JordanGuinn: Yeah using it for the handling of sequence is a good call, will update

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
JordanGuinn: Great catch - will update

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -278,6 +311,132 @@ public ReportExecutionResult executeReport(ReportExecutionRequest request) {
...
+                          .filter(c -> c.getId().equals(columnId))
+                          .findFirst()
+                          .orElseThrow(
+                              () ->
+                                  new NotFoundException(
+                                      "No matching column found for ID " + columnId));
+
+                  DisplayColumn col = displayColumnBuilder.build(report, matchingColumn, seq.get());
+
+                  seq.getAndIncrement();
```
mcmcgrath13:
```suggestion
                  DisplayColumn col = displayColumnBuilder.build(report, matchingColumn, seq.getAndIncrement());
```

does this work?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -278,6 +311,132 @@ public ReportExecutionResult executeReport(ReportExecutionRequest request) {
...
+                  basicFilter.getFilterValues().clear();
+
+                  List<FilterValue> basicFilterValues =
+                      filterValueMapper.fromBasicFilterRequest(basicFilter, matchingReq);
+                  basicFilter.getFilterValues().addAll(basicFilterValues);
+                }
+              });
+    }
+
+    reportFilterRepository.saveAll(basicFiltersById.values());
```
mcmcgrath13, line 356: (q, nb): do we need this saveall or does the report saving at the end automagically handle this if we skip it?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
mcmcgrath13: it would for advanced filters, but for saving we need to serialize it - it's its own special format, see `BasicFilterConfigurationMapper` for how we deserialize it - we need to reverse that

mcmcgrath13: and the whereclause service handles it for building the where, but this is a separate beast in terms of saving to the db

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: I'm not 100% sure tbh - I wouldn't think so though, given we're first filtering out select `ReportFilters` from the report, and then updating the `FilterValues` on those?  Just felt a bit risky to not simply save those changes explicitly here, esp since `saveReport` is already `Transactional`

mcmcgrath13: fair enough, the split between the explicit saving of filters and updating of other things felt out of balance to my eye, but on a very minor level

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -51,79 +49,79 @@ public List<FilterValue> fromBasicFilterRequest(
 
   public List<FilterValue> fromAdvancedFilterRequest(
       ReportFilter advancedFilter, AdvancedFilterRequest request) {
-    List<FilterValue> filterValues = new ArrayList<>();
+    int sequenceNumber = 1;
```
mcmcgrath13: (q, nb): should we use the atomic number here like we do in the other PR?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -51,79 +49,79 @@ public List<FilterValue> fromBasicFilterRequest(
...
           filterValues.add(operator);
+          sequenceNumber++;
         }
       } else if (rule instanceof AdvancedQuery.RuleGroup r) {
         List<FilterValue> ruleGroupValues =
-            fromAdvancedFilterRequest(
-                advancedFilter, new AdvancedFilterRequest(request.reportFilterUid(), r));
+            mapRuleGroupToFilterValues(advancedFilter, r, sequenceNumber);
         filterValues.addAll(ruleGroupValues);
+        sequenceNumber += ruleGroupValues.size();
```
mcmcgrath13: (q/nit, nb): do we need to keep track of the sequence number as we go? or could we add it at the very end with a map/for loop?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
JordanGuinn: We can, though we don't necessarily need to here.  I was doing it in the other PR as a means to bypass the fact that streams are intended to be stateless, which I suppose now that I'm writing this, I _really_ should go and change that to a traditional for loop or something 😭

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
JordanGuinn: Yeah we could do that as well, these sets should never get particularly big, so `O(2n)` vs `O(n)` is kinda whatever here.

Preference?  Would that read easier?

mcmcgrath13: probably the atomic number thing so we don't need to do the math incrementing, or just add it at the end

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,131 @@
...
+public class FilterValueMapper {
+  private final IdGeneratorService idGenerator;
+
+  public FilterValueMapper(IdGeneratorService idGenerator) {
+    this.idGenerator = idGenerator;
+  }
+
+  public List<FilterValue> fromBasicFilterRequest(
+      ReportFilter basicFilter, BasicFilterRequest request) {
+    List<FilterValue> basicFilterValues =
```
mcmcgrath13: this also needs to special handle dates, which use a different value type

also check the base days filter and text filter to see if they use different things

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -47,14 +52,27 @@ public ResponseEntity<ReportId> editReport(
...
+
+    Report existingReport = reportRepository.findById(reportId).orElse(null);
+    if (existingReport == null) {
+      throw new NotFoundException(reportService.getReportNotFoundText(reportId));
+    }
+
+    //  Only the report's owner should have permission to overwrite it
+    //  We might consider investigating into creating a custom pre-authorizer for this sort of
+    //  authorization check, should we ever need ownership permissions beyond this endpoint
+    if (!existingReport.getOwnerUid().equals(user.getId())) {
```
mcmcgrath13, line 71: also need to check they have edit permissions for the type of report they are trying to save (public vs private vs etc)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -278,7 +311,130 @@ public ReportExecutionResult executeReport(ReportExecutionRequest request) {
...
+                  return displayColumnBuilder.build(report, matchingColumn, seq.getAndIncrement());
+                })
+            .toList();
+
+    report.getDisplayColumns().addAll(newDisplayColumns);
+  }
+
+  private void updateSortColumns(Report report, SortSpec sort) {
+    if (sort == null) {
+      report.setReportSortColumns(null);
```
mcmcgrath13: (nit, b): I think `removeOrphans=true` needs to be added to that join annotation for this to work

mcmcgrath13 (commented):
Could you add an e2e that creates a report with all the genres of basic filter and then saves it with values (then saves as it in the follow on PR) (and then deletes it so it cleans up after itself on average)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
JordanGuinn: Should I add some error logging there as well potentially?  The owner of a report not being able to edit it sounds like bad state, unless there _should_ be reports that can't be edited by anyone

JordanGuinn: I suppose there are the admin routes if need truly be, but still seems valuable to know that the owner of a given report can't actually edit it.  Maybe just more specific error response messaging

mcmcgrath13: Yeah, it's very possible to assign someone as the owner of a report and then not allow them to edit it. Is it a bit odd, absolutely. Is it something the UI allows, absolutely. 😭

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,165 @@
...
+                      FilterValue.builder()
+                          .id(generateFilterValueId())
+                          .reportFilter(basicFilter)
+                          .valueType(ReportConstants.BASIC_FILTER_VALUE_TYPE)
+                          .valueTxt(value == null ? "" : value)
+                          .build())
+              .collect(Collectors.toCollection(ArrayList::new)));
+    }
+
+    if (request.includeNulls()) {
```
mcmcgrath13:
```suggestion
    if (request.includeNulls() && request.values().isEmpty()) {
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,165 @@
...
+
+      basicFilterValues.addAll(Arrays.asList(beginRangeVal, endRangeVal));
+    } else {
+      basicFilterValues.addAll(
+          request.values().stream()
+              .map(
+                  value ->
+                      FilterValue.builder()
+                          .id(generateFilterValueId())
+                          .reportFilter(basicFilter)
```
mcmcgrath13, line 74: conditionally add allow nulls here and above as the operator if it makes sense

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,165 @@
...
+              .collect(Collectors.toCollection(ArrayList::new)));
+    }
+
+    if (request.includeNulls()) {
+      basicFilterValues.add(
+          FilterValue.builder()
+              .id(generateFilterValueId())
+              .reportFilter(basicFilter)
+              .operator(ReportConstants.BASIC_FILTER_ALLOW_NULLS_OP)
+              .valueType(ReportConstants.BASIC_FILTER_VALUE_TYPE)
```
mcmcgrath13:
```suggestion
              .valueType("none")
```
but make a constant?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -0,0 +1,165 @@
...
+
+  public List<FilterValue> fromBasicFilterRequest(
+      ReportFilter basicFilter, BasicFilterRequest request) {
+    List<FilterValue> basicFilterValues = new ArrayList<>();
+
+    if (ReportConstants.BAS_TIME_RANGE_TYPES.contains(
+        basicFilter.getFilterCode().getFilterType())) {
+      if (request.values().size() != 2) {
+        throw new IllegalArgumentException(
+            "Time range filter must have exactly two values: start and end");
```
mcmcgrath13: this check might be too strict in the case where there are no values

mcmcgrath13:
```
if (there are values) {
dates + regular (with operator of allow null as needed)
} else if (allow nulls) {
special case allow null row
}
return

File: apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java [hunk 18]
```diff
@@ -0,0 +1,170 @@
...
+
+  private FilterValue buildCloseParenFilterValue(ReportFilter advancedFilter) {
+    return buildOperatorFilterValue(advancedFilter, ")");
+  }
+
+  private FilterValue buildClauseFilterValue(ReportFilter advancedFilter, AdvancedQuery.Rule rule) {
+    String valueTxt = rule.value();
+
+    if (Arrays.asList(
+            ReportConstants.Operator.IN.toString(), ReportConstants.Operator.NN.toString())
```
mcmcgrath13: (nit, nb): given the non-obviousness of `IN` and `NN` meaning, maybe add a comment here on what's going on?

mcmcgrath13: and/or pull out this array list to a static constant with a more clear name?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -47,14 +52,47 @@ public ResponseEntity<ReportId> editReport(
...
+    String authOperationType;
+    ReportConstants.ReportGroup reportGroup =
+        ReportConstants.dbCharToReportGroup(existingReport.getShared());
+
+    authOperationType =
+        switch (reportGroup) {
+          case PUBLIC -> ReportConstants.Permissions.EDITREPORTPUBLIC;
+          case PRIVATE -> ReportConstants.Permissions.EDITREPORTPRIVATE;
+          case REPORTING_FACILITY -> ReportConstants.Permissions.EDITREPORTREPORTINGFACILITY;
+          case TEMPLATE ->
```
mcmcgrath13: (nit, b): so.... while this logic is the logical logic, it doesn't seem to be what NBS 6 does in practice 😭. It's a bit odd really - PR on the UI side here to copy logic from: https://github.com/CDCgov/NEDSS-Modernization/pull/3390

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -284,7 +317,130 @@ public ReportExecutionResult executeReport(ReportExecutionRequest request) {
...
+    report.getReportSortColumns().add(sortColumn);
+  }
+
+  private boolean isAdvancedFilter(ReportFilter filter) {
+    return filter.getFilterCode().getFilterType().equals(ReportConstants.ADV_FILTER_TYPE);
+  }
+
+  private boolean isBasicFilter(ReportFilter filter) {
+    return filter.getFilterCode().getFilterType().startsWith(ReportConstants.BASIC_FILTER_PREFIX);
+  }
```
mcmcgrath13, line 430: (nit, nb): would it make more sense/be possible to add these as helpers on the `ReportFilter`?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportControllerTest.java
```diff
@@ -233,6 +240,160 @@ void deleteReport_should_return_404_status_code_when_report_not_found() {
...
+      ReportExecutionRequest request =
+          new ReportExecutionRequest(reportUid, dataSourceUid, true, null, null, null, null);
+
+      assertThatThrownBy(() -> controller.saveReport(user, reportUid, dataSourceUid, request))
+          .isInstanceOf(ForbiddenException.class)
+          .hasMessageContaining("User does not have permission to save REPORTING_FACILITY reports");
+    }
+
+    @Test
+    void saveReport_should_throw_422_if_report_is_template() {
```
mcmcgrath13: 😭

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
JordanGuinn: Yep, this was discussed and is being handled in the PR for `Save As` - https://github.com/CDCgov/NEDSS-Modernization/pull/3369#discussion_r3500688273

mcmcgrath13: excellent!

mcmcgrath13 (approved):
💾 

(presuming tests fixed to green)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -278,9 +284,28 @@ public String buildBasicWhereFragment(
                               + config.reportFilterUid()));
 
       if (BAS_TYPES.contains(type)) {
-        basicCriteria.add(buildBasicFilterCriteria(filterRequest, column));
+        LOGGER.log(
+            System.Logger.Level.TRACE, "Building IN criteria for filter type %s".formatted(type));
+
+        String standardCriteria = buildBasicFilterCriteria(filterRequest, column);
+        if (!standardCriteria.isEmpty()) {
```
JordanGuinn, line 291: There was a bug in the `WhereClauseService` where basic filters specified without any values resulted in us generating invalid SQL during report execution (i.e. `AND AND AND`) - this now handles those cases.

PR #3369, opened 2026-06-24
APP-668: Save As Report endpoint

JordanGuinn:
## Description

This endpoint builds off of [the work to add a save endpoint](https://github.com/CDCgov/NEDSS-Modernization/pull/3368), and introduces a new `Save As` endpoint, that effectively acts as a combination of `create` and subsequently `save`.

Also included here is the splitting up of the `ReportService` into a few different components, in order to navigate [a Sonar issue that cropped up during this PR](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3369&id=CDCgov_NEDSS-Modernization&open=AZ7_zqGi_9qMPqdaC1EQ) (**Classes should not depend on an excessive number of classes** (aka Monster Class) [java:S6539](https://sonarcloud.io/organizations/cdcgov/rules?open=java%3AS6539&rule_key=java%3AS6539)).  Happy to split it out into a separate PR if that would ease review.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-668)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
```diff
@@ -21,6 +21,18 @@ public FilterValueMapper(IdGeneratorService idGenerator) {
     this.idGenerator = idGenerator;
   }
 
+  public FilterValue duplicate(FilterValue filterValue) {
```
mcmcgrath13: could use `filterValue.toBuilder().id(generateFilterValueId()).build()`

along with


> If using @Builder to generate builders to produce instances of your own class (this is always the case unless adding @Builder to a method that doesn't return your own type), you can use @Builder(toBuilder = true) to also generate an instance method in your class called toBuilder(); it creates a new builder that starts out with all the values of this instance. 

From https://projectlombok.org/features/Builder

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -3,24 +3,93 @@
...
     this.idGenerator = idGenerator;
+    this.reportSortColumnMapper = reportSortColumnMapper;
+    this.reportFilterBuilder = reportFilterBuilder;
+    this.displayColumnBuilder = displayColumnBuilder;
+  }
+
+  public Report duplicate(Report report) {
+    Report newReport =
+        Report.builder()
+            .id(new ReportId(generateReportId(), report.getDataSource().getId()))
```
mcmcgrath13: same `toBuilder` comment as above (and for any other duplicates)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FilterUtils.java
```diff
@@ -0,0 +1,16 @@
+package gov.cdc.nbs.report.utils;
+
+import gov.cdc.nbs.entity.odse.ReportFilter;
+import gov.cdc.nbs.report.ReportConstants;
+
+public class FilterUtils {
+  private FilterUtils() {}
+
+  public static boolean isAdvancedFilter(ReportFilter filter) {
```
mcmcgrath13: (nit/q, nb): would it make sense to make this methods of `ReportFilter` itself?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
mcmcgrath13: this todo still holds and this is the PR to do it

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExecutionClient.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.http.MediaType;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestClient;
+import org.springframework.web.client.RestClientResponseException;
+
+@Service
+public class ReportExecutionClient {
+  private final Clock clock;
+
+  private final RestClient reportExecutionClient;
```
mcmcgrath13: (nit, nb): sonar has a point given this class name

mcmcgrath13: should this be `ReportExecutionService` instead for the class?

JordanGuinn: Ah whoops - I updated this in the test, forgot about it here, will fix

apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FilterUtils.java
JordanGuinn: Good call, will update

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/FilterValueMapper.java
mcmcgrath13: probably want to update the add/last updated times as well if relevant to the thing

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExecutionClient.java
JordanGuinn: > should this be ReportExecutionService instead for the class?

I went with this initially, but eventually talked myself out of it, as it felt a bit too close to the Python report execution service, and you'd always have to differentiate.

JordanGuinn: Thoughts?  Maybe even `ReportExecutionServiceClient`?  Gah, what are words

mcmcgrath13: `ReportExecutionServiceService` 😈

mcmcgrath13: `ReportExecutionClientService`?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -3,24 +3,86 @@
...
+            .addTime(LocalDateTime.now(clock))
+            .addUserUid(user.getId())
+            .status(new Status(Status.ACTIVE_CODE, LocalDateTime.now(clock)))
+            .build();
+
+    if (existingReport.getReportFilters() != null) {
+      List<ReportFilter> newFilters =
+          existingReport.getReportFilters().stream()
+              .map(reportFilterBuilder::duplicate)
+              .map(f -> f.toBuilder().report(newReport).build())
```
mcmcgrath13, line 47: (nit/q, b): do these also need new ids?

mcmcgrath13: ah wait, the duplicate does that?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/BasicFilterConfigurationMapper.java
```diff
@@ -21,7 +21,7 @@ public static BasicFilterConfiguration fromReportFilter(ReportFilter filter) {
 
     FilterType filterType = FilterTypeMapper.fromFilterCode(filter.getFilterCode());
 
-    if (!filterType.type().startsWith(ReportConstants.BASIC_FILTER_PREFIX)) {
+    if (!filterType.type().startsWith(ReportFilter.BASIC_FILTER_PREFIX)) {
```
mcmcgrath13: (nit, very-nb): in the other PR there's an `isBasicFilter` predicate - should we use that here?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SaveAsReportRequest.java
```diff
@@ -9,6 +9,6 @@
 public record SaveAsReportRequest(
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String reportTitle,
     @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank String sectionCode,
-    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull @NotBlank ReportConstants.ReportGroup group,
+    @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull ReportConstants.ReportGroup group,
```
mcmcgrath13, line 12: (q, b): why remove the notblank? when would this be blank?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportFetcher.java
```diff
@@ -0,0 +1,132 @@
...
+@Component
+public class ReportFetcher {
+
+  private final ReportRepository reportRepository;
+
+  public ReportFetcher(ReportRepository reportRepository) {
+    this.reportRepository = reportRepository;
+  }
+
+  public ReportConfiguration getReport(Long reportUid, Long dataSourceUid) {
```
mcmcgrath13, line 37: (q, nb): `getReportRunner` has `@Transactional(readOnly = true)` - should this too?

mcmcgrath13: (or a comment as to why it doesn't?)

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportExecutionServiceClientTest.java
```diff
@@ -0,0 +1,138 @@
...
+        .when(reportFetcher.getReport(reportUid, dataSourceUid))
+        .thenReturn(reportConfig);
+
+    ReportExecutionRequest request =
+        new ReportExecutionRequest(
+            reportUid, dataSourceUid, true, List.of(17L), null, List.of(), null);
+
+    assertThatThrownBy(() -> reportExecutionClient.executeReport(request))
+        .isInstanceOf(NotImplementedException.class)
+        .hasMessage("Report not implemented for java");
```
mcmcgrath13: (nit, nb): shouldn't this say python?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportExecutionServiceClientTest.java
```diff
@@ -0,0 +1,138 @@
...
+        "SELECT * FROM [NBS_ODSE].[dbo].[PHC_Demographic]",
+        LocalDateTime.of(2025, Month.MAY, 5, 12, 23));
+  }
+
+  private ReportConfiguration mockReportConfiguration(boolean isPython) {
+    ReportConfiguration reportConfig = mock(ReportConfiguration.class);
+    Library library = mock(Library.class);
+
+    Mockito.lenient().when(library.isBuiltin()).thenReturn(true);
+    Mockito.lenient().when(library.runner()).thenReturn(isPython ? "python" : "java");
```
mcmcgrath13:
```suggestion
    Mockito.lenient().when(library.runner()).thenReturn(isPython ? "python" : "sas");
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/SaveAsReportRequest.java
JordanGuinn: Per convo with @emyl3, `@NotBlank` doesn't work for enums, only strings

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportFilter.java
```diff
@@ -68,4 +68,15 @@ public class ReportFilter {
...
+  public static final String BASIC_FILTER_PREFIX = "BAS_";
+  public static final String ADV_FILTER_TYPE = "ADV_WCB";
+
+  public boolean isAdvancedFilter() {
+    return getFilterCode().getFilterType().equals(ADV_FILTER_TYPE);
+  }
+
+  public boolean isBasicFilter() {
+    return getFilterCode().getFilterType().startsWith(BASIC_FILTER_PREFIX);
+  }
```
brick-green: q, nb: since this is testing against filter code should this code be placed in `FilterCode`?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -96,29 +104,49 @@ public ResponseEntity<ReportId> saveReport(
...
+          case PRIVATE -> ReportConstants.Permissions.CREATEREPORTPRIVATE;
+          case REPORTING_FACILITY -> ReportConstants.Permissions.CREATEREPORTREPORTINGFACILITY;
+          case TEMPLATE ->
+              throw new IllegalArgumentException(
+                  "Template reports cannot be created using 'saveAs'");
+        };
+
+    String authority = authOperationType + "-" + ReportConstants.Permissions.REPORTINGOBJECT;
+
+    if (user.getAuthorities().stream().noneMatch(a -> a.getAuthority().equals(authority))) {
```
brick-green: sugg, nb: maybe convert to a helper method since it is repeated from line 97.

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -96,29 +104,49 @@ public ResponseEntity<ReportId> saveReport(
...
       @AuthenticationPrincipal NbsUserDetails user,
       @PathVariable Long reportUid,
       @PathVariable Long dataSourceUid,
       @Valid @RequestBody SaveAsReportRequest request) {
-    //  @TODO: Finish implementation NOSONAR
-    return null;
+    String authOperationType;
+    ReportConstants.ReportGroup reportGroup = request.group();
+
+    authOperationType =
```
brick-green, line 131:
```suggestion
    String authOperationType =
```

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportFilter.java
JordanGuinn: At this point we really only care about it from the context of a filter - I can move this logic to an equivalent `isAdvancedFilterCode`/`isBasicFilterCode` method pair, but I imagine we'd still want methods in here referencing those.  Thoughts?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
JordanGuinn: This made me realize that there's already a `hasPermission` method on `NbsUserDetails`, so I'm gonna update this to use that!

libs/database-entities/src/main/java/gov/cdc/nbs/entity/odse/ReportFilter.java
brick-green: oh yeah for sure would need something still in ReportFilter. I meant that the equality comparison could be in FilterCode looking something like this:
```
public static final String BASIC_FILTER_PREFIX = "BAS_";
public static final String ADV_FILTER_TYPE = "ADV_WCB";

public boolean isAdvanced() {
    return ADV_FILTER_TYPE.equals(this.filterType);
}

public boolean isBasic() {
    return this.filterType != null && this.filterType.startsWith(BASIC_FILTER_PREFIX);
}
```

Then in ReportFilter it could be:
```
public boolean isAdvancedFilter() {
    return getFilterCode() != null && getFilterCode().isAdvanced();
}

public boolean isBasicFilter() {
    return getFilterCode() != null && getFilterCode().isBasic();
}
```


idk, definitely don't have big feelings here. I do think some of those null checks are important though.

JordanGuinn: Yep - good call on the null checks, will update

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -18,9 +16,7 @@ private AdvancedFilterConfigurationMapper() {}
 
   public static AdvancedFilterConfiguration fromReportFilter(
       ReportFilter filter, List<DataSourceColumn> columns) {
-    FilterType filterType = FilterTypeMapper.fromFilterCode(filter.getFilterCode());
-
-    if (!filterType.type().equals(ReportConstants.ADV_FILTER_TYPE)) {
+    if (!filter.isAdvancedFilter()) {
```
mcmcgrath13, line 19: ✨

mcmcgrath13 (approved):
🥇

brick-green (approved):

PR #3382, opened 2026-07-01
fix: cleanup types and other small issues

mcmcgrath13:
## Description

* Appease typescript in `src` (splitting NumericInput into two components, one for range and one for plain makes this easier (and is how things work for dates))
* Fix width content limiting in card
* Provide delimiter to result data's csv parser so if there's a one column, header-only result, that doesn't display an error

### DIFF
--- a/apps/modernization-ui/src/design-system/card/card.module.scss
+++ b/apps/modernization-ui/src/design-system/card/card.module.scss
@@ -2,7 +2,7 @@
 @use 'styles/borders';
 @use 'styles/components';
 
-$content-width: (
+$content-max-width: (
     'widescreen': 90rem,
 );
 
@@ -38,7 +38,7 @@ section.card {
         @include components.required();
     }
 
-    @each $name, $value in $content-width {
+    @each $name, $value in $content-max-width {
         &--#{$name} {
             header {
                 div {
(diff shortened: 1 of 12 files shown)


apps/modernization-ui/src/design-system/card/card.module.scss
```diff
@@ -49,7 +49,7 @@ section.card {
 
                 // target the content div
                 + div {
-                    width: $value;
+                    max-width: $value;
```
emyl3, line 52: looks like the advanced query card is not actually getting this max-width applied

mcmcgrath13: I see the content max-widthing as I expected - what are you seeing?

emyl3: I was seeing this:
Screenshot 2026-07-01 at 12 42 09

I forgot to screenshot the width but it was over 1600px -- let me know if you are seeing this too!

mcmcgrath13: image

mcmcgrath13: or wait - that's on my main PR - this might need a fix on call site - please hold

mcmcgrath13: pushed that up

emyl3 (approved):
LGTM! Thank you for the cleanup!! 🤩

PR #3384, opened 2026-07-02
fix: cleanup errors when opening nbs6 saved report with 7

mcmcgrath13:
## Description

Rage-cleanup of error state when you try to open an NBS 6 save'd report with 7. This could happen to end users (in strange scenarios) and not just in dev, so make it a user facing error with a good message. Also assume "sas" as runner if no library found

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
@@ -141,6 +141,12 @@ public class ReportService {
         .findById(id)
         .map(
             report -> {
+              ReportLibrary library = report.getReportLibrary();
+              if (library == null) {
+                throw new IllegalArgumentException(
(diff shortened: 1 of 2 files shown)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
```diff
@@ -141,6 +141,12 @@ public ReportConfiguration getReport(Long reportUid, Long dataSourceUid) {
         .findById(id)
         .map(
             report -> {
+              ReportLibrary library = report.getReportLibrary();
```
JordanGuinn, line 144: (nit, nb): Do you think it'd be worth adding an explicit test for this situation, if there's not one already?

JordanGuinn (approved):
👏 🚀

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportService.java
mcmcgrath13: already added!

PR #3386, opened 2026-07-02
feat: add row checking at point of query, not just at end

mcmcgrath13:
## Description

Add belt to suspenders for ensuring we don't fetch overly large results for reporting


apps/report-execution/src/db_transaction.py
```diff
@@ -42,16 +52,42 @@ def _column_names(self) -> list[str]:
...
 
         # not sure why this is needed - it shouldn't be per docs:
         # https://github.com/microsoft/mssql-python/wiki/Connection#context-manager
         connection.commit()
+
+
+def check_row_limits(num_rows: int, is_export: bool):
+    """Ensure the number of rows in the result set does not exceed limits.
+
+    When the is_export/run status is not available, assume export, which has a large.
```
JordanGuinn:
```suggestion
    When the is_export/run status is not available, assume export, which has a larger limit.
```

JordanGuinn (approved):
📈 ⏹️

apps/report-execution/src/db_transaction.py
mcmcgrath13:
```suggestion
```

apps/report-execution/src/config.py
adaball:
```suggestion
            f'Please ensure it is registered in load_report_configurations().'
```

Looks like a reference to a function that doesn't exist.

adaball (approved):
One small suggestion, overall LGTM :shipit:

PR #3388, opened 2026-07-02
feat: add default state when no sections available to configure

mcmcgrath13:
## Description

Add default state when no sections are available on the run configuration page

image


brick-green (approved):
tested and looks good

PR #3390, opened 2026-07-06
fix: save permissions in ui

mcmcgrath13:
## Description

While reviewing the save api PR, I was checking some logic and found the save permissions are a lil creative in 6. This PR updates the UI to match that logic

image


brick-green (approved):
weird functionality but looks like it matches :shipit:

PR #3391, opened 2026-07-06
[APP-643] PA05 conversion

timoballard:
## Conversion notes
PA05.SAS contains a bug on line 63. A set of datetime variables (MIN_CA_INTERVIEWER_ASSIGN_DT, MAX_CA_INTERVIEWER_ASSIGN_DT, MIN_CC_CLOSED_DT, MAX_CC_CLOSED_DT ) is created, but they are not given an explicit format, and so they default to the number of seconds since 1960-01-01, e.g. 2022019200. The variables then format this value in scientific notation, e.g. 2022E9.

The problem arises when, later on, we derive date values from those variables (line 90), which have effectively been rounded (2022019200 --> 2022E9 --> 2022000000 ). This produces date values that don't match the ones we originally queried from STD_HIV_DATAMART1, resulting in the dataset being truncated of rows that would otherwise be included.

This bug was not replicated in Python, and so we should expect discrepancies for the various “Num. of OI’S” and “Num. of RI’S” fields.

- [x] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelogreport/execution/03_ODSE_Data_Report_Library_Init.sql`

## Tickets

* [Jira Ticket](url)


apps/report-execution/src/libraries/pa_05.py
```diff
@@ -0,0 +1,688 @@
...
+    # affect SELECT DISTINCT row counts), and neither is referenced anywhere
+    # else in PA05.sas either -- they're dead columns in the original report.
+    return f"""
+    WITH shd AS (
+        {subset_query}
+    ),
+    std_cases AS (
+        SELECT shd.*
+        FROM shd
+        INNER JOIN RDB.DBO.INVESTIGATION i
```
brick-green: sugg, b: need to use the configuration value here instead of `RBD`. See pa_03.py for an example.
```suggestion
        INNER JOIN {nbs_rdb}.DBO.INVESTIGATION i
```

apps/report-execution/tests/integration/assets/tablefaker_schema/pa_05.yaml
brick-green, line 1: sugg, nb: can this file be named `pa_05.yaml`? That structure seems to match with the others

brick-green (dismissed):
Need to switch over the DB names using the new `get_cached_config_value()` method

apps/report-execution/src/libraries/pa_05.py
timoballard: ty- fixed in https://github.com/CDCgov/NEDSS-Modernization/pull/3391/commits/78fa1448490ab3bc6332bbe87ee239a233764cf9

apps/report-execution/tests/integration/assets/tablefaker_schema/pa_05.yaml
timoballard: agreed, fixed in https://github.com/CDCgov/NEDSS-Modernization/pull/3391/commits/78fa1448490ab3bc6332bbe87ee239a233764cf9

krista-skylight (commented):
q, nb: I see slight discrepancies in the sas comparison output. are those entirely attributable to the bug?

timoballard:
> q, nb: I see slight discrepancies in the sas comparison output. are those entirely attributable to the bug?

The SAS bug creates discrepancies in the "Num. of IO'S", "Num. of RI's" and "Clusters Init" values

testing/regression/cypress/e2e/features/reports/run.feature
```diff
@@ -51,3 +51,23 @@ Feature: Run report
...
+        Then I should see a "heading" labelled "Your report has opened in a new tab."
+
+    Scenario: I can run a report for library pa_05 (interview assign date)
+        When I navigate to "Public" report with reportUid: 10066745 and dataSourceUid: 23
+        And I enter "04/28/2025" to the From date
+        And I enter "04/28/2026" to the To date
+        And I select "950 - AIDS" from the "DIAGNOSIS_CD" dropdown menu
+        And I select "Fulton LocalUser" from the "INVESTIGATOR_INTERVIEW_QC" dropdown menu
+        And I click the "Run" button
+        Then I should see a "heading" labelled "Your report has opened in a new tab."
```
brick-green, line 93: q, b: Looks like the report targeted here is the same between the tests (`reportUid: 10066745 and dataSourceUid: 23`). Also I think you only need to test one of the reports since the goal is to test the report libraries. Does that feel right?


If we want to test the reports then this `Examples` feature is pretty useful.

```suggestion
    Scenario Outline: I can run a report for library pa_05 (case close date)
        When I navigate to "Public" report with reportUid: <reportUid> and dataSourceUid: <dataSourceUid>
        And I enter "04/28/2025" to the From date
        And I enter "04/28/2026" to the To date
        And I select "950 - AIDS" from the "DIAGNOSIS_CD" dropdown menu
        And I select "Fulton LocalUser" from the "INVESTIGATOR_INTERVIEW_QC" dropdown menu
        And I click the "Run" button
        Then I should see a "heading" labelled "Your report has opened in a new tab."
        Examples:
            | reportUid | dataSourceUid |
            | 10066744  | 23            |
            | 10066745  | 23            |
```

timoballard: agh yeah this is copy/pasta, thanks for catching. This felt more like we were testing the reports than the libraries, so my thinking was that we'd want to ensure that both reports fire without error, but I definitely could be misunderstanding. For the moment I [fixed this to do what I originally intended](https://github.com/CDCgov/NEDSS-Modernization/pull/3391/commits/e6245b3496db8c68ebfa351147e520497de91bd2)

apps/report-execution/src/libraries/pa_05.py
```diff
@@ -0,0 +1,693 @@
...
+            -- operator treats missing = missing as a match. That means an open
+            -- candidate case (CC_CLOSED_DT IS NULL) matches whenever the subset
+            -- itself contains any open case, regardless of whether the subset also
+            -- has closed cases. T-SQL's IN never matches NULL, so this clause
+            -- reproduces that behavior explicitly to keep parity with SAS.
+            OR (std.CC_CLOSED_DT IS NULL AND ob.NUM_OPEN > 0)
+      );
+    """
+
+
```
krista-skylight: (nit, b) I saw Ada put constants and helper functions into a `support` subdirectory in this [pa01 PR](https://github.com/CDCgov/NEDSS-Modernization/pull/3405/changes#diff-39773ef0b660073f9436cd648c1c8850806f00c8884a10f30a72b3c98912444d). I'd suggest following that convention just to keep this module from getting too long! I'm going to do the same for pa02.

apps/report-execution/tests/integration/assets/tablefaker_schema/pa_05.yaml
krista-skylight, line 1: q, nb: any particular reason it was difficult to wrap this into `std_hiv_datamart.yaml`? I know I have some sensitive logic in there to make the potential duplicates library work. Was it hard not to break that? I'm wondering if it makes more sense to separate out potential duplicates into another yaml so other libraries can build on `std_hiv_datamart.yaml` more easily.

krista-skylight (commented):
Just one minor suggestion and question!

apps/report-execution/tests/integration/assets/tablefaker_schema/pa_05.yaml
timoballard: this may have been a misunderstanding on my part -- I was under the impression that we were leaning toward a 1:1ish mapping between libraries and tablefaker schemas so that it's clear what needs to be loaded to get any particular library to have data. Seems like a tradeoff between avoiding having many tablefaker setups to maintain for some of the the same underlying tables vs. the (presumably) increasing complexity of keeping a small number of schemas in a state where they work for all of the libraries

happy to try to fold this into std_hiv_datamart if that's the direction we want (or have already decided) we want to go

krista-skylight: My understanding was we should try to wrap everything up in `std_hiv_datamart.yaml` as much as possible. @mcmcgrath13 can you confirm?

apps/report-execution/src/libraries/pa_05.py
timoballard: ➕  rearranged things similarly to PA01 here: https://github.com/CDCgov/NEDSS-Modernization/pull/3391/commits/01fba6e433235096354e2836adbbe18758eb78ca

apps/report-execution/tests/integration/assets/tablefaker_schema/pa_05.yaml
mcmcgrath13: If it's reasonable to re-use a schema, that's preferred. That way if we find an issue with the state of a field, we have fewer places to check/maintain it (such as we did with the program jurisdiction oid in the past)

timoballard: consolidated in https://github.com/CDCgov/NEDSS-Modernization/pull/3391/commits/02d0754841a07027de4656789a1dd571f44297a9

krista-skylight (approved):
thank you, lgtm!

PR #3406, opened 2026-07-08
[APP-696] feat: actually use sort spec in libraries

mcmcgrath13:
## Description

Use the order by clause passed by the api in the report execution service

Notes:
* It turn out you *cannot* use an order by in a CTE, so including the order by in the subset query is problematic. Due to this, it was removed from the subset query and libraries that need to use sort (those that have `allow_column_selection: 'Y'` in their metadata) need to incorporate it where they need it
* You also cannot order by the same column twice, so needed to do a bit of refactoring to the potential duplicates library to make this logic robust
* `uv.lock` is a trailing change from report hardening, but does make sense

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-696


apps/report-execution/src/libraries/potntl_dup_inv_sum.py
```diff
@@ -42,6 +50,14 @@ def execute(
         days_value = 3650
 
     select_clause = ', '.join([f'd.[{item[1]}]' for item in column_map])
+    # can't sort by the same column twice - remove sort_by from the base sort
+    base_order_by = ', '.join(
+        [
+            f'd.[{col_dict[col["name"]]}] {col["sort"]}'
+            for col in REQUIRED_COLS
+            if sort_by is None or f'[{col_dict[col["name"]]}]' not in sort_by
```
adaball, line 58: comment/gripe: string matching seems a bit fragile (it depends on `sort_by` having its columns wrapped in `[]` and assumes the case of both the `col_dict` value and the column in the string will be the same) but I honestly can't see a different way of accomplishing the goal here.

adaball (approved):
LGTM :shipit:

apps/report-execution/src/libraries/potntl_dup_inv_sum.py
mcmcgrath13: Yeah, I also felt 🥴 about it, but this is a closed part of the implementation and not public API, so it seems safe if not perfectly ideal

PR #3410, opened 2026-07-08
[APP-694] SAS Conversion for SR19 and SR20

brick-green:
## Description

This PR converts the SR19 and SR20 SAS libraries. These have been combined and named nbs_sr_19. The difference between the two SAS libraries was the date column used. Typically that would be part of the subset_query but in this library the output converts the date to MMMYYYY format. So the column needs to be passed down from the report to the library. In report-execution this is accomplished via the `library_params`.

## Tickets

* [Jira Ticket](url)

- [X] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelogreport/execution/03_ODSE_Data_Report_Library_Init.sql`
  - [x] Consolidate reports by using parameters (parameter design tbd)
  - [x] Improve the report spec or result expectations


apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,120 @@
...
+    trx: Transaction,
+    subset_query: str,
+    data_source_name: str,
+    library_params: dict,
+    **kwargs,
+):
+    """SR19: TB Record Count - Count of cases by "countability" and month.
+
+    Conversion Notes:
+    * Parameterized date column handled safely.
```
mcmcgrath13: (nit, b): could you explicitly note that SR20 is implemented in here as well

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,120 @@
...
+    * Pushes heavy aggregations to SQL Server using fast integer date functions.
+    """
+
+    # --- Helper Function ---
+    def get_sas_date(year, month, day=1):
+        sas_epoch = date(1960, 1, 1)
+        target_date = date(year, month, day)
+        return (target_date - sas_epoch).days
```
mcmcgrath13: (nit, nb): move this outside of the main function to the top level

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # --- Parameter Validation ---
+    if not isinstance(library_params, dict) or 'count_column' not in library_params:
+        raise ValueError(
```
mcmcgrath13: (nit, b): instead of generic `ValueError` here, could you create a library mis-configured type exception with an appropriate status code in the `errors.py` file (and also use this in qa07... 🙏 )

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,120 @@
...
+        non_counted = int(row[3])
+
+        db_map[(y, m)] = (counted, non_counted)
+
+    # --- Track timeline boundaries ---
+    years_months = list(db_map.keys())
+    start_date = date(years_months[0][0], years_months[0][1], 1)
+    end_date = date(years_months[-1][0], years_months[-1][1], 1)
+
+    # --- MonthYear Matrix Build to fill in months with no cases ---
```
mcmcgrath13: (sugg, nb): sr_05 also has this problem, but fills in the 0s in sql - could that solution work here as well? or too different of a case?

apps/report-execution/src/libraries/nbs_sr_19.py
mcmcgrath13: (sugg, nb): can sqls `DATEDIFF` be used to push this calculation into sql instead of post-hoc

apps/report-execution/tests/integration/assets/tablefaker_schema/tb_datamart.yaml
```diff
@@ -16,3 +16,12 @@ tables:
...
       - column_name: calc_disease_site
         data: random.choice(["Pulmonary", "Extra Pulmonary", "Both", "Unknown"])
+      - column_name: count_date
+        data: fake.date_between(start_date=datetime(2020, 1, 1), end_date=datetime(2025, 12, 31))
+        null_percentage: 0.1
+      - column_name: date_reported
+        data: fake.date_between(start_date=datetime(2020, 1, 1), end_date=datetime(2025, 12, 31))
+        null_percentage: 0.1
+      - column_name: count_status
+        data: r"Count as a TB Case"
```
mcmcgrath13: (nit, maybe-b): the count date and count status being "count" should probably be co-populated and not independent variables in reality. Not sure that matters for library testing or not 🤔

apps/report-execution/tests/integration/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,147 @@
...
+    def test_execute_report_check_data(self, snapshot):
+        report_spec = ReportSpec.model_validate(
+            {
+                'is_export': True,
+                'is_builtin': True,
+                'report_title': 'SR 19',
+                'library_name': 'nbs_sr_19',
+                'data_source_name': '[RDB].[dbo].[TB_DATAMART]',
+                'subset_query': 'SELECT * FROM [RDB].[dbo].[TB_DATAMART]',
+                'library_params': '{"count_column": "COUNT_DATE"}',
```
mcmcgrath13: (nit, b): can we also add coverage for the `REPORT_DATE`

apps/report-execution/tests/integration/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,147 @@
...
+        status_query = """
+            SELECT 
+                CAST('2025-06-01' AS DATE) AS COUNT_DATE, 
+                'Count as a TB Case' AS count_status
+            UNION ALL
+            SELECT CAST('2025-06-02' AS DATE) AS COUNT_DATE, NULL AS count_status
+            UNION ALL
+            SELECT 
+                CAST('2025-06-03' AS DATE) AS COUNT_DATE, 
+                'Suspected Case' AS count_status
```
mcmcgrath13, line 169: (q, nb): would incorporating more statuses into the fake data make this simpler?

apps/report-execution/src/libraries/nbs_sr_19.py
mcmcgrath13: If you move that into SQL, then you can also move the sql stuff into sql, and the date formatting. Thoughts on whether that's worthwhile?

mcmcgrath13:
(request, b): could you upload the comparison test results to the SAS report catalog?

apps/report-execution/src/libraries/nbs_sr_19.py
brick-green: yup, add this one

apps/report-execution/tests/integration/libraries/nbs_sr_19.py
brick-green: yes though kind of like having an explicit test for it

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,96 @@
...
+):
+    """SR19: TB Record Count - Count of cases by "countability" and month.
+
+    Conversion Notes:
+    * Parameterized date column handled safely. This change allowed for the
+    merging of SR20 into SR19 as the input date column was the only difference
+    * count_column is now a required library_param
+    """
+    # --- Parameter Validation ---
+    if not isinstance(library_params, dict) or 'count_column' not in library_params:
```
mcmcgrath13: (nit, b): this field should really be named `date_column` or similar and not `count_column`

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,96 @@
...
+            WHERE {count_column} IS NOT NULL
+        ),
+        
+        -- compile list of month/years filling in missing months
+        calendar AS (
+            SELECT start_date AS month_date, end_date 
+            FROM boundaries 
+            WHERE start_date IS NOT NULL
+            UNION ALL
+            SELECT DATEADD(month, 1, month_date), end_date
```
mcmcgrath13, line 52:
```suggestion
            SELECT start_date AS month_date
            FROM boundaries 
            WHERE start_date IS NOT NULL
            UNION ALL
            SELECT DATEADD(month, 1, month_date)
```

that column doesn't seem to be used

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,96 @@
...
+        -- join to fill in missing months
+        SELECT 
+            UPPER(FORMAT(c.month_date, 'MMMyyyy')) AS monthYear,
+            DATEDIFF(day, '1960-01-01', c.month_date) AS sasdate,
+            CAST(ISNULL(a.counted_cases, 0) AS INT) AS counted_cases,
+            CAST(ISNULL(a.non_counted_cases, 0) AS INT) AS non_counted_cases,
+            CAST(
+                ISNULL(a.counted_cases, 0) 
+                + ISNULL(a.non_counted_cases, 0) AS INT
+            ) AS total_cases
```
mcmcgrath13: (q, nb): why do we need these casts? would using `COALESCE` instead of `ISNULL` simplify anything?

testing/regression/cypress/step_definitions/report/run.steps.js
```diff
@@ -42,6 +42,16 @@ When('I enter {string} to the To date', (date) => {
     cy.get('input[id$="-to"]').type(date);
 });
 
+When('I enter From Month: {string} and From Year: {string}', (month, year) => {
+    cy.selectDropdownByLabel(0, "From Month", month);
+    cy.selectDropdownByLabel(0, "From Year", year);
```
mcmcgrath13: heads up that a PR is open that changes these to `From month` and `From year` - whichever merges second will need to update this

apps/report-execution/src/errors.py
```diff
@@ -106,3 +106,10 @@ class ConfigurationIntegrityError(BaseReportExecutionError):
 
     def __init__(self, message: str):
         super().__init__(f'NBS_Configuration Data Integrity Error: {message}', 422)
+
+
+class InvalidLibraryParamsError(BaseReportExecutionError):
```
mcmcgrath13, line 111: could you either in this PR or a ticket/follow on also update qa_07 to use this error message 🙏

mcmcgrath13:
(request, b): could you also upload a comparison test for SR20?

apps/report-execution/src/errors.py
brick-green: done!

testing/regression/cypress/step_definitions/report/run.steps.js
brick-green: updated

apps/report-execution/src/libraries/nbs_sr_19.py
brick-green: I think when I was converting from python my mind was in python land still and not SQL land. I agree that `COALESCE` is cleaner.

apps/report-execution/src/libraries/nbs_sr_19.py
brick-green: It's used in the where clause to break the recursion when the start_date is no longer less than the end_date

apps/modernization-api/src/main/resources/db/report/execution/libraries/nbs_sr_20.sql
```diff
@@ -0,0 +1,50 @@
+-- Migrate the NBSSR00020.SAS library to the nbs_sr_19 python library
```
mcmcgrath13:
```suggestion
-- Migrate the NBSSR00020.SAS library to the nbs_sr_20 python library
```

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,94 @@
...
+        ),
+        
+        -- Find the first month/year and last month/year from the subset
+        boundaries AS (
+            SELECT 
+                MIN(CAST(DATEFROMPARTS(
+                    YEAR({date_column}), MONTH({date_column}), 1) AS DATE)
+                ) AS start_date,
+                MAX(CAST(DATEFROMPARTS(
+                    YEAR({date_column}), MONTH({date_column}), 1) AS DATE)
```
mcmcgrath13: (sugg, nb): `DATEFROMPARTS` returns a date - can the casts be dropped?

apps/report-execution/src/libraries/nbs_sr_19.py
```diff
@@ -0,0 +1,94 @@
...
+    data_source_name: str,
+    library_params: dict,
+    **kwargs,
+):
+    """SR19: TB Record Count - Count of cases by "countability" and month.
+
+    Conversion Notes:
+    * Parameterized date column handled safely. This change allowed for the
+    merging of SR20 into SR19 as the input date column was the only difference
+    * date_column is now a required library_param
```
mcmcgrath13, line 18: (nit, b): can you note here the choice to include the total even when non-counted is 0 as that is a difference from sas?

testing/regression/cypress/e2e/features/reports/run.feature
```diff
@@ -66,6 +66,18 @@ Feature: Run report
...
+    Scenario Outline: I can run a report for library <library>
+        When I navigate to "Public" report with reportUid: <reportUid> and dataSourceUid: <dataSourceUid>
+        And I enter From Month: "1" and From Year: "2025"
+        And I enter To Month: "12" and To Year: "2025"
+        And I click the "Run" button
+        Then I should see a "heading" labelled "Your report has opened in a new tab."
+
+        Examples:
+            | library | reportUid | dataSourceUid |
+            | nbs_sr_19   | 10066724  | 15            |
```
mcmcgrath13:
```suggestion
            | library     | reportUid | dataSourceUid |
            | nbs_sr_19   | 10066724  | 15            |
```

mcmcgrath13 (approved):
a couple super minor suggestions/nits, but overall looks great!

PR #3416, opened 2026-07-09
test: e2es for saving/(as) every filter

mcmcgrath13:
## Description

- Adds e2e test for running/saving as with every filter applied
- Adds e2e test for running/saving a report with just "Include nulls" checked

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-806)

## Checklist before requesting a review
- [ ] PR focuses on a single story
- [ ] Code has been fully tested to meet acceptance criteria
- [ ] PR is reasonably small and reviewable (Generally less than 10 files and 500 changed lines)
- [ ] All new functions/classes/components reasonably small
- [ ] Functions/classes/components focused on one responsibility
- [ ] Code easy to understand and modify (clarity over concise/clever)
- [ ] PRs containing TypeScript follow the [Do's and Don'ts](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)
- [ ] PR does not contain hardcoded values (Uses constants)
- [ ] All code is covered by unit or feature tests

PR #3422, opened 2026-07-13
[APP-818] fix: enable keyboard submit

mcmcgrath13:
## Description

As part of #3416 found that keyboard-based submit didn't work. This PR fixes that and adds more robust testing of permissions on the report run page in general

Notes:
* Submit button needs to be inside `form`, so moved that out to `ReportRunPage`
* Handle submit still working when only `Export` button allowed


JordanGuinn (approved):

apps/modernization-ui/src/apps/report/run/ReportRunPage.tsx
```diff
@@ -150,7 +151,9 @@ const ReportRunPage = () => {
         </>
     ) : status === 'configuring' ? (
         <FormProvider {...form}>
-            <ReportConfigurationPage config={config} handleSubmit={onSubmit} />
+            <form onSubmit={(e) => onSubmit(e, !canRunReport)}>
```
JordanGuinn, line 154: 📈

PR #3430, opened 2026-07-14
APP-812: Remove "Regions" filters and filter codes

JordanGuinn:
## Description

This PR removes the `Regions` and `Regions (Including NULLS)` filter codes from the DB, along with all corresponding ReportFilters that reference it, and all places across the codebase.  It also adds `ON DELETE CASCADE` to the ReportFilter FKs in FilterValue and ReportFilterValidation, so those get automatically deleted when their parent report is deleted.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-812)


JordanGuinn:
It should be noted that currently, this change requires clearing the `report.filter.options` key from local storage in order to have the changes reflected correctly in the UI, specifically in the basic filter dropdown.  Are we okay with that (maybe with just a bit of messaging around it), or should we try to handle that programmatically on our end?

apps/modernization-api/src/main/resources/db/odse/02_ODSE_REPORT_FILTER_CASCADE_DELETES.sql
```diff
@@ -0,0 +1,31 @@
...
+  Alters the two ReportFilter foreign key constraints in the Report_Filter_Validation and Filter_value
+  tables, specifically to cascade deletes from the Report_Filter table.
+ */
+USE [NBS_ODSE]
+
+-- Cascade deletes from Report_Filter to Report_Filter_Validation
+IF OBJECT_ID('dbo.FK_ReportFilter_ReportFilterValidation', 'F') IS NULL
+BEGIN
+    -- Remove legacy auto-gen'd constraint name
+    ALTER TABLE dbo.Report_Filter_Validation
```
mcmcgrath13: (sugg, probably-b): the altering the schema here makes me a bit nervous, especially the second key that looks auto-genned in name. What about instead doing a `SELECT INTO` temp table to collect the filters we need to delete, then deleting all the things?

JordanGuinn: Yeah that works too, I just remember us talking about potentially harnessing cascading deletes for this work, so I thought this was desired.

apps/modernization-api/src/main/resources/db/changelog/odse-changelog.yml
```diff
@@ -8,3 +8,7 @@ databaseChangeLog:
             path: db/odse/01_ODSE_NBS_NOTE.sql
             splitStatements: false
             endDelimiter: ""
+        - sqlFile:
```
mcmcgrath13: (nit, b): this should be part of the report changelog and not the general ODSE one - it might be fine with 6, but I don't want to need to test it on 6 and 7

JordanGuinn: Will update - figured it had nothing to do with report execution, so didn't seem appropriate initially

krista-skylight (approved):
lgtm, thank you!

mcmcgrath13 (approved):
Looks great! NBS 6 does seem to also respect the removal across the board, but I think the messaging is a bit clearer to have it tied to report module adoption. @maggiemanteuffel-tech something to touch base with CDC on before cutting the ga release

PR #3432, opened 2026-07-14
[APP-751] fix: fix up error page styling

mcmcgrath13:
## Description

Update the error page styling to be less bare bones. Skipped the part of the design about the go back button if we're in NBS as it turned out to be finnicky and the last thing we want in an error page is finnicky code that might error or cause errors

image

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-751

### DIFF
--- a/apps/modernization-ui/src/components/FullPageBlock/FullPageBlock.tsx
+++ b/apps/modernization-ui/src/components/FullPageBlock/FullPageBlock.tsx
@@ -2,7 +2,7 @@ import { ReactNode } from 'react';
 import classNames from 'classnames';
 import styles from './full-page-block.module.scss';
 
-// Needs to be inside a space that will grow to space, otherwise defaults to minimum height
+// Will not take up full page unless inside an element that will grow to the full space, otherwise defaults to minimum height
 const FullPageBlock = ({ className, children }: { className?: string; children?: ReactNode }) => {
     return <div className={classNames(className, styles.layout)}>{children}</div>;
 };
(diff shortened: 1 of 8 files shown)


apps/modernization-ui/src/components/FullPageBlock/FullPageBlock.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from 'react';
+import classNames from 'classnames';
+import styles from './full-page-block.module.scss';
+
+// Needs to be inside a space that will grow to space, otherwise defaults to minimum height
```
JordanGuinn: What does this comment mean?

JordanGuinn (approved):

apps/modernization-ui/src/components/FullPageBlock/FullPageBlock.tsx
mcmcgrath13: Basically, the block will take up the space available, so if the thing it's in is not a full page thing, then it won't fill it - thought on how to re-word?

mcmcgrath13:
```suggestion
// Will not take up full page unless inside an element that will grow to the full space, otherwise defaults to minimum height
```

JordanGuinn: Updated comment LGTM ✅

PR #3433, opened 2026-07-15
[APP-766] fix: keep multi-select open while selecting

mcmcgrath13:
## Description

The `Multiselect` component uses `react-select`, which relies on controlling focus on its dummy input. The use of the standard `checkbox` input inside the menu list was causing the focus to shift away and the menu to then close. This PR introduces a `DummyCheckbox` which looks like a checkbox, but defers the control and semantics (the outer div already has the correct aria tags) to `react-select`

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-766


apps/modernization-ui/src/apps/report/run/filters/basic/OptionSelectFilter.tsx
```diff
@@ -34,7 +34,7 @@ const OptionSelectFilter: BasicFilterComponent = ({ filter, value, onChange, ...
     if (filter.selectType === BasicFilterConfiguration.selectType.SINGLE) {
         return (
             <SelectInput
-                value={value?.[0] ?? null}
+                value={value?.[0] ?? undefined}
```
mcmcgrath13, line 37: there was an error in the console complaining about this being null, so translated it over to undefined instead. As long as the form sees `null`, we're good

JordanGuinn: This was changed to `null` as part of the fix for https://github.com/CDCgov/NEDSS-Modernization/pull/3421, correct?

Is this saying that the value in the input itself can/should remain `undefined` if it's absent, but needs to be presented as `null`  in the `onChange` handler?  If so, why?

mcmcgrath13: Exactly. The form needs null, but the component needs undefined as the null-ish state

brick-green (approved):
tested and works! 🥇

PR #3434, opened 2026-07-15
fix: use NavLinkButton for internal link buttons

mcmcgrath13:
## Description

Small fix to use `NavLinkButton` instead of `LinkButton` for internal nbs 7 links as the performance is much zippier


brick-green (approved):
much zippier indeed

JordanGuinn (approved):

PR #3436, opened 2026-07-15
[APP-700]: TB_Summary_Count and TB_Summary_Count_Date library conversion

brick-green:
## Description

This PR converts the TB_Summary_Count and TB_Summary_Count_Date libraries from SAS to python

## Tickets

* [APP-700](https://cdc-nbs.atlassian.net/jira/software/c/projects/APP/boards/626?selectedIssue=APP-700)

- [X] Check the existing description for the library in `apps/modernization-api/src/main/resources/db/changelogreport/execution/03_ODSE_Data_Report_Library_Init.sql`
  - [X] If the report execution service is not yet in use by any STLT (even beta), you should update this file
  - [X] Update the description in the library-specific migration
  - [X] Make sure pertinent text embedded in the reports is included in the description of the report
  - [X] Note in the doc string any deviation from the SAS
- [X] Keep an eye out for things in the sas library that are/could be centralized to pre- or post-processing or a util would be helpful for (e.g. common formatting needs)
- [X] Add integration tests in a `apps/report-execution/tests/libraries/<your_library_here>py` file following the conventions established for other libraries
  - [X] The `subset_sql` can be assumed to be well tested by the modernization-api, so focus on any logic and additional joins/queries/analysis that is added in the library
  - [X] The database used in integration testing is our standard golden db + seed used in other testing - if additional data is needed to test the report, add it to the seeding process
- [X] Add e2e test coverage for the report
- [X] Check that the report has parity with the NBS 6 version using (https://cdc-nbs.atlassian.net/wiki/spaces/NE/pages/2372272139/Comparison+Testing+Libraries)
- [X] Keep an eye out for opportunities to:
  - [X] Consolidate reports by using parameters (parameter design tbd)
  - [X] Improve the report spec or result expectations
- [X] Update the [SAS to Python tracking spreadsheet](https://docs.google.com/spreadsheets/d/1xpl-znNQIql_tii39i5WYPIrO6vsLRqLBG1dbtuq8-4/edit?usp=sharing) with information about the newly converted library  
- [X] Update/clarify these instructions in the epic based on your experience

[APP-700]: https://cdc-nbs.atlassian.net/browse/APP-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/modernization-api/src/main/resources/db/report/execution/libraries/tb_summary_count.sql
```diff
@@ -0,0 +1,50 @@
+-- Migrate the NBSSR00019.SAS library to the nbs_sr_19 python library
```
adaball: suggestion: update the SQL comment with the `tb_summary_count` SAS file and library names

apps/modernization-api/src/main/resources/db/report/execution/libraries/tb_summary_count_date.sql
```diff
@@ -0,0 +1,50 @@
+-- Migrate the NBSSR00019.SAS library to the nbs_sr_19 python library
```
adaball: suggestion: update the SQL comment with the `tb_summary_count_date` SAS file and library names

apps/report-execution/src/libraries/nbs_sr_20.py
```diff
@@ -7,11 +7,12 @@ def execute(
     trx: Transaction,
     subset_query: str,
     data_source_name: str,
+    library_params: dict,
     **kwargs,
 ):
     """Wrapper report leveraging SR19 core engine with a modified column layout."""
     nbs_sr_19_report_result = execute_nbs_sr_19(
-        trx, subset_query, data_source_name, **kwargs
+        trx, subset_query, data_source_name, library_params, **kwargs
```
adaball: question: Why is this line being updated?  It seems like it's unrelated to the PR.

apps/report-execution/src/libraries/tb_summary_count.py
```diff
@@ -0,0 +1,123 @@
+from src.config import get_cached_config_value
+from src.db_transaction import Transaction
+from src.errors import InvalidLibraryParamsError
+from src.models import ReportResult
+
+
+def execute(
+    trx: Transaction,
+    subset_query: str,
+    data_source_name: str,
```
adaball: suggestion: you can take out `data_source_name` here as it's not part of the ReportSpec anymore.

adaball (commented):
I've put the raw data comparison [in this spreadsheet](https://docs.google.com/spreadsheets/d/1wEI9ZzaT6qnGI4w8ZlL-bHT4hgS637IDM3FlV8DBHss/edit?usp=sharing).

I ran the reports locally with both SAS and Python using the date range `01/2005` to `07/2026` and noticed 2 quirks:

1. The SAS report for both variants will output rows in `MM/YYYY` dates that don't have source data but are included in the input date range.  The SQL query that goes into the Python report is where the date range is so it's not possible to get the From and To dates to include them in the report.  I don't think this is a bug, I think it's a limitation of having the filters coming into the Python library via the `subset_query`.  I think it would be worth mentioning in the conversion notes.
~2. The SAS report of the Report Date variant is missing `total_cases` values for some rows where the Python report has them.  I think this may be a SAS bug and is worth mentioning in the conversion notes.~ I can see you already put this in the conversion notes.

apps/report-execution/src/libraries/nbs_sr_20.py
brick-green: Ooops, I was trying something then forgot to revert.

apps/report-execution/src/libraries/tb_summary_count.py
```diff
@@ -0,0 +1,125 @@
...
+            INNER JOIN {nbs_ods}.dbo.NBS_RDB_METADATA rdb 
+                ON ui.NBS_UI_METADATA_UID = rdb.NBS_UI_METADATA_UID
+            INNER JOIN {nbs_srt}.dbo.CONDITION_CODE cc 
+                ON cc.INVESTIGATION_FORM_CD = ui.INVESTIGATION_FORM_CD
+            INNER JOIN {nbs_srt}.dbo.CODESET cs 
+                ON cs.CODE_SET_GROUP_ID = ui.CODE_SET_GROUP_ID
+            INNER JOIN {nbs_srt}.dbo.CODE_VALUE_GENERAL cvg 
+                ON cs.CODE_SET_NM = cvg.CODE_SET_NM
+            WHERE ui.QUESTION_IDENTIFIER IN ('INV1109')
+              AND cc.CONDITION_CD = '102201'
```
timoballard, line 53: (nit, nb): maybe pull these out into named constants

adaball (approved):
Once the tests stabilize LGTM

:shipit:

PR #3442, opened 2026-07-20
APP-872: Escape '%' & '[' characters in LIKE clauses

JordanGuinn:
## Description

This PR updates the `FieldFormatter` to escape the wildcard and escape characters `%` and `[`, respectively, so they can be included in `LIKE` clauses as part of `CONTAINS` and `STARTS WITH` report filters.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-872)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/utils/FieldFormatter.java
```diff
@@ -12,7 +12,7 @@
 public class FieldFormatter {
   public String formatField(String type, String value) {
     return switch (type.toUpperCase()) {
-      case "STRING" -> "'" + value.replace("'", "''") + "'";
+      case "STRING" -> "'" + value.replace("'", "''").replace("%", "[%]") + "'";
```
mcmcgrath13: (q, maybe-b): does this mean `[` also need to be escaped?

mcmcgrath13: quick sanity test tells me yes 😭

mcmcgrath13: (and presumably ']')

JordanGuinn: Was just thinking that, will update

JordanGuinn: Not immediately sure if I need to escape closing brackets if we're already escaping opens, will check

JordanGuinn: Confirmed, looks like we should only need to escape the opening ones

mcmcgrath13 (approved):
🙃 ]

PR #3451, opened 2026-07-21
[APP-877] fix: handle missing database objects better

mcmcgrath13:
## Description

Surface missing data sources and columns as user-facing 422's instead of an internal system error

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-877

### DIFF
--- a/apps/report-execution/src/db_transaction.py
+++ b/apps/report-execution/src/db_transaction.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from contextlib import contextmanager
 
 import mssql_python
@@ -6,6 +7,9 @@ import mssql_python
 from . import errors
 from .config import get_cached_config_value
 from .models import Table
+
+INVALID_OBJECT_REGEX = re.compile("Invalid object name ('.*').")
+INVALID_COLUMN_REGEX = re.compile("Invalid column name ('.*').")
 
 
 class Transaction:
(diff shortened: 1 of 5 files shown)


apps/report-execution/tests/integration/execute_report.py
```diff
@@ -429,3 +430,42 @@ def execute_method(self, trx, subset_query, data_source_name, **kwargs):
...
+                'is_export': True,
+                'is_builtin': True,
+                'report_title': 'Test Report',
+                'library_name': 'nbs_custom',
+                'data_source_name': '[NBS_ODSE].[dbo].[Filter_operator]',
+                'subset_query': 'SELECT nope FROM [NBS_ODSE].[dbo].[Filter_operator]',
+            }
+        )
+
+        # TODO: error handling for sql issues  # noqa: FIX002
```
brick-green: nit, nb: can you add more context here? I'm not sure what this means

apps/report-execution/src/db_transaction.py
```diff
@@ -24,7 +28,24 @@ def query(self, query: str, parameters: tuple = ()) -> Table:
...
+                    'Datasource', datasource_match.group(1)
+                ) from None
+
+            col_match = INVALID_COLUMN_REGEX.search(e.message)
+            if col_match is not None:
+                raise errors.MissingDbObjectError(
+                    'Column', col_match.group(1)
+                ) from None
+
+            # re-reaise
```
brick-green:
```suggestion
            # re-raise
```

apps/report-execution/src/libraries/qa_07.py
```diff
@@ -36,9 +36,9 @@ def execute(
             (e.g., 30, 60, 90), got {library_params}
         """)
     days = library_params.get('days_value')
-    if days is None:
+    if days is None or type(days) is not int or days < 0:
         raise InvalidLibraryParamsError(f"""
-            library_params must contain 'days_value' \
+            library_params must contain positive integer 'days_value' \
```
brick-green: q, b: the if checks for `days < 0` then the message says must contain `positive integer`. Which one should it be?

brick-green (approved):
A few smaller things. Close enough for a ✅ for me though :)

apps/report-execution/src/libraries/qa_07.py
mcmcgrath13: if the days is < 0 means that we currently have a non-positive integer, so we're in a bad state

apps/report-execution/tests/integration/execute_report.py
mcmcgrath13: left over copy paste from the generic "bad sql" = 500 test - removed it

apps/report-execution/src/libraries/qa_07.py
brick-green: sorry, I meant that it's checking for less that zero. So it could be set to zero and zero isn't a positive integer

mcmcgrath13: ahhhhh - I think 0 is technically fine, though a very constricted check, I'll change the wording to `non-negative`

PR #3453, opened 2026-07-21
[APP-852] fix: better logging setup for gateway

mcmcgrath13:
## Description

Make sure errors when reaching out to mod api actually log and generally check our default log application yml

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-852

### DIFF
--- a/apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportExecutionRouteLocatorConfiguration.java
+++ b/apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportExecutionRouteLocatorConfiguration.java
@@ -134,6 +134,10 @@ class ReportExecutionRouteLocatorConfiguration {
               .expand(getParamsFromBody(exchange))
               .toUriString();
 
+      LOGGER.log(
+          System.Logger.Level.DEBUG,
+          "Querying modernization-api for report metadata: %s".formatted(path));
+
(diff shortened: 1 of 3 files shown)


apps/nbs-gateway/src/main/java/gov/cdc/nbs/gateway/report/ReportExecutionRouteLocatorConfiguration.java
```diff
@@ -167,6 +171,10 @@ private AsyncPredicate<ServerWebExchange> isModPredicate(
                 return request
                     .exchangeToMono(
                         response -> {
+                          if (!response.statusCode().is2xxSuccessful()) {
```
timoballard, line 174: what does this part mean functionally? That we won't silently failover to nbs6?

mcmcgrath13: we will still fail over to nbs6, but we'll log about it along the way instead of fully silently doing it. Do you think we should instead fail out the transaction?

timoballard: no I think failover + log is the way to go 👍

timoballard (approved):

PR #3455, opened 2026-07-21
fix: add warning logs when encryption fails with the error details

mcmcgrath13:
## Description

For sanity reasons (this is a relatively common mis-configuration path)


JordanGuinn (approved):
🔍

PR #3463, opened 2026-07-23
fix: handle empty date ranges

emyl3:
## Description
- Fixes a bug in the basic filter with a date range that is not required, a date is entered and then cleared -> resulted in a 422 error
- With this fix, the report should run

A good report to test with is `/report/10066731/22/run` (Case Lab Line Listing)

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-888)

### DIFF
--- a/apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
+++ b/apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
@@ -468,10 +468,7 @@ public class WhereClauseService {
     String colType = column.sourceTypeCode();
     List<String> formattedValues = new ArrayList<>();
 
-    if (!values.isEmpty()) {
-      if (values.stream().allMatch(String::isEmpty)) {
-        return "";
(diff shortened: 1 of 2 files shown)


apps/modernization-api/src/main/java/gov/cdc/nbs/report/WhereClauseService.java
```diff
@@ -469,6 +469,9 @@ private String buildBetweenCriteria(
     List<String> formattedValues = new ArrayList<>();
 
     if (!values.isEmpty()) {
+      if (values.stream().allMatch(String::isEmpty)) {
+        return "";
+      }
```
```suggestion
if (!values.isEmpty() && !values.stream().allMatch(String::isEmpty) ) {
```

(q, b): how does this play with the `includeNulls` case? I'd think we still want to hit that logic

mcmcgrath13: could you add a test for that combo?

mcmcgrath13 (approved):
📆

PR #3470, opened 2026-07-27
fix: page overflow

mcmcgrath13:
## Description

I noticed on a very long error message the page wasn't scrolling. Add `overflow-x-auto` to our layout so pages will scroll when the content is too wide

(plus fixed a typo)


emyl3 (approved):
Codewise this looks good to me... How would you repro this? 🤔

mcmcgrath13:
> Codewise this looks good to me... How would you repro this? 🤔

You can make your screen skinny and see that it allows scrolling now vs just cutting off. You can also use the react dev tools to manually set the state of the runreportpage to `status: 'submitting'` and the second state to a really long string

apps/modernization-ui/src/apps/report/layout/ReportLayout.tsx
```diff
@@ -1,6 +1,6 @@
 import { ReactNode, useId } from 'react';
 import { SkipLink } from 'SkipLink';
-import { ReportHeader, ReportHeaderProps } from './ReporHeader';
+import { ReportHeader, ReportHeaderProps } from './ReportHeader';
```
JordanGuinn, line 3: 🙏

JordanGuinn (approved):

PR #3474, opened 2026-07-27
override 'js-yaml' to version '4.3.0' to fix vulnerability

adaball:
## Description

This PR overrides the transitive dependency `js-yaml` to version `4.3.0` which fixes the vulnerability without having to update any parent packages.

Vulnerability reports:
- https://github.com/CDCgov/NEDSS-Modernization/security/dependabot/312
- https://github.com/CDCgov/NEDSS-Modernization/security/dependabot/316


emyl3 (approved):
Thank you, ada! 🚢

PR #3476, opened 2026-07-27
override version for 'js-yaml' to '4.3.0' to fix vulnerability in cypress tests

adaball:
## Description

This PR is for the cypress testing found at `testing/regression`.

Sets `js-yaml` to version `4.3.0` in overrides to fix transitive dependency without changing versions of parent packages.

Vulnerability: https://github.com/CDCgov/NEDSS-Modernization/security/dependabot/316


emyl3 (approved):
🫡

PR #3480, opened 2026-07-27
build: allow disabling of secure cookie usage

mcmcgrath13:
## Description

Turns out NBS 7 just doesn't work at all on safari in local development because https isn't used and per-spec, secure cookies only work on https (even though other browsers are more forgiving)


brick-green (approved):
tested and works too 🥇

PR #3482, opened 2026-07-27
fix: allow `esc` to cancel confirmation modal unless disabled

mcmcgrath13:
## Description

Per 508 review, allow esc to cancel the confirmation modals. When the modal's interaction is disabled (such as when saving/deleting), don't allow esc to do anything as the user could end up in a weird state


brick-green (approved):
tested and works for me 🏆

PR #3489, opened 2026-07-28
fix: on-this-page stickiness, card consistency

mcmcgrath13:
## Description

#3470 did fix page overflow, but lost the stickiness of the aside. This PR fixes that and also makes sure the cards all take up the full column width and also don't let their contents overflow

Testing notes:
* open any report and note the "On this page" side nav is sticky
* ensure when the screen shrinks, the main content is scrollable (all screens)
* check the cards are all the same size (full width)

### DIFF
--- a/apps/modernization-ui/src/apps/report/layout/ReportLayout.tsx
+++ b/apps/modernization-ui/src/apps/report/layout/ReportLayout.tsx
@@ -18,7 +18,7 @@ const ReportLayout = ({ noSkipLink = false, children, ...headerProps }: ReportRu
             <header id={headerId}>
                 <ReportHeader {...headerProps} />
             </header>
-            <main className="overflow-x-auto">{children}</main>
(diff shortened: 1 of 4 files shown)


emyl3:
Looks like at certain widths when the header gets longer and has to wrap the sidebar gets cut a little... Not sure what widths we wanted to support...

Screenshot 2026-07-28 at 15 53 06

emyl3 (approved):
This works as expected for me! Code-wise looks good too. Do you think we need to run this by Kenny?

emyl3:
If Kenny ends up needing to review, here's some screenshots!

Screenshot 2026-07-29 at 08 36 41

⚠️ Longer titles will need truncation even at "wider" screen sizes with this change

Screenshot 2026-07-29 at 08 37 01

But side nav experience works consistently now at smaller widths

Screenshot 2026-07-29 at 08 38 39

PR #3490, opened 2026-07-28
[APP-893] perf: refactor new tab opening to stay in app

mcmcgrath13:
## Description

Change the approach to opening a new tab with the report result to improve performance.

Notes:
* Users noted a flash of unstyled content when loading results (most notable with slower internet, or when watching the e2e's go by locally)
* This is likely due to the browser caching of styles not translating to the bare non-NBS-scoped window
* This approach creates a dummy report result route to open, which takes a result id as a param
* Before opening the tab, the result data is serialized and saved to local storage by the run/configuration page to then be read by the new result tab
* The data is deleted from local storage once read
* A warning is displayed if there is no data in local storage
* We attempt to warn users about leaving the page and losing the data, but it requires the user to interact with the page, so doesn't work 100% of the time
* User who noted the bad vibe, confirms the vibe is now good
* e2e's also seem to be about 5 minutes faster

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-893


.github/workflows/Build-and-deploy-modernization-api.yaml
```diff
@@ -9,6 +9,7 @@ on:
       - "apps/modernization-ui/**"
       - "libs/**"
       - "gradle/libs.versions.toml"
+      - ".github/workflows/Build-and-deploy-modernization-api.yaml"
```
emyl3, line 12: I don't think we need this change anymore right since you were trying to test your branch? 👀 Or did we want to keep it for ease of testing later?

apps/modernization-ui/src/routes/AppRoutes.tsx
```diff
@@ -46,6 +47,20 @@ const routing: RouteObject[] = [
             ...reportRouting,
         ],
     },
+    // This path is for rendering report results. It opens in a new tab and relies on
```
emyl3, line 50: 🫶 🥲

.github/workflows/Build-and-deploy-modernization-api.yaml
mcmcgrath13: I wanted to keep it as in principal if I change a workflow, I always want that workflow to run and make sure it worked

apps/modernization-ui/src/apps/report/run/ResultDataPage.spec.tsx
```diff
@@ -58,4 +66,26 @@ describe('ResultDataPage', () => {
...
+
+        expect(getByRole('heading', { name: 'No result found' })).toBeVisible();
+    });
+
+    describe('loadReportResult', () => {
+        it('reads data from storage and parses when available', async () => {
+            window.localStorage.setItem('reportResult.1', '{"hi": 2}');
+            const res = await loadReportResult({ params: { resultId: '1' } } as unknown as LoaderFunctionArgs<any>);
+            expect(res).toEqual({ hi: 2 });
+            expect(window.localStorage.getItem('reportResult.1')).toBeNull();
```
emyl3, line 82: 👍

emyl3 (approved):
code-wise looks good to me and initial testing works! Going to merge as per discussion with Mary and deploy to alpha

PR #3492, opened 2026-07-29
[APP-910] fix: advanced filter dnd

mcmcgrath13:
## Description

The advanced filter drag and drop broke when the error handling was added as it turns out the query builder dnd has special handling for rules and rule groups and this over-rode it. Move the passing of those control elements to the dnd provider directly to make things happy again.

https://cdc-nbs.atlassian.net/browse/APP-910


JordanGuinn (approved):
🙏

PR #3494, opened 2026-07-29
APP-809: Improve logging within reports module

JordanGuinn:
## Description

This PR adds a handful of log statements across the reports module.  It also adds configuration for Tomcat HTTP access logs (filtered down to just NBS API requests), and additionally includes a unique UUID as part of each unique 4xx/5xx error for debugging assistance across the system.

## Ticket

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-809)


apps/modernization-api/src/main/java/gov/cdc/nbs/config/AccessLogFilter.java
```diff
@@ -0,0 +1,25 @@
...
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AccessLogFilter implements Filter {
+  private static final String NBS_BASE_PATH = "/nbs/api";
```
mcmcgrath13: (sugg, nb): could you add some commentart in docs on what this filter is for and why this base path (there are others also in the app - so not sure if it's worth making this a list of things to check to make it more clear it's meant to be slightly living/expandable?)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
```diff
@@ -36,6 +36,9 @@ public static AdvancedFilterConfiguration fromReportFilter(
         exceptionMsg = e.getMessage();
         LOGGER.log(System.Logger.Level.WARNING, "Unable to parse saved advanced filter query", e);
       }
+    } else {
+      LOGGER.log(
```
mcmcgrath13, line 44: (sugg, nb): I like the inclusion of the id here - should we add that to the logs on line 37 as well?

mcmcgrath13: (and really any other existing logs in here where we could add an id?)

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
```diff
@@ -32,6 +34,8 @@ public ReportMapper(
   }
 
   public Report duplicate(Report existingReport, NbsUserDetails user) {
+    LOGGER.log(System.Logger.Level.TRACE, "Duplicating report " + existingReport.getId());
```
mcmcgrath13, line 37: (q, nb): why trace here vs debug?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -56,6 +60,8 @@ public ResponseEntity<ReportId> editReport(
       @PathVariable Long reportUid,
       @PathVariable Long dataSourceUid,
       @Valid @RequestBody AdminReportRequest request) {
+    LOGGER.log(System.Logger.Level.TRACE, "EDIT report request received from user " + user.getId());
```
mcmcgrath13: (sugg, nb): should we also include the report+ds uids on these logs? and

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -193,7 +221,11 @@ public ResponseEntity<ReportExecutionResult> runReport(
   @PostMapping("/export")
   @PreAuthorize("hasAuthority('EXPORTREPORT-REPORTING')")
   public ResponseEntity<ReportExecutionResult> exportReport(
-      @Valid @RequestBody ReportExecutionRequest request) {
+      @Valid @RequestBody ReportExecutionRequest request,
+      @AuthenticationPrincipal NbsUserDetails user) {
+    LOGGER.log(
```
mcmcgrath13, line 245: (q, nb): Did you end up exploring [using tracing](https://docs.spring.io/spring-boot/reference/actuator/tracing.html)? I'm guessing it's heavier than we want to do at this point, but curious about what you found

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
```diff
@@ -19,60 +20,108 @@ public class ReportExceptionHandler {
...
   public record ErrorResponseBody(
-      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull String message) {}
+      @Schema(requiredMode = Schema.RequiredMode.REQUIRED) @NotNull String message, String id) {}
 
   @ExceptionHandler(MethodArgumentNotValidException.class)
   public ResponseEntity<ErrorResponseBody> handleValidationExceptions(
       MethodArgumentNotValidException ex) {
+    String errorId = UUID.randomUUID().toString();
+    LOGGER.log(
+        System.Logger.Level.ERROR, DEFAULT_ERROR_LOG.formatted(errorId, ex.getMessage()), ex);
```
mcmcgrath13: (sugg, b): For exception types that are not necessarily system errors, can we downgrade the log level to something lower? I'm not sure exactly what is appropriate, but I don't want to pollute the logs with un-actionable errors

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
```diff
@@ -19,60 +20,108 @@ public class ReportExceptionHandler {
...
   @ExceptionHandler(HttpMessageNotReadableException.class)
   public ResponseEntity<ErrorResponseBody> handleFailedSerialization(
       HttpMessageNotReadableException ex) {
+    String errorId = UUID.randomUUID().toString();
+    LOGGER.log(
+        System.Logger.Level.ERROR, DEFAULT_ERROR_LOG.formatted(errorId, ex.getMessage()), ex);
+
     return new ResponseEntity<>(
-        new ErrorResponseBody(ex.getMessage()), HttpStatus.UNPROCESSABLE_ENTITY);
+        new ErrorResponseBody(ex.getMessage(), errorId), HttpStatus.UNPROCESSABLE_ENTITY);
```
mcmcgrath13: (sugg, nb): this chunk of code is very repetitive throughout the file - add a helper method?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
```diff
@@ -19,60 +20,108 @@ public class ReportExceptionHandler {
...
+      LOGGER.log(
+          System.Logger.Level.WARNING,
+          "Unexpected error response shape from ReportExecutionServiceClient: %s"
+              .formatted(ex.getResponseBodyAsString()),
+          e);
+    }
+
+    String message =
+        err != null && err.message != null ? err.message : ex.getResponseBodyAsString();
+    String errorId = err != null && err.id != null ? err.id : UUID.randomUUID().toString();
```
mcmcgrath13, line 81: these sorts of code paths make me really appreciate js's nullish operators 😭

apps/modernization-api/src/main/resources/application-development.yml
```diff
@@ -1,30 +1,40 @@
...
-          core: INFO
-
     nbs:
       classic:
         outgoing: INFO   # Logs any RestTemplate interactions with NBC Classic
     gov:
       cdc:
         nbs:
           search: INFO
+          report: INFO
```
mcmcgrath13, line 37: (q, nb): should/can we move this up and just have `gov.cdc.nbs: INFO`?

apps/report-execution/src/main.py
```diff
@@ -38,7 +38,9 @@ def execute_report_api(report_spec: models.ReportSpec):
 @app.exception_handler(errors.BaseReportExecutionError)
 async def api_exception_handler(request: Request, exc: errors.BaseReportExecutionError):
     """Handle application errors."""
+    logging.error(f'Error ({exc.id}) occurred during report execution: {exc.message}')
```
mcmcgrath13: (sugg, b): similar to in the java code, can we split this so we only error on 500's and do a lower log level for 400's to not polute the logs with things that are fine

mcmcgrath13 (commented):
mostly questions, some level/pollution concerns. Also added a commit to pull through the ID to the UI

apps/modernization-api/src/main/resources/application-development.yml
JordanGuinn: I was thinking that initially, but ended up going this route as enabling `DEBUG` or `TRACE` fully across the board could make things real noisy real fast.  So figured we could benefit from having some granularity there, and control what modules we make noisy

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
JordanGuinn: Yep, I ended up getting a POC working using OTLP pushing to Jaeger locally, just in a separate branch (ended up going the OTEL route since it looks like we already collect such logs for higher level cloud env health check things).  PR for that will come separately

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
JordanGuinn: Can do - was there more you wanted me to add?  Looks like you got cut off

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
JordanGuinn: No good reason outside of it felt like a `TRACE` when I wrote it ¯\_(ツ)_/¯ Happy to adjust

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/AdvancedFilterConfigurationMapper.java
JordanGuinn: can do

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
JordanGuinn: Can update to `WARN` - were you thinking this for all of them then?  None of these are really system errors - the closest thing would probably be our generic `Exception` handler, but only cuz that's a catch-all

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
JordanGuinn: Can do - I left them duplicated initially b/c even though they are largely identical now, they don't actually need to be, and didn't wanna be prescriptive about that

apps/report-execution/src/main.py
JordanGuinn: I would't categorize these 422 errors as being fine, but will downgrade them to `WARNING`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/mappers/ReportMapper.java
mcmcgrath13: it's borderline for me - no strong feeling either, was curious if you had a clearer line

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
mcmcgrath13: nope - I think that's it, just bad at typing

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportExceptionHandler.java
mcmcgrath13: the system internal one should definitely be error

mcmcgrath13: warn or info or debug maybe? it's not like we plan to do anything about them unless it turns out we're wrong about the cause - hmmm - maybe an info with the message, and a debug with the full exception/stack trace?

JordanGuinn: During development I believe the presence of these errors typically indicated bugs in the UI, or at least mismatches between what the UI and API expected.  So I would def expect we want to do something about most of these if we encounter them (maybe save for `Forbidden` and `Not Implemented`)

mcmcgrath13: I generally agree, but it's also technically a public UI that anyone (with credentials) can use, which could make this noisy. I'm not sure what the right balance is here tbh. I guess what's the standard development log level vs production log level?

apps/report-execution/src/errors.py
mcmcgrath13: Could we add the id to the logging here to tie it to the next log on the handler?

apps/report-execution/src/main.py
```diff
@@ -38,7 +38,16 @@ def execute_report_api(report_spec: models.ReportSpec):
 @app.exception_handler(errors.BaseReportExecutionError)
 async def api_exception_handler(request: Request, exc: errors.BaseReportExecutionError):
     """Handle application errors."""
+    if exc.http_code >= 500:
+        logging.error(
+            f'Error ({exc.id}) occurred during report execution: {exc.message}'
+        )
+    else:
+        logging.warning(
+            f'Error ({exc.id}) occurred during report execution: {exc.message}'
```
mcmcgrath13: should this have a different message since it's an exception, but not necessarily an errorerror (and just grepping through could be tricky?)

mcmcgrath13 (approved):
could super minor remaining thoughts/questions, but overall this is a really nice observability improvement!

PR #3504, opened 2026-07-30
ci: enforce no warnings, make any an error

mcmcgrath13:
## Description

Fail CI on warnings, make using `any` an error (had downgraded it to warning in a prior life as there were too many to fail CI on)


emyl3 (approved):
:shipit:

PR #3505, opened 2026-07-30
[APP-914] fix: rules of hooks

mcmcgrath13:
## Description

Turn on 'rules-of-hooks' and fix up manually the findings


apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
             ],
             'react/react-in-jsx-scope': 'off',
             'react/no-unescaped-entities': 'off',
-            'react-hooks/rules-of-hooks': 'off',
+            'react-hooks/rules-of-hooks': 'warn',
```
mcmcgrath13, line 54: This was off, but this is good info and largely relatively easily addressed. Turn on.

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            // KLUDGE: this should be on, but doesn't play well with some of the current patterns
             'react-hooks/exhaustive-deps': 'off',
```
mcmcgrath13, line 56: I tried to turn this on, but it was a lot and there are a lot of cases where we do expressions in dependency lists which this doesn't play nicely with - needs more thinking in general

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            'react/jsx-curly-brace-presence': [2, 'never'],
```
mcmcgrath13, line 57: Don't use curly braces around things when we don't need to `e.g. prop={"hi"}` is fixed to `prop="hi"`

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            'react/jsx-boolean-value': [2, 'always'],
```
mcmcgrath13, line 58: don't use implicit true values e.g. `required` => `required={true}`

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
             'storybook/hierarchy-separator': 'off',
+            'dot-notation': 'error',
```
mcmcgrath13, line 59: always use dot notation when possible e.g. `hi["there"]` => `hi.there`

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            eqeqeq: 'error',
```
mcmcgrath13, line 61: error on `==` and `!=` (this is JS not java or python or so many other languages 😈 )

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            'no-var': 'error',
```
mcmcgrath13, line 62: don't use `var` - it's evil (on average)

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            'object-shorthand': 'error',
```
mcmcgrath13: use es6 fanciness for object shorthand always e.g. `{x: x}` => `{x}`

apps/modernization-ui/eslint.config.mjs
```diff
@@ -51,9 +51,17 @@ export default defineConfig([
...
+            'prefer-const': 'error',
```
mcmcgrath13, line 63: always use `const` when we don't actually reassign a var (`let` only for mutable cases)

apps/modernization-ui/eslint.config.mjs
```diff
@@ -82,6 +90,7 @@ export default defineConfig([
             ...js.configs.recommended.rules,
             '@typescript-eslint/no-explicit-any': 'off',
             '@typescript-eslint/no-unused-vars': 'off',
+            'react-hooks/rules-of-hooks': 'off',
```
mcmcgrath13, line 94: be a bit less strict on tests and stories

emyl3 (approved):

PR #3510, opened 2026-07-30
[APP-914] fix: add rule eqeqeq: 'error'

mcmcgrath13:
## Description

Add rule eqeqeq: 'error' and manually mass find/replace `!=` => `!==` and ` == ` => ` === ` in .ts and .tsx files

### DIFF
--- a/apps/modernization-ui/eslint.config.mjs
+++ b/apps/modernization-ui/eslint.config.mjs
@@ -56,6 +56,7 @@ export default defineConfig([
             'react/jsx-curly-brace-presence': [2, 'never'],
             'react/jsx-boolean-value': [2, 'always'],
             'dot-notation': 'error',
+            eqeqeq: 'error',
             'storybook/hierarchy-separator': 'off',
         },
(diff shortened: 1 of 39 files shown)


apps/modernization-ui/eslint.config.mjs
```diff
@@ -56,6 +56,7 @@ export default defineConfig([
             'react/jsx-curly-brace-presence': [2, 'never'],
             'react/jsx-boolean-value': [2, 'always'],
             'dot-notation': 'error',
+            eqeqeq: 'error',
```
emyl3, line 59: ❓ Do we need to surround in single quotes?

emyl3 (commented):
Left one question!

apps/modernization-ui/eslint.config.mjs
mcmcgrath13: no, it's a valid object key. Most of the rule names aren't

JordanGuinn (approved):

PR #3514, opened 2026-07-30
[APP-914] exhaustive deps, but only reports things

mcmcgrath13:
## Description

I started going down the turning on exhaustive deps check rabbit hole and it's just too big and risky. I looked at the ones in the report app and did some clean up there instead

Notes:
* Mostly not including functions in deps
* The `activeId` really didn't need to be state with fresh eyes, just made it a regular variable
* Made the option loading more hook-y so the linter better understood what was happening


apps/modernization-ui/eslint.config.mjs
```diff
@@ -9,6 +9,85 @@ import globals from 'globals';
...
+    'react/jsx-boolean-value': [2, 'always'],
+    'dot-notation': 'error',
+    'object-shorthand': 'error',
+    eqeqeq: 'error',
+    'no-var': 'error',
+    'prefer-const': 'error',
+    'storybook/hierarchy-separator': 'off',
+    'object-shorthand': 'error',
+
+    'import/default': 'error',
```
JordanGuinn, line 34: (nit, nb): Do we wanna bundle the `import` rules into a group similar to the `jsDocRules` while we're at it?

apps/modernization-ui/package-lock.json
```diff
@@ -45,7 +45,6 @@
                 "@graphql-codegen/typescript-react-apollo": "^4.4.1",
                 "@storybook/addon-docs": "^10.4.2",
                 "@storybook/addon-onboarding": "^10.3.4",
-                "@storybook/react": "^10.5.6",
```
JordanGuinn, line 48: Sorry about that, I think that was me 🤦

JordanGuinn (approved):

apps/modernization-ui/eslint.config.mjs
mcmcgrath13: I'd like the import rules to always apply to all our code, so leaving it in base config feels right to me - where would you see using it separate from that?

JordanGuinn: Nowhere, didn't realize that was your intention

brick-green (approved):

PR #3517, opened 2026-08-03
refactor: simplify useConfiguration as it is always loaded in practice

mcmcgrath13:
## Description

When working on #3514 I was staring at the `ready` dependency and wondering "why do we need this again?" and realized we really don't. In practice, the configuration is a one and done and always present - this PR simplifies the code to make that a reality

Notes:
* I confirmed with @mpeels that there's not greater meaning behind why this is the way it is that removing the complexity would cause problems for us in general
* While this is ready for review now, I don't plan to merge until post-GA as this goes beyond our current semi-freddo code state


JordanGuinn (approved):
📈 🙏

PR #3519, opened 2026-08-03
APP-916, Part 1: Apply import ordering across modernization UI

JordanGuinn:
## Description

This PR adds the requisite ESLint plugin for import ordering and applies a handful of rules taken largely from [DIBBS ECR Viewer](https://github.com/CDCgov/dibbs-ecr-viewer/blob/main/containers/ecr-viewer/eslint.config.mjs), with a few caveats:

- Imports without corresponding identifiers (i.e. `import './my/thing.js'`) are ignored, as it didn't seem appropriate to me to bundle them along with their corresponding import groups.  Open to feedback here though!
- Because of our `baseUrl` set as `src` and us not using path aliasing, there are lots of places where internal imports are imported using project-scoped absolute paths.  And unfortunately, there [doesn't appear to be an easy way](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md#how-imports-are-grouped) to let ESLint distinguish between these imports and external imports with an equivalent import format 😕 I've tried a few things, including the strategies outlined [here](https://github.com/import-js/eslint-plugin-import/issues/807#issuecomment-640393439) and [here](https://github.com/import-js/eslint-plugin-import/issues/2008#issuecomment-801263294), but in all cases, **non-relative `internal` imports and `external` imports are both treated as `external`, and subsequently grouped together**.  We can probably more easily address this as part of [the work to do away with `baseUrl` entirely](https://cdc-nbs.atlassian.net/browse/APP-645) (as it will be removed in TypeScript 7), but that felt much beyond the scope of this ticket.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-916)


mcmcgrath13 (approved):
🤩 

so tidy!

I think it's find the external muddies with internal right now as it's really the fault of the basePath/alias situation that needs separate dealing with

Were there any other rules from import that are worth incorporating?

JordanGuinn:
@mcmcgrath13 Oh yeah lots!  In my head this ticket was just for ordering - I can put up a separate PR for other rules, just so the diff is more apparent

emyl3 (approved):
So organized!!! Thank you, Jordan! 🤩

PR #3521, opened 2026-08-03
[APP-914] test: fix up console emitting issues - part 2

mcmcgrath13:
## Description

🥈 - this gets two more tests passing - all related to page builder, which isn't prod yet, so not thoroughly fixed fixes (and man there's a lot of repetition in that part of the code base that should probably be revisited someday)


JordanGuinn (approved):
Looks good to me, though this seems to be primarily styling changes rather than console text fixes?  Is that intended?

PR #3540, opened 2026-08-04
[APP-917] Automated typescript check

brick-green:
## Description

This PR addresses several existing TypeScript errors across component, test, and configuration files. In addition, it establishes a dedicated type-checking step in the `Verify Modernization UI` workflow to ensure type safety across the development lifecycle.

To run type checking: `npm run typecheck`

## Tickets

* [Jira Ticket](url)


apps/modernization-ui/tsconfig.app.json
```diff
@@ -1,10 +1,11 @@
 {
     "compilerOptions": {
         /* base vite config */
+        "allowJs": true,
         "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
-        "target": "ES2020",
+        "target": "ES2023",
```
mcmcgrath13: does this have implications for browser support?

apps/modernization-ui/tsconfig.app.json
```diff
@@ -25,8 +26,17 @@
...
+        "paths": {
+            "*": ["./src/*"]
+        },
+        "types": ["vite/client", "node"]
     },
-    "exclude": ["src/**/*.spec.ts", "src/**/*.spec.tsx", "src/setupTests.ts"],
+    "exclude": [
+        "src/**/*.spec.ts",
+        "src/**/*.spec.tsx",
+        "src/**/*.test.tsx",
```
mcmcgrath13: (sugg, nb): instead of add this as a path, can we rename any `.test.tsx` files to `.spec.tsx` to match the standard?

apps/modernization-ui/tsconfig.app.json
brick-green: hmmm yeah there is some implication. I reverted the target but left for `lib`

apps/modernization-ui/tsconfig.app.json
brick-green: yup good call. It turned out to only be one file

apps/modernization-ui/src/apps/deduplication/configuration/match-configuration/notification-cards/AlgorithmNotConfigured.tsx
```diff
@@ -1,6 +1,7 @@
-import { Button } from 'design-system/button';
 import { useNavigate } from 'react-router';
 
+import { Button } from 'design-system/button';
```
JordanGuinn, line 3: I see there are hundreds of these import changes throughout this PR (which is **great**, seems like ESLint is finally distinguishing between internal and external packages somehow?!).

Guess I'm just curious where they came from?  They don't seem like the kind of changes I'd expect from a TS check - maybe from the removal of `baseUrl` and the equivalent specification of `paths`?

apps/modernization-ui/tsconfig.app.json
```diff
@@ -1,10 +1,11 @@
...
     "compilerOptions": {
         /* base vite config */
+        "allowJs": true,
         "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
         "target": "ES2020",
         "useDefineForClassFields": true,
-        "lib": ["ES2020", "DOM", "DOM.Iterable"],
+        "lib": ["ES2023", "DOM", "DOM.Iterable"],
         "module": "ESNext",
         "skipLibCheck": true,
```
JordanGuinn, line 11: (thought, super-nb): I wonder how far off we far from being able to set `strict` to `true`, and if that's something we might consider pursuing if it's not actually a huge lift

apps/modernization-ui/src/apps/deduplication/configuration/match-configuration/notification-cards/AlgorithmNotConfigured.tsx
brick-green: I was seeing that too and I think it's because two things combined:

- Like you said updating the `paths` configuration allowed the ESLint resolver to distinguish internal from external packages.
- The script had `"lint:fix": "eslint --fix"` without the `.`. So I ran `npx eslint --fix .` to apply import sorting rules from the root. I just remembered that I forgot to update the script line so I'll go do that.

mcmcgrath13:
Let me know if this is a tedious ask, but this PR is rather large. Could you separate out the change to the paths stuff and the auto fix of that to a separate PR so that can be more skim-reviewed. And then leave the true typescript fixes for this PR for reviewing once that's merged in?

And/or, are there other ways to reasonably break down this review (but keep checks green along the way)?

brick-green:
> Let me know if this is a tedious ask, but this PR is rather large. Could you separate out the change to the paths stuff and the auto fix of that to a separate PR so that can be more skim-reviewed. And then leave the true typescript fixes for this PR for reviewing once that's merged in?
> 
> And/or, are there other ways to reasonably break down this review (but keep checks green along the way)?

yup, I should have thought of that. Just opened another [PR](https://github.com/CDCgov/NEDSS-Modernization/pull/3553). Just changing the path and running lint:fix changed 500+ files

apps/modernization-ui/tsconfig.app.json
brick-green: I'm not following. In this tsconfig I already see `"strict": true`. Are you talking about something else?

JordanGuinn: 🤦 My bad, I was basing that assumption off the diff, not the _entire_ `tsconfig` - ignore me!

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.spec.ts
```diff
@@ -19,7 +18,7 @@ describe('validateRule', () => {
                 field: '~',
                 operator: '~',
                 value: '',
-            } as RuleType,
+            } as unknown as QbQuery,
```
emyl3: shouldn't these be only `as QbQuery`?

apps/modernization-ui/src/apps/page-builder/page/management/edit/section/manage/ManageSectionTile/ManageSectionTile.tsx
```diff
@@ -66,10 +66,10 @@ export const ManageSectionTile = ({
                             <div className={styles.warningModalContent}>
                                 <div className={styles.content}>
                                     <div className={styles.warningDrag} {...provided.dragHandleProps}>
-                                        <IconComponent name="drag" size="3" />
+                                        <IconComponent name="drag" size="3" alt="" />
```
emyl3: I think we might want to add "drag handle" or something like that as the alt here

emyl3: possibly for the other drag icons too that don't have a label associated with it

apps/modernization-ui/src/design-system/date/validateDateEntry.spec.ts
emyl3: when i run `npm run typecheck` I am seeing this error:
```
src/design-system/date/validateDateEntry.spec.ts:3:1 - error TS6133: 'internalizeDate' is declared but its value is never read.

import { internalizeDate } from 'date';
```

emyl3 (commented):
Left a couple comments! Let me know what you think, Brick! 🤩

apps/modernization-ui/src/design-system/date/validateDateEntry.spec.ts
brick-green: hmmm I assume that is from one of the merges from main. I re-ran the github job and saw the same error which is good so I removed the unused import

apps/modernization-ui/src/apps/page-builder/page/management/edit/section/manage/ManageSectionTile/ManageSectionTile.spec.tsx
```diff
@@ -88,14 +88,14 @@ describe('when ManageSectionTile renders', () => {
                                 <ManageSectionTile
                                     section={section}
                                     index={1}
-                                    setSelectedForDelete={vi.fn}
+                                    setSelectedForDelete={vi.fn()}
```
JordanGuinn, line 91: Nothing really to be done about it if TS was complaining, but it's kinda surprising to see we actually need to _invoke_ the functions supplied to these properties?  👀

apps/modernization-ui/src/apps/page-builder/services/pagesAPI.ts
```diff
@@ -1,7 +1,7 @@
 import { PageControllerService, PagePageSummary, PagesService, PageSummaryService } from 'apps/page-builder/generated';
 
 export const fetchPageSummaries = (
-    token: string,
+    _token: string,
```
JordanGuinn, line 4: (q, nb): I see a handful of places where the `_` character is prepended to specific parameters, presumably to indicate they're not being used.  Are we not able to remove these parameters outright in these cases?

JordanGuinn (approved):
🔥 👏 🚀

apps/modernization-ui/src/apps/page-builder/condition/search/ConditionTable.tsx
```diff
@@ -82,11 +82,15 @@ export const ConditionTable = ({ conditions, isLoading, onSelectionChange, onSor
         onSelectionChange(selected);
     }, [selected]);
 
-    const handleSelect = (event: ChangeEvent<HTMLInputElement>, item: { id: number }) => {
+    const handleSelect = (event: ChangeEvent<HTMLInputElement>, item: TableBody) => {
+        if (item.id === undefined || item.id === null) return;
+
+        const itemId = typeof item.id === 'number' ? item.id : Number(item.id);
```
mcmcgrath13: (nit, b): this could end up with the itemId being `NaN` if the id was truly a string this could end up weird. We know from practice that the rows here really are number, instead of changing this function, could we add a local `type ConditionTableBody = TableBody & { id: number}` and then use that for the state and here? I think that will get us better signal if someone ever goes and decides not to use a number as an ID

apps/modernization-ui/src/apps/page-builder/page/management/edit/reorder/ReorderQuestion/ReorderQuestion.tsx
```diff
@@ -24,9 +24,9 @@ export const ReorderQuestion = ({ question, index, visible }: Props) => {
                 >
                     <div className={styles.tile}>
                         <div className={styles.handle} {...prov.dragHandleProps}>
-                            <Icon name="drag" size="m" />
+                            <Icon name="drag" size="m" alt="" />
                         </div>
-                        <Icon name="question" size="m" />
+                        <Icon name="question" size="m" alt="" />
```
mcmcgrath13: (nit, b): could you give these meaningful alts?

apps/modernization-ui/src/apps/page-builder/page/management/edit/reorder/ReorderSection/ReorderSection.tsx
```diff
@@ -46,15 +46,15 @@ export const ReorderSection = ({ section, index, visible }: Props) => {
...
-                                <Icon name="navigate-next" size="s" />
+                                <Icon name="navigate-next" size="s" alt="Expand section" />
                             )}
                         </div>
                         <div className={styles.handle} {...provided.dragHandleProps}>
-                            <Icon name="drag" size="m" />
+                            <Icon name="drag" size="m" alt="Drag handle to reorder section" />
                         </div>
-                        <Icon name="group" size="m" />
+                        <Icon name="group" size="m" alt="" />
```
mcmcgrath13: (nit, b): meaningful alt

apps/modernization-ui/src/apps/page-builder/page/management/edit/reorder/ReorderSubsection/ReorderSubsection.tsx
```diff
@@ -45,15 +45,15 @@ export const ReorderSubsection = ({ subsection, index, visible }: Props) => {
...
-                                <Icon name="expand-more" size="s" />
+                                <Icon name="expand-more" size="s" alt="Collapse subsection" />
                             )}
                         </div>
                         <div className={styles.handle} {...provided.dragHandleProps}>
-                            <Icon name="drag" size="m" />
+                            <Icon name="drag" size="m" alt="Drag handle to reorder subsection" />
                         </div>
-                        <Icon name="subsection" size="m" />
+                        <Icon name="subsection" size="m" alt="" />
```
mcmcgrath13: (nit, b): meaningful alt

apps/modernization-ui/src/apps/page-builder/page/management/edit/section/manage/ManageSectionTile/ManageSectionTile.tsx
```diff
@@ -112,10 +112,10 @@ export const ManageSectionTile = ({
                     ) : (
                         <div className={styles.manageSectionTile}>
                             <div className={styles.handle} {...provided.dragHandleProps}>
-                                <IconComponent name="drag" size="3" />
+                                <IconComponent name="drag" size="3" alt="Drag handle to reorder section" />
                             </div>
                             <div className={styles.label}>
-                                <IconComponent name="group" size="3" />
+                                <IconComponent name="group" size="3" alt="" />
```
mcmcgrath13: (nit, b): meaningful alt

apps/modernization-ui/src/apps/page-builder/page/management/edit/section/manage/ManageSection.tsx
```diff
@@ -13,7 +13,7 @@ import { ManageSectionTile } from './ManageSectionTile/ManageSectionTile';
 import styles from './managesection.module.scss';
 
 type ManageSectionProps = {
-    id: string;
+    id?: string;
```
mcmcgrath13: (nit, b): instead of making this optional, pass in an ID. The modal a11y requires that if this is in a modal (which it is in practice), then the id's set here are used back in the caller to set up proper aria attributes. It's a bit ugly and something that should be revisited for real when we get to page builder, but here we are

apps/modernization-ui/src/apps/page-builder/pages/BusinessRulesLibrary/TargetQuestion/TargetQuestion.tsx
```diff
@@ -180,7 +180,7 @@ export const TargetQuestion = ({
                                                 setTargetList([]);
                                             }}
                                         >
-                                            <Icon name="group" size="m" />
+                                            <Icon name="group" size="m" alt="" />
```
mcmcgrath13: (nit, b): meaningful alts

apps/modernization-ui/src/apps/report/run/filters/advanced/validator.spec.ts
```diff
@@ -19,7 +18,7 @@ describe('validateRule', () => {
                 field: '~',
                 operator: '~',
                 value: '',
-            } as RuleType,
+            } as QbQuery,
```
mcmcgrath13: (q, nb): do we still need these `as`?

apps/modernization-ui/src/apps/report/run/filters/basic/DateRangeFilter.tsx
```diff
@@ -32,7 +32,10 @@ const dateRangeValidator = (filter: BasicFilterConfiguration, label: string) =>
             return filter.isRequired ? validateRequiredRule(label).required.message : true;
         }
 
-        const errorMsg = validateDateRange(value, label);
+        const errorMsg = validateDateRange(
+            value.map((v) => v ?? ''),
```
mcmcgrath13, line 36: (q, nb): what was ts complaining about here? given the checks on line 31, `v` can't be nullish - is it because ts doesn't know it's always an array of length two?

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.spec.ts
```diff
@@ -42,8 +42,8 @@ describe('rangeValidator', () => {
         it('should not return an error msg for a valid numeric', () => {
             [
                 ['90210', '100000'],
-                [-1, 1.0],
-                [0.125, '0.9887'],
+                ['-1', '1.0'],
+                ['0.125', '0.9887'],
```
mcmcgrath13, line 46: (q, b): why does validate numeric range take strings here - shouldn't it accept numbers?

apps/modernization-ui/src/apps/search/useSearchCriteria.ts
```diff
@@ -100,7 +100,7 @@ const useSearchCriteria = <C extends object>({ defaultValues }: Options<C>): Int
                 .then((criteria) => dispatch({ type: 'wait', criteria }));
         } else if (state.status === 'preparing') {
             //  encrypt the new criteria
-            encrypt(state.criteria)
+            encrypt(state.criteria as Record<string, unknown>)
```
mcmcgrath13: (q, nb): instead of casing here could we tighten the definition of C in the signature to `extend Record<string, unknown>` instead of `object`?

apps/modernization-ui/src/components/ModalComponent/ModalComponent.tsx
```diff
@@ -33,7 +33,7 @@ export const ModalComponent = ({
             forceAction={disabled}
             ref={modalRef}
             isLarge={isLarge}
-            id={id}
+            id={id ?? ''}
```
mcmcgrath13: (nit, b): use `useId` here instead of an empty string - this could lead to weird a11y semantics

apps/modernization-ui/src/design-system/checkbox/Checkbox.spec.tsx
```diff
@@ -28,7 +28,7 @@ describe('Checkbox testing', () => {
     });
 
     it('should render checked', () => {
-        const { container } = render(<Checkbox id="test" label="Test Label" selected={true} />);
```
mcmcgrath13, line 31: (nit, b): using `screen` on line 33 is fishy. Instead of using that, use `container` and then this variable becomes used. (really, the use of container all over here is also kinda 🐟 , but less so than screen)

apps/modernization-ui/src/libs/patient/demographics/general/edit/GeneralInformationDemographicFields.tsx
```diff
@@ -88,7 +88,7 @@ const GeneralInformationDemographicFields = ({
                     orientation={orientation}
                     onBlur={onBlur}
                     onChange={onChange}
-                    value={value}
+                    value={value ?? undefined}
```
mcmcgrath13: (nit, b): using `undefined` as the default here could cause the input to go from uncontrolled to controlled. Is `null` viable instead?

apps/modernization-ui/src/libs/patient/demographics/sex-birth/edit/SexBirthDemographicFields.tsx
```diff
@@ -225,7 +225,7 @@ const SexBirthDemographicFields = ({
                             label={labels.order}
                             id={name}
                             name={name}
-                            value={value}
+                            value={value ?? undefined}
```
mcmcgrath13: (nit, b): save controlled<->uncontrolled issue

apps/modernization-ui/tsconfig.app.json
```diff
@@ -1,10 +1,11 @@
 {
     "compilerOptions": {
         /* base vite config */
+        "allowJs": true,
```
mcmcgrath13: (q, nb): where do we have js in src? any way we could not?

apps/modernization-ui/tsconfig.app.json
```diff
@@ -25,9 +26,11 @@
 
         /* nbs additions */
         "plugins": [{ "name": "typescript-plugin-css-modules" }],
+        "ignoreDeprecations": "6.0",
         "paths": {
             "*": ["./src/*"]
-        }
+        },
+        "types": ["vite/client", "node"]
```
mcmcgrath13: (q, b): this code won't run in a node environment - why do we need to include `node` in the types?

apps/modernization-ui/tsconfig.app.json
```diff
@@ -25,9 +26,11 @@
 
         /* nbs additions */
         "plugins": [{ "name": "typescript-plugin-css-modules" }],
+        "ignoreDeprecations": "6.0",
```
mcmcgrath13: (q, nb): what sort of things is this ignoring? should we create a follow on ticket to deal with them?

apps/modernization-ui/tsconfig.app.json
```diff
@@ -1,10 +1,11 @@
 {
     "compilerOptions": {
         /* base vite config */
+        "allowJs": true,
         "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
         "target": "ES2020",
         "useDefineForClassFields": true,
-        "lib": ["ES2020", "DOM", "DOM.Iterable"],
+        "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
mcmcgrath13, line 7: (q, b): reading https://stackoverflow.com/questions/42093758/need-clarification-of-the-target-and-lib-compiler-options it seems like the lib appeases typescript, but doesn't guarantee polyfills are available for things that are in 2023, but not in our target of 2020 - do we know if vite or something else includes those polyfills for us? or is it safer to leave this at 2020?

apps/modernization-ui/src/apps/page-builder/services/pagesAPI.ts
brick-green: I was trying to make as few changes as possible. I started removing them but started feeling like I didn't have enough context for why they were added in the first place. So I opted to leave them like this...I don't love either option tbh.

apps/modernization-ui/src/apps/report/run/filters/basic/DateRangeFilter.tsx
brick-green: yeah the check on line 31 only confirms that at least one element exists in the array. So the type of `value` is `(string | undefined)[]` but validateDateRange expects `string []`. I could go with something like this though:


```
if (!value || !value[0] || !value[1]) {
            return filter.isRequired ? validateRequiredRule(label).required.message : true;
}

const errorMsg = validateDateRange([value[0], value[1]], label);
```

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.spec.ts
brick-green: `export const validateNumericRange = (value: string[], field: string)` takes sting array then converts with `parseFloat`. I don't know why exactly it was written that way though.

apps/modernization-ui/src/apps/report/run/filters/basic/DateRangeFilter.tsx
mcmcgrath13: image
this seems to appease ts as well and is maybe more robust? maybe. thoughts?

apps/modernization-ui/src/apps/report/run/filters/utils/rangeValidator.spec.ts
mcmcgrath13: looks like it always dealing with strings, so this is all correct 💡

apps/modernization-ui/src/libs/patient/demographics/general/edit/GeneralInformationDemographicFields.tsx
brick-green: I checked `NumericInput` types, and its value prop was typed as `value?: number` (it doesn't accept null). So I updated `NumericInput` to accept `number | null`.

apps/modernization-ui/tsconfig.app.json
brick-green: Just checked again and it was only `reportWebVitals.js` called in `index.tsx`. But the output wasn't even logged anywhere so I commented out the line in index and removed the option in `tsconfig`

apps/modernization-ui/tsconfig.app.json
brick-green: This was an interesting topic to investigate. In the end I included `@vitejs/plugin-legacy` to add in the polyfills then `terser` to minimize

apps/modernization-ui/tsconfig.app.json
brick-green: I'm not sure. I removed it.

apps/modernization-ui/tsconfig.app.json
mcmcgrath13: heads up that I just deleted `reportWebVitals.js`, so once you merge in main, that should also go away

apps/modernization-ui/src/apps/page-builder/condition/search/ConditionTable.tsx
```diff
@@ -39,8 +39,10 @@ const columnToSortMap = new Map<string, ConditionSortField>([
     [ConditionColumn.Status, ConditionSortField.STATUS],
 ]);
 
-const asTableRow = (condition: Condition): TableBody => ({
-    id: condition.id,
+type ConditionTableBody = TableBody & { id: number };
```
mcmcgrath13: I lead you astray here, condition's id is absolutely a string (and not always a number at all (e.g. `DE-41665` is in our db) - we should have this be string and make sure the usage matches all that - apologies 🙏

apps/modernization-ui/tsconfig.app.json
mcmcgrath13: do we need the terser? looks like vite auto-minifies using `oxc`

https://vite.dev/config/build-options#build-minify

mcmcgrath13: ah, I see it's not in the version of vite we have right now

https://vite.dev/config/build-options#build-minify

do we need to set it as a build option for minification? maybe with a comment as to why and that we can probably remove once we update vite?

brick-green: hmm I don't think we need a separate minifier. Looks like to me that the default for vite version 8+ is oxc and we're on `"vite": "^8.0.16"`. So I'll remove terser.

apps/modernization-ui/src/apps/page-builder/condition/search/ConditionTable.tsx
brick-green: done :)

apps/modernization-ui/src/apps/page-builder/pages/AddNewPage/AddNewPage.tsx
```diff
@@ -65,7 +65,7 @@ export const AddNewPage = () => {
         });
     }, [watch.eventType]);
 
-    const handleAddConditions = (conditions: number[]) => {
+    const handleAddConditions = (conditions: string[]) => {
         const newConditions = conditions
             .map((id) => String(id))
```
mcmcgrath13: (nit): shouldn't need this map anymore
```suggestion
```

mcmcgrath13 (approved):
🧹

PR #3544, opened 2026-08-05
[APP-935] fix: sonar smell - extract mock to local variable and avoid complicated regex

mcmcgrath13:
## Description

Address sonar smells to not create a mock inline inside another mock's when, and to not use over-complicated regex

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-935

### DIFF
--- a/apps/modernization-api/src/test/java/gov/cdc/nbs/search/redirect/simple/SimpleSearchRedirectSteps.java
+++ b/apps/modernization-api/src/test/java/gov/cdc/nbs/search/redirect/simple/SimpleSearchRedirectSteps.java
@@ -14,16 +14,12 @@ import io.cucumber.java.en.Given;
 import io.cucumber.java.en.Then;
 import io.cucumber.java.en.When;
 import java.time.LocalDate;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 import org.springframework.test.web.servlet.ResultActions;
 import org.springframework.test.web.servlet.result.JsonPathResultMatchers;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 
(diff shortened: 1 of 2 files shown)


apps/modernization-api/src/test/java/gov/cdc/nbs/search/redirect/simple/SimpleSearchRedirectSteps.java
```diff
@@ -84,13 +80,10 @@ public void i_perform_a_search_from_the_nbs_home_screen() {
     String location = result.andReturn().getResponse().getRedirectedUrl();
 
     if (location != null) {
-      Matcher matcher = ENCRYPTED_LOCATION.matcher(location);
-
-      if (matcher.find()) {
-        String encrypted = matcher.group(1);
+      String[] parts = location.split("/");
+      String encrypted = parts[parts.length - 1];
```
emyl3, line 84: do we need to check that the length is greater than 1?

mcmcgrath13: since location is non-null, if the split doesn't split we end up with an array containing the original string, so no check needed here

mcmcgrath13: (this is also inside test code, so belt + suspenders seems like overkill)

emyl3: OOO gotcha! I missed that it was in the tests 😅

emyl3 (approved):
🫡

PR #3546, opened 2026-08-05
test: fix e2e selecting non-existant jd

mcmcgrath13:
## Description

I noticed the regression tests on main had failed consistently a few times not selecting jurisdiction. The jurisdiction it was trying to select isn't one in our golden, so not sure how this was working before, but let's get it fixed!


emyl3 (approved):
oooo nice catch... I am also wondering how this was working 😅

mcmcgrath13:
> oooo nice catch... I am also wondering how this was working 😅

Based on a screenshot, I think it sometimes happily went with a default 🤷

PR #3551, opened 2026-08-05
APP-937: Fix Sonar code smell (Date/Clock issues)

JordanGuinn:
## Description

Fixes the following Sonar smells:
- `Explicitly specify the time zone by passing a ZoneId or a Clock to the .now() method.`
- `Use the "java.time" API for date and time.` (this one is suppressed for the sake of not changing behavior)

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-937)


apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/filter/querydsl/QueryDSLDateFilterApplier.java
```diff
@@ -17,7 +18,7 @@ static BooleanExpression apply(final DateFilter filter, final TemporalExpression
 
       TemporalExpression<Instant> instant = (TemporalExpression<Instant>) expression;
 
-      LocalDate from = LocalDate.now();
+      LocalDate from = LocalDate.now(Clock.systemDefaultZone());
```
JordanGuinn, line 21: Had to specify the timezone here and `AuthenticationSupportSettingsConfiguration.java` manually as there doesn't appear to be a way to inject and autowire our ClockProvider in either case (this is just a standard Java class rather than a Spring `Component`, and `AuthenticationSupportSettingsConfiguration.java` is a Bean)

libs/testing/auth-cucumber/src/main/java/gov/cdc/nbs/testing/authorization/AuthenticationSupportSettingsConfiguration.java
```diff
@@ -11,6 +12,7 @@ class AuthenticationSupportSettingsConfiguration {
   @Bean
   AuthenticationSupportSettings authenticationSettings(
       final @Value("${testing.data.createdBy:9999}") long createdBy) {
-    return new AuthenticationSupportSettings(createdBy, LocalDateTime.now());
+    return new AuthenticationSupportSettings(
+        createdBy, LocalDateTime.now(Clock.systemDefaultZone()));
```
mcmcgrath13, line 16: (q, maybe-b): does the setting of the system default zone differ from how we set the clock? is there risk these times will be different?

JordanGuinn: Nope - in fact I pulled this choice from our existing [ClockProvider.java](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/modernization-api/src/main/java/gov/cdc/nbs/time/ClockProvider.java)

JordanGuinn: And [per docs for `LocalDateTime.now`](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDateTime.html#now--), the system clock with the default time-zone is used when no tz is provided:

> Obtains the current date-time from the system clock in the default time-zone.
> This will query the system clock in the default time-zone to obtain the current date-time.

mcmcgrath13 (approved):
⏲️

emyl3 (approved):
🕦

PR #3572, opened 2026-08-11
build: unpin jackson bom to match jackson version

mcmcgrath13:
## Description

dependabot opened a couple of PRs to update this pin, but we don't need to pin it at all - the jackson version in general is already set to 2.22.1


emyl3 (approved):
🔒

PR #3574, opened 2026-08-11
build: run npm audit fix, unpin js-yaml

mcmcgrath13:
## Description

Run `npm audit fix`. The `js-yaml` version was overriden to a version with a CVE, so unpinned that so it could go to latest


emyl3 (approved):
🔒

PR #3576, opened 2026-08-11
[APP-957] fix: simplify numeric and many dates

mcmcgrath13:
## Description

Change `Numeric` to be only externally controlled by removing state. Also remove the footgun of null vs undefined by standardizing the types on `number | null`... which then turned this into much more of a Thing as many of the date inputs use `Numeric` under the hood and then switch those all over to explicit null, which sprawled a bit. apologies.

Notes:
* When using react use hook form, we've been bitten quite a few times by using `undefined` as a default value as sometimes when un-setting things, this gets lost as it's seen as no data instead of data is none. This is a bigger issue across the codebase, but this takes away some of the footgun and encourages proper use for the inputs in the design system
* While this is ready for review, I'll hold off on merging til post-GA as it goes beyond the semi-freddo code state we're in in terms of scope

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-957


apps/modernization-ui/src/design-system/input/numeric/Numeric.tsx
```diff
@@ -25,31 +24,19 @@ const Numeric = ({
     placeholder,
     ...props
 }: NumericProps) => {
-    const { current, change, clear, initialize } = useNumeric(value);
-
-    useEffect(() => {
```
emyl3, line 30: nice! getting rid of these unnecessary useEffects! 🤩

apps/modernization-ui/src/design-system/input/numeric/useNumeric.ts
```diff
@@ -1,45 +0,0 @@
-import { useReducer } from 'react';
-
-type NumericValue = number | undefined | null;
-type State = { value?: NumericValue };
-type Action = { type: 'initialize'; value: NumericValue } | { type: 'change'; value: number } | { type: 'clear' };
-
-const reducer = (current: State, action: Action): State => {
```
emyl3, line 7: 👋 🙈

emyl3 (approved):
Thank you for fixing these up! Will be a good reference point going forward 🙌

apps/modernization-ui/src/apps/deduplication/configuration/match-configuration/pass-configuration/pass-form/matching-bounds/numeric-hint-input/NumericHintInput.tsx
```diff
@@ -13,11 +13,11 @@ type Props = {
     label: string;
     error?: string;
     onBlur: () => void;
-    onChange: () => void;
+    onChange: (v: number | null) => void;
```
JordanGuinn, line 16: (q, nb): Not sure how annoying this would be given the scope of impact, but maybe worth adding a comment to at least the most frequently used components clarifying that the use of `null` instead of `undefined` is intentional here?

Just thinking about anyone who may come across this down the line and, not knowing any better, change these params back to potentially `undefined` with `?`

JordanGuinn (approved):
🔥 🙏

apps/modernization-ui/src/apps/deduplication/configuration/match-configuration/pass-configuration/pass-form/matching-bounds/numeric-hint-input/NumericHintInput.tsx
mcmcgrath13: added a couple comments on the base-est components

PR #3577, opened 2026-08-11
fix: operator insertion when saving advanced filters

mcmcgrath13:
## Description

Add operators between rules and rule groups. Don't try to be smart and re-use operators, it backfires when we go to insert into the DB


emyl3 (approved):
Do you want me (as a follow on) to update the e2e for the advanced filter to account for saving the report and making sure the filter is saved?

Thank you for fixing 😭 🙌

mcmcgrath13:
> Do you want me (as a follow on) to update the e2e for the advanced filter to account for saving the report and making sure the filter is saved?
> 
> Thank you for fixing 😭 🙌

I feel good about the unit test coverage, but belt and suspenders are nice as a follow on

PR #3578, opened 2026-08-11
(fix): Remove extra whitespace from Report column sections

JordanGuinn:
## Description

Filters were expanding to use far more height than they should - this fixes that


mcmcgrath13 (approved):
tested it out and looks good!

emyl3 (approved):
LGTM! Nice fix!!

PR #3579, opened 2026-08-11
(fix): Move jaeger config to docker compose, to allow for local non-Docker dev without it

JordanGuinn:
## Description

This PR moves the Jaeger tracing config vars from `application-development.yml` to `docker-compose.yml`, so that local dev can still be spun up outside of Docker and not break if Jaeger is missing.  Also adds Jaeger as an service in the `build_modernized.sh` script just to be explicit.

## Tickets

* N/A


cdc-sandbox/build_modernized.sh
```diff
@@ -32,7 +32,7 @@ echo "Elasticsearch is ready"
 
 # Start application services (without NiFi) to initialize Elasticsearch indices
 echo "Starting modernized application services"
-docker compose -f $BASE/docker-compose.yml up modernization-api pagebuilder-api nbs-gateway report-execution keycloak --build -d
+docker compose -f $BASE/docker-compose.yml up jaeger modernization-api pagebuilder-api nbs-gateway report-execution keycloak --build -d
```
JordanGuinn, line 35: I don't think we actually _need_ this change since it's specified as a dependency of `modernization-api` in the root `docker-compose`, and thus is started as a part of initializing mod, but can't hurt to be explicit

mcmcgrath13 (approved):
🐳

emyl3 (approved):

PR #3581, opened 2026-08-11
fix: allow comma in advanced filter text value

mcmcgrath13:
## Description

Noticed when doing some type cleaning up that this case would cause issues with a genuine comma in a text field. Make sure we support that


JordanGuinn (approved):
🔍

emyl3 (approved):
OO nice catch!

PR #3583, opened 2026-08-11
fix: make filter requiredness validation reactive

mcmcgrath13:
## Description

When changing from a filter with other required fields to one without one, the requiredness was sticking. Make the validation function based instead of spec based to make sure it runs on the most current state of the world

### DIFF
--- a/apps/modernization-ui/src/apps/report/management/configuration/FilterRepeatingBlock.tsx
+++ b/apps/modernization-ui/src/apps/report/management/configuration/FilterRepeatingBlock.tsx
@@ -1,6 +1,6 @@
-import { useEffect, useState } from 'react';
+import { ReactNode, useEffect, useState } from 'react';
 
-import { Controller, FieldError, useWatch } from 'react-hook-form';
+import { Controller, useWatch } from 'react-hook-form';
 
 import { Shown } from 'conditional-render';
 import { RepeatingBlock } from 'design-system/entry/multi-value';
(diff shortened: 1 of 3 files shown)


JordanGuinn (approved):

apps/modernization-ui/src/apps/report/management/configuration/FilterRepeatingBlock.tsx
```diff
@@ -241,67 +241,154 @@ const FilterConfigForm = ({
...
+    const filterVal = useWatch<FilterConfig, 'filter'>({ name: 'filter' });
+    const needsSelectType = SELECTABLE_FILTER_IDS.has(filterVal?.value);
+
+    // reset when showability changes
+    useEffect(() => {
+        if (!needsSelectType) onChange(null);
+    }, [needsSelectType]);
+
+    return (
+        <Shown when={needsSelectType}>
```
JordanGuinn: (q, nb): Does the `Shown` element just hide the child element within?  DEF not for this moment, and def take any of my UI thoughts with a big grain of salt, but if it _does_ just hide them, I wonder what the implications might be of instead simply removing the child element from the DOM entirely when `needsSelectType` or similar is falsey

JordanGuinn (approved):
Makes sense to me ✅ :shipit:

apps/modernization-ui/src/apps/report/management/configuration/FilterRepeatingBlock.tsx
mcmcgrath13: that is indeed what shown does - I waffle back and forth on whether it's useful abstraction, but it's convention at this point, so I tend to use it in those cases where it feels configurable (vs just logic) 🤷

PR #3593, opened 2026-08-14
build: pin verison for org.apache.httpcomponents.core5:httpcore5 to fix High vuln

adaball:
## Description

Pin verison for `org.apache.httpcomponents.core5:httpcore5` to fix High vuln.


emyl3 (approved):
Thank you! 🔒

PR #3608, opened 2026-08-18
[APP-1006] fix: harden row limit checks - don't over-fetch data

mcmcgrath13:
## Description

the `rownumber` check was ineffective in practice - instead fetch batches until we surpass the limit or there's no more data to fetch

Notes:
* The batch size of 4096 is a SWAG - if I use a batch size of the row limit then the tests OOM in CI, so it appears some amount of pre-allocation happens based on this size (hence the power of two maybe being friendly to the computer). On the other hand we don't want to need to do a ton of DB round trips. Thoughts welcome on what's a good setting

## Tickets

* https://cdc-nbs.atlassian.net/browse/APP-1006


apps/report-execution/src/db_transaction.py
```diff
@@ -70,6 +63,18 @@ def execute(self, query: str, parameters: tuple = ()) -> None:
         self._cursor.execute(query, parameters)
         return None
 
+    def _fetch_rows(self):
+        row_limit = get_row_limit(self.is_export)
+        data = self._cursor.fetchmany(row_limit + 1)
+
+        # If there are any more rows to fetch beyond the limit, the result is too big
+        if len(data) == row_limit + 1:
```
brick-green: q, maybe b: do you need `row_limit + 1` in the fetch and in the check?

mcmcgrath13: Yes (I think). That way we know if the result size is too big or not, otherwise it could be perfectly at the limit - thoughts?

brick-green: oh duh, my brain read the `==` as a `>` for some reason

apps/report-execution/src/db_transaction.py
```diff
@@ -11,6 +11,8 @@
 INVALID_OBJECT_REGEX = re.compile("Invalid object name ('.*').")
 INVALID_COLUMN_REGEX = re.compile("Invalid column name ('.*').")
 
+BATCH_SIZE = 2**12
```
JordanGuinn:
```suggestion
BATCH_SIZE = 2**12 # 4096
```

apps/report-execution/src/errors.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, missing_columns: list[str]):
 class ResultTooBigError(BaseReportExecutionError):
     """The returned results are larger than allowed by configuration."""
 
-    def __init__(self, is_export: bool, row_limit: int, num_rows: int):
+    def __init__(self, is_export: bool, row_limit: int, num_rows: int | str):
```
JordanGuinn, line 45: (q, nb): When do we expect `num_rows` to be a string?

apps/report-execution/sample.env
```diff
@@ -7,4 +7,4 @@ NBS_DATASOURCE_SERVER="localhost,1433"
 DATABASE_CONN_STRING="Server=${NBS_DATASOURCE_SERVER};UID=${DATABASE_USER};PWD=${DATABASE_PASSWORD};Database=NBS_ODSE;TrustServerCertificate=yes;"
 
 # Configuration parameters with reasonable defaults set in the code
-# LOG_LEVEL=INFO
+# LOG_LEVEL=DEBUG
```
JordanGuinn, line 10: Intentional change?

JordanGuinn (approved):
🚀 👏

apps/report-execution/src/errors.py
mcmcgrath13: when we use `over <limit>` as the value - because of the batch strategy we don't really know how many there are

apps/report-execution/sample.env
mcmcgrath13: kinda - I thought it was a more useful commented out option as that's usually why you're un-commenting it. happy to revert if you'd prefer

JordanGuinn: No this works!  Just wanted to confirm 🙂

PR #3613, opened 2026-08-19
[APP-976] Remove unused Java files in mod api

adaball:
## Description

Removing Java files that are not used anywhere.

## Tickets

* [Jira Ticket](https://cdc-nbs.atlassian.net/browse/APP-976)


libs/options-api/src/main/java/gov/cdc/nbs/option/EnableOptions.java
```diff
@@ -1,16 +0,0 @@
...
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
-import org.springframework.context.annotation.ComponentScan;
-
-@Target(ElementType.TYPE)
-@Retention(RetentionPolicy.RUNTIME)
-@Documented
-@Inherited
-@ComponentScan
-public @interface EnableOptions {}
```
JordanGuinn, line 16: (q, nb): It looks like there are a handful of places where this annotated `@interface` paradigm is [defined](https://github.com/CDCgov/NEDSS-Modernization/blob/main/libs/redirect/src/main/java/gov/cdc/nbs/redirect/EnableRedirection.java) and [subsequently used](https://github.com/CDCgov/NEDSS-Modernization/blob/main/apps/question-bank/src/main/java/gov/cdc/nbs/questionbank/QuestionBankApplication.java), seemingly as a convenience method to enable said services in other services?

I don't know I have a strong feeling in either direction, but would there be any value in potentially keeping this around should we want to consume the `options-api` from the mod api or similar?

JordanGuinn (approved):
👏 🔥 🪓

libs/options-api/src/main/java/gov/cdc/nbs/option/EnableOptions.java
adaball: Not a bad idea.  Reverted here: https://github.com/CDCgov/NEDSS-Modernization/pull/3613/commits/5fd366345f0636c01b3882f58fa5ae4e218f9225

PR #3614, opened 2026-08-19
[APP-1006] fix: convert api => ui to streaming csv

mcmcgrath13:
## Description

Pull forward the same streaming csv content changes through to the api => ui exchange


apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/LibraryExecutionResult.java
```diff
@@ -1,10 +1,7 @@
 package gov.cdc.nbs.report.models;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import io.swagger.v3.oas.annotations.media.Schema;
+import java.io.InputStream;
 
 /** Model returned from the report-execution service when a library is invoked */
 public record LibraryExecutionResult(
```
mcmcgrath13: can we delete this file?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/models/ReportExecutionResult.java
```diff
@@ -1,11 +1,7 @@
 package gov.cdc.nbs.report.models;
 
-import io.swagger.v3.oas.annotations.media.Schema;
-import jakarta.validation.constraints.NotNull;
 import java.time.LocalDateTime;
```
mcmcgrath13: can we delete this file?

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -221,34 +237,102 @@ public ResponseEntity<ReportId> deleteReport(
...
+      if (contextHeader != null) {
+        responseToSet.setHeader("X-Report-Context-Header", contextHeader);
+      }
+
+      if (description != null) {
+        responseToSet.setHeader("X-Report-Description", description);
+      }
+
+      responseToSet.setHeader("X-Report-Timestamp", LocalDateTime.now(this.clock).toString());
+      responseToSet.setHeader("X-Report-Query", reportSpec.subsetQuery());
```
adaball: q: 

1. For sufficiently large subset queries do we think that it will exceed the HTTP server's limit?
2. Is there a need to escape the plain-text query here so it can be properly be read by the recipient?

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportControllerTest.java
```diff
@@ -627,11 +643,51 @@ void getReportRunner_should_return_422_status_code_when_report_has_no_library()
...
+    private final HttpHeaders responseHeaders = new HttpHeaders();
+
+    private final RestClient.RequestHeadersSpec.ConvertibleClientHttpResponse
+        reportExecHttpResponse =
+            mock(RestClient.RequestHeadersSpec.ConvertibleClientHttpResponse.class);
+
+    @SneakyThrows
+    @BeforeEach
+    void setUp() {
+      ServletOutputStream mockServlet = mock(ServletOutputStream.class);
```
adaball: nit: `s/mockServlet/mockServletOutputStream/`

apps/modernization-ui/src/apps/report/run/fetchReport.ts
```diff
@@ -0,0 +1,38 @@
+import { ReportExecutionRequest } from 'generated';
+import { ApiResult } from 'generated/core/ApiResult';
+import { catchErrorCodes, getResponseBody } from 'generated/core/request';
+
+// Manually invoking this endpoint instead of using the generated API client
+// because said API client doesn't correctly support the 'text/csv' media type.
```
adaball: q: Do we know why the generated clients don't correctly support `text/csv`?  Seems like it would be ideal to continue using said client if possible.

apps/modernization-api/src/test/java/gov/cdc/nbs/report/ReportControllerTest.java
```diff
@@ -743,19 +827,60 @@ void exportReport_should_return_422_status_code_when_report_not_export() {
...
```
adaball: nit: `s/mockServlet/mockServletOutputStream/`

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
mcmcgrath13: 1. my googling lead me to believe the limit is somewhere in the 8-16kb, which is pretty big in practice - I doubt we'll run into that limit. Do you think this limit is correct and/or this will become an issue?
2. What would need escaping? I know new lines are a problem, but we don't use those in the query building

apps/modernization-ui/src/apps/report/run/fetchReport.ts
mcmcgrath13: there are a couple issues at play, I'll update the comment to be more accurate

1. It will read the content into a string, which is inefficient for the file download case
2. Most problematically, it doesn't give us access to the headers, where other metadata is shoved and we need to get ou

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -221,34 +237,102 @@ public ResponseEntity<ReportId> deleteReport(
...
+      HttpServletResponse response) {
     LOGGER.log(
         System.Logger.Level.TRACE,
         "EXPORT report request received from user %s for report %s and datasource %s"
             .formatted(user.getId(), request.reportUid(), request.dataSourceUid()));
 
     if (!request.isExport())
       throw new IllegalArgumentException("isExport must be true when exporting a report");
-    return new ResponseEntity<>(reportExecutionClient.executeReport(request), HttpStatus.OK);
+    reportExecutionClient.executeReport(request, this.handleReportRes(response));
```
brick-green: nit, nb: just to match the other call to handleReportRes()
```suggestion
    reportExecutionClient.executeReport(request, handleReportRes(response));
```

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
```diff
@@ -221,34 +237,102 @@ public ResponseEntity<ReportId> deleteReport(
...
+      }
+
+      responseToSet.setHeader("X-Report-Timestamp", LocalDateTime.now(this.clock).toString());
+      responseToSet.setHeader("X-Report-Query", reportSpec.subsetQuery());
+
+      responseToSet.setContentType("text/csv");
+      responseToSet.setCharacterEncoding("UTF-8");
+
+      try (InputStream inputStream = resp.getBody()) {
+        inputStream.transferTo(responseToSet.getOutputStream());
```
brick-green: Question, nb: Do you need to flush the output stream at the end? My googling makes me think it's doesn't add value but i'm not 100% sure

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
adaball: 1. I agree that's pretty big in practice.  I was thinking of edge cases as I believe if header value's exceed the HTTP server's limit it result in a `4XX`.  Not a blocker just something to think about.
2. On second thought the plaintext should be OK, and given that there aren't any newlines there's no tokenization to deal with.

mcmcgrath13: 1. I tripped it 😅 ` org.apache.coyote.http11.HeadersTooLargeException: An attempt was made to write more data to the response headers than there was room available in the buffer. Increase maxHttpHeaderSize on the connector or write less data into the response headers.`

adaball (approved):
Overall LGTM

:shipit:

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
mcmcgrath13: My understanding is the combination of the try with resources syntax plus being in the request context means it's all handled automagically/properly by spring

apps/modernization-api/src/main/java/gov/cdc/nbs/report/ReportController.java
mcmcgrath13: this has since been addressed by limiting the query return to just the logic fragment and then additionally truncating that if somehow it's real long

JordanGuinn (approved):

PR #3618, opened 2026-08-20
[APP-1006] feat: compress results in local storage

mcmcgrath13:
## Description

Large reports that are _run_ can eat through the local storage quota. Compress the results before storing to make this less likely


emyl3 (approved):
LGTM! With this in place would the guidance/default continue to be 10K max rows on run for STLTs?

mcmcgrath13:
> LGTM! With this in place would the guidance/default continue to be 10K max rows on run for STLTs?

I think that's still reasonable. The issue is more the width of the data I was testing with than the length alone

brick-green (approved):

PR #3623, opened 2026-08-24
build: make pandas a dev dependency

mcmcgrath13:
## Description

Pandas is now only used in tests. Should we make it a dev dependency?

Pros:
* We don't use it in prod, so don't carry it around

Cons:
* If custom libraries want it, it's not there - is it worth keeping for its base data munging capabilities as part of the contract with custom libraries?


mcmcgrath13:
decided to keep pandas for custom library usage

PR #3616, opened 2026-08-20
Revert "[APP-957] fix: simplify numeric and many dates"

mcmcgrath13:
Reverts CDCgov/NEDSS-Modernization#3576

It was showing some issues with the date and number entering in the value input - needs more testing/iterating

PR #3626, opened 2026-08-25
Revert "Revert "[APP-957] fix: simplify numeric and many dates""

mcmcgrath13:
Reverts CDCgov/NEDSS-Modernization#3616

Changes to fix issues:
* use null instead of empty string when clearing out dates
* simplify change handler for single values in advanced filter and cop we don't really know the type


emyl3 (approved):
Working as expected for me! checked the adv filters, search and patient file

PR #3627, opened 2026-08-26
build: update minimal node version to match jsdoc requirements

mcmcgrath13:
## Description

After updating #3625 I saw some warnings when installing about supported engines, so updated the node engine version for the package to be in compliance


emyl3 (approved):
confirmed updating to that version of node and re-installing packages did not show jsdoc related warnings!

JordanGuinn (approved):

PR #3628, opened 2026-08-26
test: use pandas in custom library to ensure dep is exercised

mcmcgrath13:
## Description

Pandas is now only available/used by custom libraries, so make sure we exercise using it to make sure it stays a production dependency


adaball (approved):
LGTM once the checks go green.

:shipit:

JordanGuinn (approved):

